PR #1, opened 2020-02-11
First pass on proto definition

skearnes:
A few notes (besides abundant TODOs in the code):
* Modified a few names for brevity; e.g. `ReactionInputType` -> `ReactionRole`
* Included units in variable names where appropriate
* Removed `TemporalData` in favor of local `Measurement` messages for better unit support
* For enums, note that the 0 value is the default; usually I put `UNKNOWN, but sometimes `NONE` when that seemed more reasonable (e.g. for `ElectrochemistryType`). I think this probably makes sense for any Type enum that doesn't support custom values.
* Due to enum scoping restrictions (can't use `UNKNOWN` more than once in a message), I prefixed all enum values with their type names. The protoc error is helpful: `Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "UNKNOWN" must be unique within "PressureConditions", not just within "Atmosphere".` This helps justify some of the naming stutter :)

Let me know if you hate any of these changes.

### DIFF
--- a/proto/ord_schema.proto
+++ b/proto/ord_schema.proto
@@ -1,0 +1,369 @@
+// Schema for the Open Reaction Database.
+syntax = "proto3";
+
+message Reaction {
+  string reaction_smiles = 1;
+  string atom_mapped_reaction_smiles = 2;
+  string how_atom_mapped = 3;
+  string rinchi = 4;
+  string reaction_name = 5;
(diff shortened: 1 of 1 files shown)


skearnes:
(Obviously I still need to add a bunch of comments to the proto, but not in this PR.)

PR #2, opened 2020-02-12
Create run_protoc.yml

skearnes:
Add a GitHub action to run protoc to make sure the protocol buffers compile.


skearnes:
I think this will work after #1 is merged.

Issue #3, opened 2020-02-12
Correct place for unit definition


connorcoley:
Should the units expected of each field be defined (a) in a comment, (b) in the name of the field, or (c) in a separate enum to allow users to define values however they would like?

skearnes:
My preference is for (a) and (b). This provides a couple of advantages:
* When the unit is part of the variable name, it makes it easier for people to remember which units to use.
* If we force people to use a single unit, we can avoid having to do conversions to a common unit for e.g. searching and it will reduce the complexity of the API/schema overall.

connorcoley:
One concern I have is whether manual entry might result in incorrect unit conversions. If we let people select units from an enum, we can at least simplify conversions in some validation/standardization script. Time (hours / minutes / etc.) is fine, but I have slightly less confidence in manual conversions of temperature (F / C) and pressure (bar / psi / atm).

skearnes:
That's a good point. I'm still worried about the complexity of an extra enum; what if instead we used a oneof with a couple different units that we expect? Then the variable names would still be explicit.

For example:
```
oneof pressure {
  float pressure_in_bar = 1;
  float pressure_in_atom = 2;
  float pressure_in_psi = 3;
}
```
What do you think?

connorcoley:
I recognize the clarity of having units in the field names to avoid any ambiguity, but it feels sloppier to me than having ```pressure_value``` and ```pressure_units``` fields. I'm not as clear on the downside of having too many enums. Being able to extend the schema to handle different/new units by adding options to a single enum seems better to me than extending a oneof.

skearnes:
OK, let's give enums a try.

PR #4, opened 2020-02-13
Second pass

connorcoley:
- Changing all units to have value and unit fields; pre-specify allowed units (as per issue #3)
    - This led to the creation of many message types (requires review)
- Make most important messages top-level message
- Use "kind" for all one-of variables
- Use separate repeated "Identifier" for Compound. Have field for "value" and field for "type" with extensible enum (smiles, niche, etc.)
    - Same for reactions; in the process, removed field for "how_atom_mapped"
- Keep map <string, bytes> for automation_code
- Temperature measurement should have enum for methods, with an allowed custom. Use UNKNOWN as default 
- Add custom for flow and electrochemistry conditions
- Added measurements for electrochemistry
- Keep stirring rate as *not* a oneof
- Added enum for flow tube material
- YieldType makes use of ReactionAnalysis.AnalysisType so that analytical methods are defined in a single enum in the schema
- No units for selectivity need to be specified
- Create string timestamp for DateTime and worry about parsing later (requires review)
- Extend AnalysisType enum to include LCMS, GCMS, ELSD
- Add string field for analysis details to allow additional specification (e.g., of column type, gradient)
- Use two spaces before inline comments
- Changed custom fields to be ```<field name of enum>_custom``` rather than ```custom_<field name of enum>```
- Added flow rate support for flow synthesis
- DOI and patent no longer oneof, since could have both describing a single physical experiment
- Documentation of how I envision solutions or supported catalysts to be used


proto/ord_schema.proto
```diff
@@ -45,9 +61,15 @@ message ReactionInput {
...
+  // Whether this species should be considered a limiting reactant.
   bool is_limiting = 3;
+  /**
+   * Used to define order of addition. ReactionInputs with the same 
+   * addition_stage were added simultaneously. One ReactionInput with a 
+   *  lower addition_stage than another was added earlier in the procedure.
+   */
   int32 addition_stage = 4;
-  float addition_time = 5;
+  float addition_time = 5;  // Reaction time (in hours)
```
skearnes: Should this have an explicit unit, `addition_time_in_h`?

Or, use a Time message here?

proto/ord_schema.proto
```diff
@@ -60,50 +82,148 @@ message ReactionInput {
...
+ * We discourage chemicals from being defined only by a name.
+ */
+message CompoundIdentifier {
+  // Possible identifier types are listed in an enum for extensibility
+  enum IdentifierType {
+    COMPOUND_IDENTIFIER_TYPE_UNKNOWN = 0;  // Any other identifier
+    COMPOUND_IDENTIFIER_TYPE_SMILES = 1;  //  Simplified molecular-input line-entry system.
+    COMPOUND_IDENTIFIER_TYPE_INCHI = 2;  // IUPAC International Chemical Identifier.
+    COMPOUND_IDENTIFIER_TYPE_MOLFILE = 3;  // MDL Molfile V3000.
+    COMPOUND_IDENTIFIER_TYPE_IUPAC_NAME = 4;  // Chemical name following IUPAC nomenclature recommendations.
```
skearnes: Let's keep an 80 char limit; in this case we can move the comments to their own line above (and maybe be consistent within the enum)

proto/ord_schema.proto
```diff
@@ -296,74 +490,206 @@ message ReactionProduct {
...
+// redundant, but I think this is actually a fairly clean way to do it, rather
+// than putting all of the possible measurement/unit types into one giant enum
+/**
+ * To allow users to describe synthetic processes in whatever units they find
+ * most natural, we define a fixed list of allowable units for each measurement
+ * type. Upon submission to a centralized database, or using a validation and
+ * canonicalization script, we will convert all values to the default units
+ * (the first item in each enum).
+ */
+message Time {
```
skearnes: Maybe ReactionTime to make it clearer that we are measuring from the reaction start?

proto/ord_schema.proto
```diff
@@ -296,74 +490,206 @@ message ReactionProduct {
...
+}
+
+message Person {
+  string username = 1;
+  string name = 2;
+  string orcid = 3;
+  string organization = 4;
+}
+
+// TODO(ccoley) is it appropriate to have these be part of messages? It looks
```
skearnes: I like this separation by kind. I might suggest making the units embedded messages rather than top-level so they are defined closer to where they are used (makes it easier to read). However, if we have units that are used in more than one place we should be consistent and define them all at the same level.

proto/ord_schema.proto
connorcoley: I prefer keeping the message as just ```Time``` (since that's more like a primitive value) but specifying reaction time in fields as needed

proto/ord_schema.proto
connorcoley: Good catch

proto/ord_schema.proto
connorcoley: Time is certainly used in many places; Length and Volume are too. The others are more self-contained (e.g., current and voltage for electrochemistry) but I agree that consistency is more important

PR #5, opened 2020-02-15
Add setup.py and some simple tests for the python/proto interface

skearnes:
To use:
```
$ python setup.py build_py
$ python setup.py install
$ python ord_schema/proto/ord_schema_pb2_test.py
```

After this is merged, I'll update the CI to run these tests automatically.

### DIFF
--- a/.github/workflows/run_protoc.yml
+++ b/.github/workflows/run_protoc.yml
@@ -1,6 +1,12 @@
 name: CI
 
-on: [push]
+on:
+ push:
+   branches: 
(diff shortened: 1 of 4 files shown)


skearnes:
FYI I'm using Google python style: http://google.github.io/styleguide/pyguide.html

skearnes:
Thanks! I like absl a lot; we use it internally and absl.flags has the best flag parsing setup I've seen: https://abseil.io/docs/python/guides/flags

PR #14, opened 2020-02-25
Handling mixtures more elegantly, adding useful enums

connorcoley:
Following this morning's discussion...
- "addition_stage" -> "addition_order" // 1-indexed
- How to differentiate all flavors of NMR? [in details]
- How to allow stock solutions with more than one solute? [now using ```components``` field of Compound message; please review]
- Compound preparation enum: add COMPOUND_PREPARATION_SYNTHESIZED = 5;
- Add glass vial to ReactionSetup enum, since it is very common

### DIFF
--- a/proto/ord_schema.proto
+++ b/proto/ord_schema.proto
@@ -300,22 +300,30 @@ message ReactionSetup {
   VesselType vessel_type = 1;
   // Used to describe an UNKNOWN vessel type or elaborate.
   string vessel_type_custom = 2;
+  enum VesselMaterial {
+    VESSEL_MATERIAL_UNKNOWN = 0;
+    VESSEL_MATERIAL_GLASS = 1;
+    VESSEL_MATERIAL_POLYPROPYLENE = 2;
+    VESSEL_MATERIAL_PLASTIC = 3;
+  }
+  VesselMaterial vessel_material = 3;
+  string vessel_material_custom = 4;
   enum VesselPrep {
     VESSEL_PREP_UNKNOWN = 0;
     VESSEL_PREP_NONE = 1;
     VESSEL_PREP_OVEN_DRIED = 2;
   }
-  VesselPrep vessel_prep = 3;
+  VesselPrep vessel_prep = 5;
   // Used to describe an UNKNOWN vessel prep.
-  string vessel_prep_custom = 4;
+  string vessel_prep_custom = 6;
   // Size (volume) of the vessel.
-  Volume vessel_volume = 5;
+  Volume vessel_volume = 7;
   // Specification of automated protocols.
-  bool is_automated = 6;
+  bool is_automated = 8;
   // Automated platform name, brand, or model number.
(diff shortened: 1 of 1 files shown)


connorcoley:
Latest commit addresses #15

proto/ord_schema.proto
```diff
@@ -40,73 +40,174 @@ message Reaction {
...
   AdditionSpeed addition_speed = 6;
 }
 
 message Compound {
+  // Set of identifiers used to uniquely define this compound.
+  // Solutions or mixed compounds should use the NAME identifier
+  // and list all constituent compounds in the "components" field.
   repeated CompoundIdentifier identifiers = 1;
+  // Quantity of this compound used in the reaction.
   Amount amount = 2;
```
skearnes: Should amount/components be a Oneof?

proto/ord_schema.proto
```diff
@@ -186,74 +284,21 @@ message Amount {
...
 message ReactionSetup {
   enum VesselType {
     VESSEL_TYPE_UNKNOWN = 0;
     VESSEL_TYPE_ROUND_BOTTOM_FLASK = 1;
-    VESSEL_TYPE_WELL_PLATE = 2;
-    VESSEL_TYPE_MICROWAVE_VIAL = 3;
-    VESSEL_TYPE_TUBE = 4;
-    VESSEL_TYPE_CONTINUOUS_STIRRED_TANK_REACTOR = 5;
-    VESSEL_TYPE_PACKED_BED_REACTOR = 6;
+    VESSEL_TYPE_VIAL = 2;  // Regular glass vial
```
skearnes: Useful to distinguish between glass/plastic?

proto/ord_schema.proto
```diff
@@ -599,114 +660,126 @@ message Person {
  */
 message Time {
   enum TimeUnit {
-    TIME_UNIT_HOUR = 0;
-    TIME_UNIT_MINUTE = 1;
-    TIME_UNIT_SECOND = 2;
+    TIME_UNIT_UNKNOWN = 0;
```
skearnes: Why the UNKNOWN values for these units? Is this just to have a better default? If so, that's fine but I wanted to check.

proto/ord_schema.proto
connorcoley: I envision a use-case where "amount" is used to specify the volume of a stock solution that is added, while "components" specifies the make-up of the solution, whose total volume can exceed the amount that was added in this particular reaction.

proto/ord_schema.proto
connorcoley: Yes, good point. Would a separate enum be overkill? Not all plastics are equal, but I do think that ```VESSEL_MATERIAL_xxxx``` could still be useful

proto/ord_schema.proto
connorcoley: Yes, sorry for not documenting the reason/issue well

PR #19, opened 2020-03-02
Descriptive error for when protoc isn't available

skearnes:


skearnes:
Duplicate; sorry for the noise.

PR #24, opened 2020-03-03
Shorten enum names

skearnes:
We were including prefixes to avoid collisions, but this made working with them somewhat unwieldy. This PR removes the prefixes except for the UNKNOWN cases.


skearnes:
Just double checking before I merge this:

I think this is a nice compromise between (a) the style guide [suggestion](https://developers.google.com/protocol-buffers/docs/style#enums) to prefix enums rather than nest them in an enclosing message and (b) reducing complexity for users. But it does lose some of the formal specification / scoping (e.g. `REACTION_SMILES` vs. `REACTION_IDENTIFIER_TYPE_REACTION_SMILES`).

The nice thing is that there are only a few instances of actual collisions in our schema (when a message contains multiple enums), and we are unlikely to have these collisions outside of the `UNKNOWN` value.

connorcoley:
Would ```NAME``` and ```NONE``` lead to conflicts as well? I noticed you kept the full prefix for the ```UNKNOWN``` options.

skearnes:
The current proto compiles; NAME is used multiple places, but in different messages, so it doesn't lead to a conflict. I guess the question is whether we're ok with using 'NAME' in both places and letting the surrounding code provide the scope. I think I am ok with this since the context should be pretty clear and the prefixes mostly felt redundant.

connorcoley:
I'm not opposed to using NAME and NONE in multiple contexts. Feel free to merge this PR

PR #25, opened 2020-03-03
Use a common Measurement message

skearnes:
(Note: this is out for feedback and I haven't updated the tests or anything yet.)

We have a lot of different places where measurements are recorded. For `Amount` in particular, I don't really like how `value` isn't a top-level field (it's nested in one of the oneof messages like `Mass`).

This PR proposes a common `Measurement` message that is used anywhere we need to record a value. It includes time and measurement precision information (both optional), and groups all unit enums under a single message type. One immediately noticeable consequence: `Amount` is no longer needed; we just have a `Measurement` under `Compound`, so we've removed at least one level of nesting.

This should make for a more consistent experience when using the schema, since all values are just `Measurements`.

Thoughts?

### DIFF
--- a/proto/ord_schema.proto
+++ b/proto/ord_schema.proto
@@ -97,8 +97,12 @@ message Compound {
   // Solutions or mixed compounds should use the NAME identifier
   // and list all constituent compounds in the "components" field.
   repeated CompoundIdentifier identifiers = 1;
-  // Quantity of this compound used in the reaction.
-  Amount amount = 2;
+  // The quantitative Amount of a Compound used in a particular reaction.
+  // Compounds added in their pure form should have their value defined by mass,
+  // moles, or volume. Compounds prepared as solutions should be defined in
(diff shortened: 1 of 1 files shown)


connorcoley:
I've been thinking about this proposal and am not yet convinced it would be an improvement:

Pros:
- Removes an extra layer of nesting a compound's "amount" 
- Reduces the number of unique messages  (is this a pro?)

Cons:
- No longer strongly enforce values with units to be of the correct type (e.g., can say that the amount of a compound added is 5 volts)
- The oneof field used in a measurement has a name that defines the type, rather than a single "units" field. ```Measurement(value=5, voltage=VOLT)``` seems less intuitive than ```Measurement(value=5, units=VOLT)``` to me. The latter would require us to have a single enum that lists every measurement unit, so I also don't like that solution. 

For me, I find the first con to be very significant.

skearnes:
I agree with these concerns, and the importance of the first con. Closing.

PR #28, opened 2020-03-05
Revisions to how inputs and compounds are delineated

### LINKED ISSUE #27: Should reaction role be moved to "compound" from "reactioninput"?
When modifying the example of a pre-mixed stock solution, I started to confuse myself as to the ```reaction_role``` and ```is_limiting``` fields of the ReactionInput. My inclination is to move these two fields to be part of the ```Compound``` message, since we should separately specify the role and is_limiting of different solutes or solvents.

Now, this makes things a little funny because the solution-level compound message won't have a defined role, whereas the pure solute-level and solvent-level compound message will. 

In keeping with the spirit of making the schema as closely aligned with a chemist's actions (or a robot's actions), what if we more strictly treat the ```ReactionInput``` as something that is added to the vessel? What I might propose is the following:

- A ```ReactionInput``` contains a ```repeated Compound components``` field that is _always_ populated. A ReactionInput can be a pure substance (with a single components message) or a mixture (with multiple components messages).
- We do not give ```ReactionInput``` messages their own ```amount``` field. A reason to consider adding one would be if the ```components``` define a large-scale stock solution and we just want to change the amount of that solution that is added for a particular reaction. However, I think this might introduce too much complexity; it could be better to force each component compound's amount to be the actual amount that was added.
- ```Compound``` messages now contain fields for ```reaction_role``` and ```is_limiting```. A compound is always a pure substance. We no longer have nested compound messages.
- ```ReactionInputs``` are given a ```string description``` field; this is for convenience and bookkeeping only and could be omitted. If a pure substance, the description could be the name of its sole component, like "24 mg of XPhos". If a stock solution, the description could be a text description like "125 uL of 4 M NaOH in water". A lightweight parsing tool with basic regex could potentially be used to fully populate the ```components``` field given this description, if so desired.

Thoughts?

connorcoley:
Implements changes discussed in #27

Edit: fixes #27

### DIFF
--- a/proto/ord_schema.proto
+++ b/proto/ord_schema.proto
@@ -53,7 +53,76 @@ message ReactionIdentifier {
 }
 
 message ReactionInput {
-  Compound compound = 1;
+  // Description or name of this pure substance or mixture that
+  // was added to the reaction vessel. This description is 
+  // intended to be used for convenience and legibility, and does not 
+  // need to be defined.
+  string description = 1;
+
+   /**   
+   * We use the components field for pure substances and mixtures.
+   * 
+   * For example, suppose we are adding
+   * 3 mL of a 4 M solution of NaOH in water.
+   * 
+   * input {
+   *   description: "3 mL of 4M NaOH solution in water"
+   *   components: [
+   *     {
+   *       identifiers: [
+   *         {type: IDENTIFIER_SMILES, value: "O"},
+   *         {type: IDENTIFIER_NAME, value: "water"}
+   *       ]
+   *       amount: {
+   *         volume: {value: 3, units: VOLUME_UNITS_MILLILITER}
+   *       }
+   *     }
+   *  components: [
+   *     {
+   *       identifiers: [
+   *         {type: IDENTIFIER_SMILES, value: "[Na+].[OH-]"},
+   *         {type: IDENTIFIER_NAME, value: "sodium hydroxide"}
+   *       ]
+   *       amount {
+   *         concentration: {value: 4, units: CONCENTRATION_UNITS_MOLAR}
+   *       }
+   *     }
+   *   ]
+   * }
+   */
+   repeated Compound components = 2;
+  /**
+   * Used to define order of addition. ReactionInputs with the same 
+   * addition_order were added simultaneously. One ReactionInput with a 
+   * lower addition_order than another was added earlier in the procedure.
+   * This field is 1-indexed.
+   */
+  int32 addition_order = 3;
+  // When the addition event took place in terms of the reaction time (or, 
+  // in the case of flow chemistry, the residence time).
+  Time addition_time = 4;
+  enum AdditionSpeed {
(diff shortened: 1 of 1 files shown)


proto/ord_schema.proto
```diff
@@ -99,78 +123,22 @@ message Compound {
...
-   *         moles: {value: 4, units: MOLES_UNIT_MOLES}
-   *       }
-   *     }
-   *   ]
-   * }
-   *
-   * A similar definition is used for supported catalysts (TODO)
-   */
-  repeated Compound components = 3;
+  enum ReactionRole {
```
skearnes: I think I still prefer having this at the level of `ReactionInput`; if the input is limiting, you add more of that input, not more of the component compound...right?

proto/ord_schema.proto
```diff
@@ -89,7 +111,9 @@ message ReactionInput {
     CONTINUOUS = 5;
   }
   // The qualitative rate of addition.
-  AdditionSpeed addition_speed = 6;
+  AdditionSpeed addition_speed = 5;
+  // For continuous synthesis, we instead specify a flow rate.
+  FlowRate flow_rate = 6;
```
skearnes: I like this move out of `Amount`.

proto/ord_schema.proto
```diff
@@ -53,32 +53,54 @@ message ReactionIdentifier {
...
-    REACTANT = 1;
-    REAGENT = 2;
-    SOLVENT = 3;
-    CATALYST = 4;
-  }
-  ReactionRole reaction_role = 2;
-  // Whether this species was intended to be a limiting reactant.
-  bool is_limiting = 3;
+  // Description or name of this pure substance or mixture that
+  // was added to the reaction vessel. This description is
```
skearnes: I agree with this interpretation that you elaborated on in #27; I think it's a great idea.

proto/ord_schema.proto
connorcoley: I struggled with this decision a bit. The reason why I think ReactionRoles should be assigned to pure species (compounds) is that we should be able to parse one input into its constituent components. I see two potential use cases:
- When templating reactions, if we have an input with one REACTANT component and one SOLVENT component, we can write python wrappers to easily exchange the REACTANT (holding the SOLVENT constant, or vice versa) by iterating through the ```components``` without having to know the order components are defined in.
- We can programmatically generate a reaction SMILES for a reaction proto that does not yet have it defined by looking for all REACTANT compounds. This would let us easily generate reaction SMILES strings that, by default, exclude solvents that are presumed to be spectators. Similarly, if we have a solid supported catalyst with one CATALYST component--palladium--and one UNKNOWN (SUPPORT?) component--carbon--we can determine that the catalytically active component is palladium.

skearnes: SGTM; I think I was confusing this with our conversation about an enum (that doesn't exist) for solute vs. solvent, etc.

PR #30, opened 2020-03-06
Add precision to measurement messages

### LINKED ISSUE #29: Add precision to all measurement types?
`Amount` currently has a `precision_in_percent` field. How would you feel about adding a precision field to all of the measurement types and removing this one?

* The precision would be closer to the value.
* There may be precision information for non-`Amount` measurements, such as voltages or flow rates.
* (optional) This would allow us to remove the nesting that stills bugs me in `Compound`, since `Amount` could be replaced with the `oneof` it currently wraps (we would change the oneof name to "amount" from "kind" for programmatic extraction of set values).

skearnes:
Fixes #29; PTAL at the TODO on line 540 of the changed file.

### DIFF
--- a/proto/ord_schema.proto
+++ b/proto/ord_schema.proto
@@ -121,8 +121,18 @@ message Compound {
   // Solutions or mixed compounds should use the NAME identifier
   // and list all constituent compounds in the "components" field.
   repeated CompoundIdentifier identifiers = 1;
-  // Quantity of this compound used in the reaction.
-  Amount amount = 2;
+  /**
+   * The quantitative Amount of a Compound used in a particular reaction.
+   * Compounds added in their pure form should have their value defined by
+   * mass, moles, or volume. Compounds prepared as solutions should be defined
+   * in terms of their volume. Compounds prepared on solid supports should
+   * define the total mass/volume including the support.
+   */
+  oneof amount {
+    Mass mass = 2;
+    Moles moles = 3;
+    Volume volume = 4;
+  }
   enum ReactionRole {
     // Unspecified.
     REACTION_ROLE_UNKNOWN = 0;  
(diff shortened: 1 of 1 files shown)


proto/ord_schema.proto
```diff
@@ -525,30 +516,30 @@ message ReactionOutcome {
...
     ER = 2;  // Enantiomeric ratio. (x:1)
     DE = 3;  // Diasteromeric ratio (x:1)
   }
-  SelectivityType selectivity_type = 8;
-  string selectivity_type_custom = 9;
-  float selectivity_value = 10;
-  float selectivity_precision_in_percent = 11;
+  SelectivityType selectivity_type = 7;
+  string selectivity_type_custom = 8;
+  // TODO(kearnes): Should these fields be replaced with Percentage?
```
connorcoley: Only in the case of EE will the selectivity be a percentage. If we want to allow ER and DE to be specified (which I think we do), we'll have to leave it as a float. ```selectivity_precision_in_percent``` is a mouthful though

skearnes: OK, that's what I thought.

Should this just be `selectivity_precision` and maintain the "in the same units" paradigm? Although I'm not entirely sure what that means for a ratio...

connorcoley: I don't know that I have ever seen EEs, DEs, ERs, etc. reported with uncertainty values. This would be a great question for Abby. I think you can safely change it to ```selectivity_precision``` and we can worry about what that means later

skearnes: Done.

skearnes:
PTAL, updated `setpoint` fields for `TemperatureConditions` and `PressureConditions` as well.

PR #34, opened 2020-03-09
Patch yield

connorcoley:
- Fixes "yield" field by changing to "compound_yield"


connorcoley:
PTAL at changes to ```purity``` and ```selectivity```. Should brevity outweigh consistency?

skearnes:
I think I'm in favor of brevity here. I might even suggest that we only change `yield` to `compound_yield` and leave the other yield instances alone...

PR #38, opened 2020-03-10
Create separate messages for enums with custom values

### LINKED ISSUE #37: Proposal: separate messages for custom enums
We have several instances of this pattern in the schema:

```proto
enum Preparation {
  COMPOUND_PREPARATION_UNKNOWN = 0;
  NONE = 1;
  REPURIFIED = 2;
  SPARGED = 3;
  DRIED = 4;
  SYNTHESIZED = 5;
}
Preparation preparation = 7;
// Custom preparation, used only when 'type' field is UNKNOWN.
string preparation_custom = 8;
```

Here's an alternative suggestion:

```proto
message Preparation {
  enum PreparationType {
    UNSPECIFIED = 0;
    NONE = 1;
    REPURIFIED = 2;
    SPARGED = 3;
    DRIED = 4;
    SYNTHESIZED = 5;
  }
  oneof kind {
    PreparationType type = 1;
    string custom = 2;
  }
}
```

This has a few advantages:
* It is unambiguous when a custom type is used.
* Groups everything related to the preparation together nicely.
* Allows use of UNSPECIFIED in all enums so we can ditch all the prefixes.
* The field names can be reused for consistency much like we do for the unit messages.

Thoughts?

skearnes:
Fixes #37 

Tricky bits to pay special attention to:
* Vessel
* Somewhat arbitrary choices of when to nest vs. not when defining new messages.

Note that I also changed UNKNOWN to UNSPECIFIED to follow the [style guide](https://developers.google.com/protocol-buffers/docs/style#enums).

### DIFF
--- a/proto/ord_schema.proto
+++ b/proto/ord_schema.proto
@@ -43,9 +43,9 @@ message ReactionIdentifier {
     RINCHI = 3;  // Reaction InChI.
     NAME = 4;  // Named reaction or reaction category.
   }
-  oneof kind {
+  oneof identifier_type {
     IdentifierType type = 1;
-    string custom = 2;
+    string type_custom = 2;
   }
   string value = 3;
 }
@@ -145,7 +145,7 @@ message Compound {
   ReactionRole reaction_role = 5;
   // Whether this species was intended to be a limiting reactant.
   bool is_limiting = 6;
-  Preparation preparation = 7;
+  CompoundPreparation preparation = 7;
   // Name of the vendor or supplier the compound was purchased from.
   string vendor_source = 8;
   // Compound ID in the vendor database or catalog.
@@ -176,7 +176,7 @@ message Compound {
  * is an oversimplification of the full procedure, which can be detailed
  * in the 'preparation_details' field.
  */
-message Preparation {
+message CompoundPreparation {
   enum PreparationType {
     UNSPECIFIED = 0;
     // Compound used as received.
@@ -190,9 +190,9 @@ message Preparation {
     // Compound synthesized in-house
     SYNTHESIZED = 5;
   }
-  oneof kind {
-    PreparationType preparation = 1;
-    string custom = 2;
+  oneof preparation_type {
+    PreparationType type = 1;
+    string type_custom = 2;
   }
   // Full description of how the received compound was prepared.
   string details = 3;
@@ -237,9 +237,9 @@ message CompoundIdentifier {
     // Protein data bank ID (for enzymes)
     PDB_ID = 13;
   }
-  oneof kind {
+  oneof identifier_type {
     IdentifierType type = 1;
-    string custom = 2;
+    string type_custom = 2;
   }
   // Value of the compound identifier; certain types (e.g., PUBCHEM_CID) may
   // cast the string as an integer for downstream processing and validation.
@@ -247,49 +247,40 @@ message CompoundIdentifier {
 }
 
 message Vessel {
-  message VesselForm {
-    enum VesselFormType {
-      UNSPECIFIED = 0;
-      ROUND_BOTTOM_FLASK = 1;
-      VIAL = 2;
-      WELL_PLATE = 3;
-      MICROWAVE_VIAL = 4;
-      TUBE = 5;
-      CONTINUOUS_STIRRED_TANK_REACTOR = 6;
-      PACKED_BED_REACTOR = 7;
-    }
-    oneof kind {
-      VesselFormType type = 1;
-      string custom = 2;
-    }
+  enum VesselType {
+    VESSEL_TYPE_UNSPECIFIED = 0;
+    ROUND_BOTTOM_FLASK = 1;
+    VIAL = 2;
+    WELL_PLATE = 3;
+    MICROWAVE_VIAL = 4;
+    TUBE = 5;
+    CONTINUOUS_STIRRED_TANK_REACTOR = 6;
+    PACKED_BED_REACTOR = 7;
   }
-  message VesselMaterial {
-    enum VesselMaterialType {
-      UNSPECIFIED = 0;
-      GLASS = 1;
-      POLYPROPYLENE = 2;
-      PLASTIC = 3;
-    }
-    oneof kind {
-      VesselMaterial type = 1;
-      string custom = 2;
-    }
+  oneof vessel_type {
+    VesselType type = 1;
+    string type_custom = 2;
   }
-  message VesselPrep {
-    enum VesselPrepType {
-      UNSPECIFIED = 0;
+  enum VesselMaterial {
+    VESSEL_MATERIAL_UNSPECIFIED = 0;
+    GLASS = 1;
+    POLYPROPYLENE = 2;
+    PLASTIC = 3;
+  }
+  oneof vessel_material {
+    VesselMaterial material = 3;
+    string material_custom = 4;
+  }
+  enum VesselPreparation {
+      VESSEL_PREPARATION_UNSPECIFIED = 0;
       NONE = 1;
       OVEN_DRIED = 2;
     }
-    oneof kind {
-      VesselPrepType type = 1;
-      string custom = 2;
-    }
+  oneof vessel_preparation {
+    VesselPreparation preparation = 5;
+    string preparation_custom = 6;
   }
-  VesselForm form = 1;
-  VesselMaterial material = 2;
-  VesselPrep prep = 3;
-  Volume vessel_volume = 4;  // Size (volume) of the vessel.
+  Volume vessel_volume = 7;  // Size (volume) of the vessel.
 }
 
 message ReactionSetup {
@@ -314,27 +305,24 @@ message ReactionConditions {
 }
 
 message TemperatureConditions {
-  message TemperatureControl {
-    enum TemperatureControlType {
-      UNSPECIFIED = 0;
-      AMBIENT = 1;
-      OIL_BATH = 2;
-      WATER_BATH = 3;
-      SAND_BATH = 4;
-      ICE_BATH = 5;
-      DRY_ALUMINUM_PLATE = 6;
-      MICROWAVE = 7;
-      DRY_ICE_BATH = 8;
-      AIR_FAN = 9;
-      LIQUID_NITROGEN = 10;
-    }
-    oneof kind {
-      TemperatureControlType type = 1;
-      string custom = 2;
-    }
+  enum TemperatureControlType {
+    UNSPECIFIED = 0;
+    AMBIENT = 1;
+    OIL_BATH = 2;
+    WATER_BATH = 3;
+    SAND_BATH = 4;
+    ICE_BATH = 5;
+    DRY_ALUMINUM_PLATE = 6;
+    MICROWAVE = 7;
+    DRY_ICE_BATH = 8;
+    AIR_FAN = 9;
+    LIQUID_NITROGEN = 10;
   }
-  TemperatureControl temperature_control = 1;
-  Temperature setpoint = 2;
+  oneof temperature_control_type {
+    TemperatureControlType type = 1;
+    string type_custom = 2;
+  }
+  Temperature setpoint = 3;
   message Measurement {
     enum MeasurementType {
       UNSPECIFIED = 0;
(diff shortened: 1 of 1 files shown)


proto/ord_schema.proto
```diff
@@ -197,6 +169,35 @@ message Compound {
...
+    REPURIFIED = 2;
+    // Compound sparged, most likely to be the case with solvents.
+    SPARGED = 3;
+    // Moisture removed, e.g., using molecular sieves.
+    DRIED = 4;
+    // Compound synthesized in-house
+    SYNTHESIZED = 5;
+  }
+  oneof kind {
+    PreparationType preparation = 1;
```
connorcoley: Will this lead to python code where we're accessing a ```preparation.preparation``` field? I might prefer ```type``` for the field name

proto/ord_schema.proto
```diff
@@ -40,15 +37,16 @@ message Reaction {
...
+    UNSPECIFIED = 0;
+    REACTION_SMILES = 1;
+    ATOM_MAPPED_SMILES = 2;
     RINCHI = 3;  // Reaction InChI.
     NAME = 4;  // Named reaction or reaction category.
   }
-  IdentifierType type = 1;
-  // Custom reaction identifier, used only when 'type' field is UNKNOWN.
-  string type_custom = 2;
+  oneof kind {
```
connorcoley: Not being that familiar with ```oneof```s, will this make it slightly more verbose to check or existence? Or make a json export of a proto less self-explanatory? I'll defer to your judgement for these cases. I do agree that ```type``` and ```custom_type``` (or ```custom```) are mutually exclusive, so maybe it is worth enforcing that

proto/ord_schema.proto
```diff
@@ -197,6 +169,35 @@ message Compound {
...
+    DRIED = 4;
+    // Compound synthesized in-house
+    SYNTHESIZED = 5;
+  }
+  oneof kind {
+    PreparationType preparation = 1;
+    string custom = 2;
+  }
+  // Full description of how the received compound was prepared.
+  string details = 3;
```
connorcoley: +1 for the normal non-custom use case where we only need to define, e.g.,

```
preparation.type = 'DRIED'
preparation.details = '4A molecular sieves'
```

proto/ord_schema.proto
```diff
@@ -237,52 +237,69 @@ message CompoundIdentifier {
...
+      GLASS = 1;
+      POLYPROPYLENE = 2;
+      PLASTIC = 3;
+    }
+    oneof kind {
+      VesselMaterial type = 1;
+      string custom = 2;
+    }
+  }
+  message VesselPrep {
```
connorcoley: "Prep" or "Preparation"? The latter is consistent with what we use for Compounds.

proto/ord_schema.proto
```diff
@@ -237,52 +237,69 @@ message CompoundIdentifier {
...
+      UNSPECIFIED = 0;
+      NONE = 1;
+      OVEN_DRIED = 2;
+    }
+    oneof kind {
+      VesselPrepType type = 1;
+      string custom = 2;
+    }
+  }
+  VesselForm form = 1;
```
connorcoley: Thinking about the use case here -- given a ```vessel```,

```
vessel.form.type = vessel.form.VIAL
vessel.material.type = vessel.material.GLASS
vessel.prep.custom = 'pre-cooled in fridge'
```

connorcoley: Having to go one level deeper into ```type``` isn't ideal. I realize that having separate (non-exclusive) ```vessel.material``` and ```vessel.material_custom``` feels less elegant, but do you think it might encourage people to use the non-custom field?

proto/ord_schema.proto
```diff
@@ -523,34 +567,30 @@ message ReactionProduct {
...
+      OIL = 3;
+    }
+    oneof kind {
+      TextureType type = 1;
+      string custom = 2;
+    }
   }
-  Texture isolated_texture = 13; 
-  string isolated_texture_custom = 14;
+  Texture texture = 8;
```
connorcoley: Another case where it seems like we've increased the level of nesting to ```outcome.texture.type```

proto/ord_schema.proto
```diff
@@ -783,3 +823,19 @@ message Percentage {
   // Precision of the measurement (with the same units as `value`).
   float precision = 2;
 }
+
+message Selectivity {
```
connorcoley: I think you forgot to keep a ```Selectivity selectivity``` field in the reaction outcome. I do like the idea of separating this out

proto/ord_schema.proto
```diff
@@ -783,3 +823,19 @@ message Percentage {
...
+  enum SelectivityType {
+    UNSPECIFIED = 0;
+    EE = 1;  // Enantiomeric excess as a percentage.
+    ER = 2;  // Enantiomeric ratio. (x:1)
+    DE = 3;  // Diasteromeric ratio (x:1)
+  }
+  oneof kind {
+    SelectivityType type = 1;
+    string custom = 2;
```
connorcoley: type_custom?

proto/ord_schema.proto
```diff
@@ -523,34 +567,30 @@ message ReactionProduct {
...
+  message Texture {
+    enum TextureType {
+      UNSPECIFIED = 0;
+      POWDER = 1;
+      CRYSTAL = 2;
+      OIL = 3;
+    }
+    oneof kind {
+      TextureType type = 1;
+      string custom = 2;
```
connorcoley: type_custom?

proto/ord_schema.proto
skearnes: oops, yes this should be `type`

proto/ord_schema.proto
skearnes: I also really like the enforcement of mutual exclusivity. Consider the following message:

```python
reaction = schema.Reaction()
reaction.identifiers.add(type='NAME', value='elephant toothpaste')
print(reaction)
>>> identifiers {
  type: NAME
  value: "elephant toothpaste"
}
```

* For JSON, the resulting dict will have only the `type` key, not `custom`.
* For SQL, both `type` and `custom` will show up as columns (and one will be NULL).
* For Python:
  ```python
  kind = reaction.identifiers[0].WhichOneof('kind')  # == 'type'`
  identifier_type = getattr(reaction.identifiers[0], kind)
  ```

proto/ord_schema.proto
skearnes: In general avoiding abbreviations is more style-compliant, so let's go with "Preparation".

skearnes: Changed:
* `Preparation` -> `CompoundPreparation`
* `VesselPrep{Type}` -> `VesselPreparation{Type}`

proto/ord_schema.proto
skearnes: Good catch; fixed.

proto/ord_schema.proto
skearnes: Yeah `Vessel` is the tricky one here since it has a bunch of different enums.

I think your suggestion of `type_custom` elsewhere is probably the right way to go; that will avoid needing more nesting for these cases. I agree that we should avoid the type of stutter you highlighted above. We'll need to bite the bullet and have some awkward prefixes for the unspecified enum values, but I think that's better than extra nesting (and the style guide agrees).

connorcoley: Is it worth pulling out VesselForm, VesselMaterial, and VesselPrep to be top-level messages so they can all use UNSPECIFIED?

proto/ord_schema.proto
```diff
@@ -297,104 +314,120 @@ message ReactionConditions {
...
+    }
   }
   TemperatureControl temperature_control = 1;
-  // Used to describe an UNKNOWN temperature control.
-  string temperature_control_custom = 2;
-  Temperature setpoint = 3;
+  Temperature setpoint = 2;
   message Measurement {
+    enum MeasurementType {
```
skearnes: I moved `MeasurementType` into `Measurement`, but I'm not sure I should have. This requires the type to be set for each measurement, which allows multiple types but could be a burden if we only expect one type. What do you think?

proto/ord_schema.proto
```diff
@@ -297,104 +314,120 @@ message ReactionConditions {
...
+    }
+    oneof kind {
+      AtmosphereType type = 1;
+      string custom = 2;
+    }
+  }
+  Atmosphere atmosphere = 3;
   message Measurement {
+    // TODO(ccoley) get input on how to expand this enum, among others
+    enum MeasurementType {
```
skearnes: Same here.

proto/ord_schema.proto
connorcoley: I like the JSON and SQL behavior, but that Python code hurts a little. In your example, would identifier_type be an integer if ```type``` is defined and a string if ```custom/type_custom``` is defined?

proto/ord_schema.proto
connorcoley: Keeping the type inside ```Measurement``` certainly increases our flexibility in case people do use multiple methods of measuring the temperature or pressure. Let's keep it inside

skearnes:
OK, now I understand why the style guide advocates prefixing vs. nesting for enums. I've reverted several of the extra nesting changes and used top-level `oneof`s instead. I think this will give us the mutual exclusivity we want without making it too hard to use the API.

Since some messages have multiple `enum`s and `oneof`s, I've changed the pattern to be that oneof fields are snake-cased enum names:

```proto
enum EnumName {
  ...
}
oneof enum_name {
  EnumName name = 1;
  string custom_name = 2;
}
```

PTAL

proto/ord_schema.proto
connorcoley: A thought about forced exclusivity here: one of the uses of the string field would be to elaborate on the enum choice, since the enums aren't going to allow much precision. For example, the VesselMaterial "plastic" could mean any number of plastics, so we would want to be able to capture any additional (string) information. So maybe these fields shouldn't actually be exclusive? Current draft ```oneof``` fields where we might want to allow additional details are:
- preparation_type (we have separate "details" field at the moment)
- vessel_type
- vessel_material
- vessel_preparation
- temperature_control_type
- measurement_type
- pressure_control_type
- atmosphere_type (additional detail for OXYGEN might be the moisture level or grade, for example)
- etc.

proto/ord_schema.proto
skearnes: I'd say no; that will require the type of nesting you highlighted above and people aren't going to be using the default enum value anyway---it will only be a check in the validation script.

proto/ord_schema.proto
skearnes: Fixed in latest commit.

proto/ord_schema.proto
skearnes: Done.

proto/ord_schema.proto
skearnes: Interesting. I like the idea of being able to provide more details, but I also like that the oneof forces users to be explicit about their values. If we removed the exclusivity, I'd be worried that people might put in details about the glass they used without specifying GLASS as the material---we'd have no automatable way to check. Am I being too paranoid?

Would you support adding a `details` field to all of these like we do for preparation_type?

proto/ord_schema.proto
skearnes: (note that this might change if we decide to add a details field; I think factoring out as a separate message makes sense if you need to carry more than just type/type_custom)

proto/ord_schema.proto
connorcoley: I don't know if you're being too paranoid. I think that we could successfully encourage users to use the enums even if additional information is provided in the custom/details field.

I like the cleanliness of the ```details``` field for the ```Preparation``` message. I'm worried that adding a string field for details will become somewhat unwieldy, especially where we would need prefixes.

skearnes: What about something like this?

```proto
enum VesselMaterial {
  VESSEL_MATERIAL_UNSPECIFIED = 0;
  VESSEL_MATERIAL_CUSTOM = 1;
  GLASS = 2;
  POLYPROPYLENE = 3;
  PLASTIC = 4;
}
VesselMaterial material = 1;
string material_details = 2;
```

Then it's explicit that CUSTOM is being used and we can required material_details when that option is chosen.

(It might make sense to factor out a separate message to avoid the enum prefixes if we do this.)

connorcoley: Would you propose we add a CUSTOM to all of the enums where we might want additional details?

skearnes: I think I'd do this for all the enums where a custom value is supported, and change `custom` to `details` everywhere. For enums that don't have a custom field but we might want details, I'd add a details field. So at the end every enum would have an accompanying details field, and some of them would have a CUSTOM option in the enum.

connorcoley: That sounds reasonable to me. Also alleviates my ```oneof``` anxiety :)

skearnes: OK, great. I'll add another comment when I've updated.

skearnes:
PTAL; I made one important change that I really like:

Many enums are now wrapped inside separate messages, but their values are accessed as `parent.child` to make the API simpler. This has the additional benefit of scoping the enum values, so now instead of e.g. `Vessel.VIAL` we have `Vessel.VesselType.VIAL` --- and the value is still created with just `Vessel(type='VIAL')`. The best of both worlds!

connorcoley:
+1 to wrapping enums in messages for scoping. It makes the schema uglier but I think the compiled code will look nicer based on your ```VesselType``` example.

- Should ```CompoundPreparation```'s ```type_details``` just be ```details``` since it's already scoped insight the ```CompoundPreparation``` message? Likewise, ```ReactionAnalysis.type_details``` could be ```ReactionAnalysis.details``` and ```ReactionWorkup.type_details``` could be ```ReactionWorkup.details```. 

- ```CompoundIdentifier.type_details``` and ```ReactionIdentifier.type_details``` are probably unnecessary. There may be cases where someone wants to use a custom identifier type, but I don't see a need to provide more details for any of these options. We can keep it for consistency if we want to make a global note that ```details``` is also to be used with ```CUSTOM```.

Thoughts?

skearnes:
Yes, I was hoping `details` could also serve as the place where people describe what CUSTOM means. I mention this in the updated top-level comment block in the file.

skearnes:
For `type_details`: I changed all occurrences of `type_details` to just `details` and left the other `<field>_details` (e.g. `material_details`) as-is. We can adjust later if needed.

Issue #16, opened 2020-02-25
Potential duplicate information in Analysis and yield


connorcoley:
One ```analysis``` entry might be present as part of a ReactionAnalysis but could also be what was used to quantify yield. Right now, the yield analysis method is cross-referenced with the ReactionAnalysis enum, but it's missing all of the extra fields (e.g., a string field with additional details, raw spectra, etc.). It's almost like we want to cross-reference how the yield was quantified with a full ReactionAnalysis, not just the type.

skearnes:
I agree; I don't really like the `ReactionAnalysis.AnalysisType` and it seems like a raw `ReactionAnalysis` would be more complete.

skearnes:
maybe rename to `yield_analysis`?

connorcoley:
If we convert to a ```ReactionAnalysis yield_analysis``` then, are we now being inconsistent w.r.t. ```conversion``` since that doesn't get linked to a full analysis? 

One analysis can produce multiple readouts (yields of multiple products, conversion, etc.). 

It's almost like product yields/identities should theoretically be a sub-field within an analysis, but I don't like the idea of burying information so deeply

skearnes:
Maybe we should remove `yield_type` entirely and just push these details into `ReactionOutcome.analyses`? That would remove the extra level of nesting.

connorcoley:
I'll have to think about this more. In the case where there are multiple analyses, it's useful to know how the experimentalist arrived at the yield. Each experiment might have a crude HPLC and NMR, but the yield could still be an isolated yield.

skearnes:
I think I'm still leaning towards adding a full `Analysis` to `ReactionOutcome`. This would avoid the awkward nested reference to `ReactionAnalysis.AnalysisType`.

skearnes:
(It would also make #38 a bit cleaner.)

connorcoley:
I don't think we can universally say that an Analysis should be a subfield of ReactionOutcome though. A reaction can have multiple analyses, where one or more were used to identify the product(s) or product yield(s) and selectivity(s). One Analysis could be used to quantify multiple products, and I don't think we would want to encourage recording the same Analysis for multiple Outcomes.

skearnes:
I think I mistyped my last comment; I agree that Analysis shouldn't be owned by ReactionOutcome. I think I meant to say that I support adding an Analysis field to ReactionProduct instead of just AnalysisType.

connorcoley:
[...]

PR #42, opened 2020-03-11
Analysis map plus

connorcoley:
This might address issue #16 using internal cross-referencing. Analysis data is defined in a single place with no redundancies. Multiple values of multiple products can be linked back to the same ReactionAnalysis message. 

This allows the rapid checking of analysis type we desire, simply by looking at ```analyses[product.analysis_yield].type```

- Includes conversion of analyses to map to re-use keys for different products
- Changes reaction inputs to a map from a repeated ReactionInput to allow easier templating (i.e., so we don't have to keep track of the order of inputs in the definition)
- Simplifies a few vessel field names

### DIFF
--- a/proto/ord_schema.proto
+++ b/proto/ord_schema.proto
@@ -308,21 +308,23 @@ message ReactionConditions {
 }
 
 message TemperatureConditions {
-  enum TemperatureControl {
-    UNSPECIFIED = 0;
-    CUSTOM = 1;
-    AMBIENT = 2;
-    OIL_BATH = 3;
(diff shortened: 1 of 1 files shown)


connorcoley:
Questions:
- Should the ```products``` in a ```ReactionOutcome``` also be a map for easier templating? (I think this is not necessary)
- Should the ```components``` in a ```ReactionInput``` be a map? (I think no, even though defining "solute" and "solvent" keys might be nice for solutions, it would be annoying to need a string key for pure species)

PR #45, opened 2020-03-15
Filling out unit resolver

connorcoley:


connorcoley:
Note: commented out ```Concentration``` because we do not actually use that at the moment. If we want to introduce it, will have to deal with ambiguity between millimeter and millimolar (mm v. mM) since our comparison is lowercase.

skearnes:
> Note: commented out `Concentration` because we do not actually use that at the moment. If we want to introduce it, will have to deal with ambiguity between millimeter and millimolar (mm v. mM) since our comparison is lowercase.

OK, that's fine for now. I think we can adjust the resolver to be aware of the required type of the unit message and deal with this appropriately, but we can save that for another PR.

PR #51, opened 2020-03-26
Progress toward analysis Methods and Data

connorcoley:
- Creating several messages to define structured analytical data
  and methods in separate analysis.proto
- Moved Compound message and unit messages to separate files for
  import so they can be used in analysis.proto
- Still working out the appropriate fields and level of detail for
  several methods; will need extensive review by the committee

### DIFF
--- a/ord_schema/units.py
+++ b/ord_schema/units.py
@@ -67,6 +67,7 @@ _UNIT_SYNONYMS = {
         schema.FlowRate.MILLILITER_PER_MINUTE: ['mL/min'],
         schema.FlowRate.MILLILITER_PER_SECOND: ['mL/s'],
         schema.FlowRate.MICROLITER_PER_HOUR: ['uL/h'],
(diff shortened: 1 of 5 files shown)


connorcoley:
@skearnes would you be able to provide a preliminary review of this message structure? I haven't been able to think of a way to avoid an absurd number of messages if we want to be able to capture highly structured data. In particular, see the ```data_kind``` and ```method_kind``` oneofs in ```ord_schema.proto```.

Messages _within_ ```analysis.proto``` aren't yet in a logical order

michaelmaser:
Realizing I haven't seen 2D NMR methods (HSQC, HMBC, NOESY, COSY, ROESY, TOCSY). MethodNMR is all nuclei for 1D techniques at the moment. Suppose could be a separate message (e.g. Method2DNMR), but being a part of "enum NMRType" could work too. Would also need nuclei specifications.

PR #53, opened 2020-03-27
Validation scripts

connorcoley:
- Every message has its own ```ValidateMESSAGE()``` function
- The generic ```ValidateMessage``` will determine the appropriate validation function based on the message type
- The generic ```ValidateMessage``` will also default to recursing through the message (PTAL at the logic used for recursion; I think it's good but there may be unforeseen edge cases)
- Compound identities/names are not checked for validity to avoid RDKit dependency for now
- Common requirements like int/float ranges and nonnegativity use helper functions
- If a value is defined for a unit'd message, we require units to be specified
- If a CUSTOM type is defined for an enum, we require the details to be specified
- In product characterization, the keys of analyses that led to certain results (e.g., ```analysis_yield```) must match defined analyses in the outcome. This actually helped me catch a typo in one of our Colab examples where ```'chiral HPLC'``` was used instead of ```'chiral SFC'```.
- DateTimes are now parsed by ```dateutil.parser.parse()```. In the schema, a DateTime message is simple a wrapper for a string value.
- DateTimes of ```experiment_start```, ```record_created```, and ```record_modified``` must be in a logical order
- ORCIDs must match expected format (though we do not use the checksum character in a meaningful way)

- The code is written so that in the future, we would be able to have the validation steps modify values within the messages (i.e., using assignments and ```CopyFrom```s. For example, canonicalize a SMILES or add a SMILES identifier based on a NAME identifier. This has not been tested.

### DIFF
--- a/ord_schema/validations.py
+++ b/ord_schema/validations.py
@@ -6,7 +6,8 @@ import re
 import math
 import dateutil.parser
 
-def ValidateMessage(message, recurse=True):
+
+def validate_message(message, recurse=True):
     """Template function for validating custom messages in the schema.
 
     Messages are not validated to check enum values, since these are enforced 
@@ -30,83 +31,31 @@ def ValidateMessage(message, recurse=True):
 
     # Recurse through submessages
     if recurse:
-        for field in message.DESCRIPTOR.fields:
+        for field, value in message.ListFields():
             if field.type == field.TYPE_MESSAGE: # need to recurse
                 if field.label == field.LABEL_REPEATED:
                     if field.message_type.GetOptions().map_entry: # map
-                        if field.message_type.fields[1].type == field.TYPE_MESSAGE: # value is message
-                            for key,submessage in getattr(message, field.name).items():
+                        if field.message_type.fields_by_name['value'].type == field.TYPE_MESSAGE: # value is message
+                            for key,submessage in value.items():
                                 submessage.CopyFrom(
-                                    ValidateMessage(submessage)
+                                    validate_message(submessage)
                                 )
                         else: # value is a primitive
                             pass
                     else: # Just a repeated message
-                        for submessage in getattr(message, field.name):
+                        for submessage in value:
                             submessage.CopyFrom(
-                                ValidateMessage(submessage)
+                                validate_message(submessage)
                             )
                 else: # no recursion needed
-                    submessage = getattr(message, field.name)
+                    submessage = value
                     submessage.CopyFrom(
-                        ValidateMessage(submessage)
+                        validate_message(submessage)
                     )
 
     # Message-specific validation
     try:
-        return {
-            schema.Reaction: ValidateReaction,
-            # Basics
-            schema.ReactionIdentifier: ValidateReactionIdentifier,
-            schema.ReactionInput: ValidateReactionInput,
-            # Compounds
-            schema.Compound: ValidateCompound,
-            schema.Compound.Feature: ValidateCompoundFeature,
-            schema.CompoundPreparation: ValidateCompoundPreparation,
-            schema.CompoundIdentifier: ValidateCompoundIdentifier,
-            # Setup
-            schema.Vessel: ValidateVessel,
-            schema.ReactionSetup: ValidateReactionSetup,
-            # Conditions
-            schema.ReactionConditions: ValidateReactionConditions,
-            schema.TemperatureConditions: ValidateTemperatureConditions,
-            schema.TemperatureConditions.Measurement: ValidateTemperatureMeasurement,
-            schema.PressureConditions: ValidatePressureConditions,
-            schema.PressureConditions.Measurement: ValidatePressureMeasurement,
-            schema.StirringConditions: ValidateStirringConditions,
-            schema.IlluminationConditions: ValidateIlluminationConditions,
-            schema.ElectrochemistryConditions: ValidateElectrochemistryConditions,
-            schema.ElectrochemistryConditions.Measurement: ValidateElectrochemistryMeasurement,
-            schema.FlowConditions: ValidateFlowConditions,
-            schema.FlowConditions.Tubing: ValidateTubing,
-            # Annotations
-            schema.ReactionNotes: ValidateReactionNotes,
-            schema.ReactionObservation: ValidateReactionObservation,
-            # Outcome
-            schema.ReactionWorkup: ValidateReactionWorkup,
-            schema.ReactionOutcome: ValidateReactionOutcome,
-            schema.ReactionProduct: ValidateReactionProduct,
-            schema.Selectivity: ValidateSelectivity,
-            schema.DateTime: ValidateDateTime,
-            schema.ReactionAnalysis: ValidateReactionAnalysis,
-            # Metadata
-            schema.ReactionProvenance: ValidateReactionProvenance,
-            schema.Person: ValidatePerson,
-            # Units
-            schema.Time: ValidateTime,
-            schema.Mass: ValidateMass,
-            schema.Moles: ValidateMoles,
-            schema.Volume: ValidateVolume,
-            schema.Concentration: ValidateConcentration,
-            schema.Pressure: ValidatePressure,
-            schema.Temperature: ValidateTemperature,
-            schema.Current: ValidateCurrent,
-            schema.Voltage: ValidateVoltage,
-            schema.Length: ValidateLength,
-            schema.Wavelength: ValidateWavelength,
-            schema.FlowRate: ValidateFlowRate,
-            schema.Percentage: ValidatePercentage,
-        }[type(message)](message)
+        return _VALIDATOR_SWITCH[type(message)](message)
     except KeyError:
         # NOTE(ccoley): I made the conscious decision to raise an error here,
         # rather than assume that the message is valid. If a message does not
@@ -115,15 +64,8 @@ def ValidateMessage(message, recurse=True):
         # us to think about what is necessary if/when new messages are added.
         raise NotImplementedError(f"Don't know how to validate {type(message)}")
 
-
 class ValidationWarning(Warning):
     pass
-
-def return_message_if_valid(func):
-    def wrapper(message):
-        func(message)
-        return message
-    return wrapper
 
 def ensure_float_nonnegative(message, field):
     if getattr(message, field) < 0:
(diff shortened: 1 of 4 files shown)


connorcoley:
@skearnes I don't know what the proper way to install ```dateutil``` is in the continuous integration pipeline

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+            f'{type(message).DESCRIPTOR.name} must be non-negative')
+
+def ensure_float_range(message, field, min, max):
+    if getattr(message, field) < min or getattr(message, field) > max:
+        raise ValueError(f'Field {field} of message '\
+            f'{type(message).DESCRIPTOR.name} must be between {min} and {max}')
+
+def ensure_units_specified_if_value_defined(message):
+    if message.value: # can't distinguish 0 and unspecified anyway
+        if message.units == message.UNSPECIFIED:
```
skearnes: You can combine these with an `and` to avoid the extra level of indentation.

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+            raise ValueError(f'Unspecified units for {type(message)} with ' \
+                f'value defined ({message.value})')
+
+def ensure_details_specified_if_type_custom(message):
+    if message.type == message.CUSTOM:
```
skearnes: Same here.

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+        # require any message-level checks (not uncommon), then it should still
+        # be listed in the dictionary switch above withpass. This will force
+        # us to think about what is necessary if/when new messages are added.
+        raise NotImplementedError(f"Don't know how to validate {type(message)}")
+
+
+class ValidationWarning(Warning):
+    pass
+
+def return_message_if_valid(func):
```
skearnes: This seems like an unnecessary redirection; why not just have each function return the message?

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+                f'value defined ({message.value})')
+
+def ensure_details_specified_if_type_custom(message):
+    if message.type == message.CUSTOM:
+        if not message.details:
+            raise ValueError(f'Custom type defined for {type(message)}, ' \
+                'but details field is empty')
+
+@return_message_if_valid
+def ValidateReaction(message):
```
skearnes: Please use snake_case for function names; http://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+
+@return_message_if_valid
+def ValidateCompoundIdentifier(message):
+    ensure_details_specified_if_type_custom(message)
+    # TODO(ccoley): Add identifier-specific validation, e.g., by using
+    # RDKit to try to parse SMILES, looking up NAMEs using online resolvers
+
+@return_message_if_valid
+def ValidateVessel(message):
+    if message.type == message.VesselType.CUSTOM:
```
skearnes: Please combine these `if` statements. (I won't comment on every one.)

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+            for key in getattr(product, field):
+                if key not in analysis_keys:
+                    raise ValueError(f'Undefined analysis key {key} '\
+                        'in ReactionProduct')
+
+@return_message_if_valid
+def ValidateReactionProduct(message):
+    if message.texture == message.Texture.CUSTOM:
+        if not message.texture_details:
+            raise ValueError(f'Custom texture defined for {type(message)}, ' \
```
skearnes: Note that the backslash is not required; the parentheses do the job.

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+                'but texture_details field is empty')
+
+@return_message_if_valid
+def ValidateSelectivity(message):
+    ensure_float_nonnegative(message, 'precision')
+    if message.type == message.EE:
+        ensure_float_range(message, 'value', 0, 100)
+        if message.value > 0 and message.value < 1:
```
skearnes: Optional: you can do `if 0 < message.value < 1`

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+@return_message_if_valid
+def ValidatePressure(message):
+    ensure_float_nonnegative(message, 'value')
+    ensure_float_nonnegative(message, 'precision')
+    ensure_units_specified_if_value_defined(message)
+
+@return_message_if_valid
+def ValidateTemperature(message):
+    if message.units == message.CELSIUS:
+        ensure_float_range(message, 'value', -273.15, inf)
```
skearnes: Optional: you could have `ensure_float_range` include defaults for `min` and `max` so you don't need to specify both.

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+    pass
+
+def return_message_if_valid(func):
+    def wrapper(message):
+        func(message)
+        return message
+    return wrapper
+
+def ensure_float_nonnegative(message, field):
+    if getattr(message, field) < 0:
```
skearnes: Optional (nit): you could define this with a call to `ensure_float_range`.

ord_schema/validations_test.py
```diff
@@ -0,0 +1,86 @@
...
+
+from ord_schema import validations
+from ord_schema.proto import ord_schema_pb2 as schema
+
+
+class UnitsTest(parameterized.TestCase, absltest.TestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.validations = validations
```
skearnes: Just use `validations` directly in the methods; not `self.validations`.

ord_schema/validations_test.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def test_reaction(self):
+        message = schema.Reaction()
+        with self.assertRaisesRegex((ValueError), 'reaction input'):
+            self.validations.ValidateReaction(message)
+
+    def test_reaction_recursive(self):
+        message = schema.Reaction()
+        with self.assertRaisesRegex((ValueError), 'reaction input'):
+            self.validations.ValidateMessage(message, recurse=False)
```
skearnes: In what situations would we want `recurse=False`? I think you'd get the same result here if `recurse` was True?

ord_schema/validations.py
connorcoley: Thought about that, but I would find an error message saying "non-negative" incrementally more informative than "between 0 and inf"

ord_schema/validations_test.py
connorcoley: I'm not sure why I thought that was the right thing to do

ord_schema/validations_test.py
connorcoley: In that case you would get the same error yes. I revised this case to show the behavior I actually wanted to test:

```
message = schema.Reaction()
with self.assertRaisesRegex((ValueError), 'reaction input'):
    validations.ValidateMessage(message)
dummy_input = message.inputs['dummy_input']
self.assertEqual(validations.ValidateMessage(message, recurse=False), 
    message)
```

Once one input is defined, the validation (with ```recurse=False```) is happy and doesn't mind that the input has no components.

I don't know in what situation we would actually choose to use ```recurse=False```, but I wanted to have that flexibility in case it is useful to do on-the-fly validation for the web form just at that level.

ord_schema/validations_test.py
```diff
@@ -0,0 +1,86 @@
+"""Tests for ord_schema.units."""
+
+from absl.testing import absltest
+from absl.testing import parameterized
+
+from ord_schema import validations
+from ord_schema.proto import ord_schema_pb2 as schema
+
+
+class UnitsTest(parameterized.TestCase, absltest.TestCase):
```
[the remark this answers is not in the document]
connorcoley: Oh dear, I've been caught blatantly copying your ```units_test.py```

ord_schema/validations.py
connorcoley: For some reason I didn't think Python supported that, thanks

connorcoley:
Your style comments are much appreciated. One early choice I made was to have these be functions and not methods within a ```Validator``` class since there would be nothing to do in the ```__init__``` and no real need for persistence. Would you rather see all of this wrapped inside a class?

ord_schema/validations.py
```diff
@@ -3,26 +3,30 @@
 from ord_schema.proto import ord_schema_pb2 as schema
 
 import re
-from math import inf
-from dateutil import parser as dateparser
+import math
+import dateutil.parser
```
skearnes: Note that you can still do `from dateutil import parser`; I think `as dateparser` or `as dateutil_parser` would also be fine for clarity.

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+        Raises:
+            ValueError: If any fields are invalid."""
+
+    # Recurse through submessages
+    if recurse:
+        for field in message.DESCRIPTOR.fields:
+            if field.type == field.TYPE_MESSAGE: # need to recurse
+                if field.label == field.LABEL_REPEATED:
+                    if field.message_type.GetOptions().map_entry: # map
+                        if field.message_type.fields[1].type == field.TYPE_MESSAGE: # value is message
```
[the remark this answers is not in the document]
skearnes: I'd still like this change.

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+                else: # no recursion needed
+                    submessage = getattr(message, field.name)
+                    submessage.CopyFrom(
+                        ValidateMessage(submessage)
+                    )
+
+    # Message-specific validation
+    try:
+        return {
+            schema.Reaction: ValidateReaction,
```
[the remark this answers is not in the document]
skearnes: I don't have a strong preference about class methods vs. free functions here. I think free functions is fine, but please do move this dict out as a constant.

ord_schema/validations.py
skearnes: I'd still like this change.

ord_schema/validations.py
skearnes: SGTM

ord_schema/validations.py
skearnes: I'd still like this change.

ord_schema/validations_test.py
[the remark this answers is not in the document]
skearnes: Good, that's what it's there for :)

ord_schema/validations.py
```diff
@@ -0,0 +1,428 @@
...
+        Returns:
+            The input message, with any unambiguous changes made as
+            needed to ensure validity.
+
+        Raises:
+            ValueError: If any fields are invalid."""
+
+    # Recurse through submessages
+    if recurse:
+        for field in message.DESCRIPTOR.fields:
```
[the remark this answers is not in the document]
connorcoley: Thanks -- good call. I missed this function

ord_schema/validations.py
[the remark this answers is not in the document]
connorcoley: Didn't see this comment for some reason, sorry

ord_schema/validations.py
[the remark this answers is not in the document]
connorcoley: Also didn't see this original comment -- I wonder if I loaded the page partway through your adding comments.

ord_schema/validations.py
connorcoley: It was more aesthetically pleasing to have the same wrapper preceding every function instead of having ```return message``` at the end of each one. But changing now

ord_schema/validations.py
connorcoley: Is it worth combining even if it makes the single if statement multiline?

ord_schema/validations.py
connorcoley: Also missed this original comment for some reason.

ord_schema/validations.py
```diff
@@ -119,156 +70,170 @@ def ValidateMessage(message, recurse=True):
...
     ensure_details_specified_if_type_custom(message)
     # TODO(ccoley): Add identifier-specific validation, e.g., by using
     # RDKit to try to parse SMILES, looking up NAMEs using online resolvers
+    return message
+
+
+def validate_vessel(message):
+    if message.type == message.VesselType.CUSTOM and not message.details:
+        raise ValueError('VesselType custom, but no details provided')
+    if message.material == message.VesselMaterial.CUSTOM and \
```
skearnes: Prefer parentheses to line breaks:

```
if (message.material == message.VesselMaterial.CUSTOM and
    not message.material_details):
```

PR #56, opened 2020-03-31
Add RecordEvent message for provenance

### LINKED ISSUE #44: Provenance
We need to keep track of enough information to handle (at least) the following cases:

- Person A ran an experiment and Person B created the proto describing it.
- Person C made a mistake while creating a record and submits a correction.

Keeping pointers to the original submission artifacts will cover many of these cases, but we probably need to expand the `Provenance` message to contain these pointers and additional information about the submitter/creator (not just the experimenter). The guiding principle, I think, is "do we know who we should talk to if there's a problem?"

Related to this is the question of duplicates (we may just want to ignore these?):

- Person E and Person F both extracted the same reaction from a publication (possibly with slightly different details or even enum values).

skearnes:
Fixes #44 

Also changes `id` to `record_id` and makes it a string field for flexibility (people shouldn't be typing record IDs by hand anyway).

### DIFF
--- a/ord_schema/validations.py
+++ b/ord_schema/validations.py
@@ -286,25 +286,35 @@ def validate_reaction_analysis(message):
 
 def validate_reaction_provenance(message):
     # Prepare datetimes
+    # TODO(kearnes): Require these to be set?
+    experiment_start = None
+    record_created = None
+    record_modified = None
     if message.experiment_start.value:
-        experiment_start = parser.parse(
-            message.experiment_start.value)
-    if message.record_created.value:
-        record_created = parser.parse(message.record_created.value)
-    if message.record_modified.value:
-        record_modified = parser.parse(message.record_created.value)
+        experiment_start = parser.parse(message.experiment_start.value)
+    if message.record_created.time:
+        record_created = parser.parse(message.record_created.time.value)
+    for record in message.record_modified:
+        # Use the last record as the most recent modification time.
+        record_modified = parser.parse(record.time.value)
     # Check if record_created undefined
-    if message.record_modified.value and not message.record_created.value:
+    if record_modified and not record_created:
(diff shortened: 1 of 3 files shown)


ord_schema/validations.py
```diff
@@ -286,28 +286,38 @@ def validate_reaction_analysis(message):
...
+    if record_modified and record_created:
         if (record_modified - record_created).total_seconds() < 0:
             raise ValueError('Record modified time should be after creation')
     # TODO(ccoley) could check if publication_url is valid, etc.
     return message
 
 
+def validate_record_event(message):
+    if not message.time.value:
+        raise ValueError('RecordEvent must have `time` specified')
```
connorcoley: To confirm, the original intent of these validations was to enforce things that the schema on its own cannot. Even though we could use the ```required``` keyword, this is still the preferred solution?

skearnes: I did some more reading on this as I reconsidered. It turns out that `required` was actually *removed* by proto3, so it's not an option at all. See https://github.com/protocolbuffers/protobuf/issues/2497#issuecomment-267422550

skearnes:
> For uniformity, would it make sense to also make `experiment_start` used this new `RecordEvent` message?
> 
> We have both a `Person` and `DateTime` associated with the provenance:
> 
> ```
> Person experimenter = 1;
> DateTime experiment_start = 3;
> ```
> 
> I don't necessarily believe this is a good idea, but I wanted to bring it up as something to consider if you hadn't already. I think I prefer the current solution of having the experimenter and start time being direct fields within `ReactionProvenance`.

I considered this and also preferred the direct field approach.

skearnes:
Also, here's another link describing why `required` was discouraged for proto2: https://stackoverflow.com/a/31814967

PR #60, opened 2020-04-02
Add BinaryData message

skearnes:
I wanted to include more metadata for `bytes` values, so I created a `BinaryData` message and replaced maps with repeated fields. 

Happy to keep using maps if we also want "names" for these messages...

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -458,7 +458,7 @@ name-group=
 
 # Regular expression which should only match function or class names that do
 # not require a docstring.
-no-docstring-rgx=^_|^test|.*Test$|^main$|setUp
+no-docstring-rgx=^_|^test|.*Test$|^main$|^setUp$
 
 # List of decorators that produce properties, such as abc.abstractproperty. Add
 # to this list to register other decorators that produce valid properties.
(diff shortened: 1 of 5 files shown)


connorcoley:
A nice aspect of the map is that we are consistent with the convention that map string keys just need to be internally consistent and don't carry outside meaning (i.e., like the names of analyses and the names of reaction inputs).

Pros:
- Having a ```description``` field

Cons:
- Another level of nested messages
- Implies a greater significance of the ```filename``` field when we really just care about the file type / extension. 

Neurals?
- Fewer map fields in the schema

How would you feel about using a ```map<string, BinaryData>``` where the ```BinaryData``` message only contains ```bytes``` and ```description``` fields? We could have an optional ```filetype``` field that could be used if the map key doesn't use a known extension.

skearnes:
> How would you feel about using a `map<string, BinaryData>` where the `BinaryData` message only contains `bytes` and `description` fields? We could have an optional `filetype` field that could be used if the map key doesn't use a known extension.

Great idea. PTAL.

connorcoley:
I approved this PR, but would you be able to add a validator for this new message type? Every time we create a new message in the schema we need to add it to the validations. Perhaps all we need to do is check that ```value``` is defined

skearnes:
> I approved this PR, but would you be able to add a validator for this new message type? Every time we create a new message in the schema we need to add it to the validations. Perhaps all we need to do is check that `value` is defined

Thanks for the reminder! Done.

Issue #58, opened 2020-04-01
For example datasets, provide example preprocessing steps


connorcoley:
We should write additional Python scripts to show the workflow of processing a set of examples and preparing a simple pandas dataframe. We can do this for specific HTE sets to prepare them for machine learning. In the future, we can encourage batches of HTE data to be associated with project-specific python scripts for preprocessing.

PR #62, opened 2020-04-06
Add Dataset message

skearnes:
Related to #58 

What do you think? I can add tests, etc. after we decide on an approach.


connorcoley:
Is it important that we have these datasets and associated pre-processing code as part of the schema? I kind of like having Reactions be the highest-level message type, i.e., every proto in the schema is a Reaction. I agree these are important fields to capture for defining grouped datasets and processing scripts.

skearnes:
> Is it important that we have these datasets and associated pre-processing code as part of the schema? I kind of like having Reactions be the highest-level message type, i.e., every proto in the schema is a Reaction. I agree these are important fields to capture for defining grouped datasets and processing scripts.

Not necessarily. The thing I like about this is that all the data for a dataset is in a structured form. We could move this to another `.proto` file to avoid any confusion with the other messages related to `Reaction`.

connorcoley:
I would support having it in a separate ```.proto```. To me, this is starting to add more functionality and metadata (metametadata?) like you would want for a centralized website. Perhaps Datasets need their own Provenance.

skearnes:
PTAL; we can add more functionality/tests once we have specific applications.

skearnes:
We can keep this open until we have more of those things :)

Note that the type of `record_ids` matches the type of `ReactionProvenance.record_id`.

skearnes:
(Closing for now.)

skearnes:
Reopening based on discussion in #109; the major change is that now the `Dataset` message stores the actual `Reaction` messages, not just their IDs.

skearnes:
(Working on some tests now.)

skearnes:
> Some questions:
> 
> 1. Is a `map` for reactions better than just a repeated field? Is there a significant benefit to having the reaction's identifier be required as a key rather than iterating through the list (if needed)? From a contributor's point of view, it's going to be very annoying to come up with placeholder strings as keys and use the dictionary-type definition rather than a list-type `append`.

I fought with myself on this, but I think the user experience is more important that the fast lookup, which could be written somewhere else if needed for a downstream application. I'll change this to a repeated field.

> 2. Are we comfortable with the restriction that a single `Dataset` cannot contain reactions from multiple DOIs?

I was actually imagining that this field would contain a DOI for this dataset, one that we assign. This could be pointed to in publications, etc. Not that it matches the source of any Reaction data taken from the literature.

> 3. What is the benefit of having the `url` link to the official version of the dataset in the database? Who/what assigns the `version`?
> 

I thought this might be useful for quickly finding where a dataset came from in the official repo. I imagined that the version would be incremented by an automated script every time any of the contained reactions is modified. But I'll drop these for now until we can discuss further.

> P.S. I'm also imagining a use case that will make you cringe: running `exec` on function definitions contained in `scripts`

To check that they run? I could probably be persuaded if we can find a safe way to do this...

PR #65, opened 2020-04-08
Add RDKIT_BINARY identifier types

skearnes:
Parsing molecules (and presumably reactions) from RDKit's binary format is *much* faster than re-parsing the SMILES each time. In the live version of the database, I expect we will want binary representations of compounds and reactions to enable faster searching/matching. 

Note that I punted on the question of adding RDKit as a dep here (I think it's a good idea, but it will take some work on the CI side to get RDKit installed when running tests for PRs); we can address that in another PR.

### DIFF
--- a/ord_schema/proto/reaction_pb2_test.py
+++ b/ord_schema/proto/reaction_pb2_test.py
@@ -1,6 +1,10 @@
 """Tests for ord_schema.proto.reaction_pb2."""
 
 from absl.testing import absltest
+try:
+    from rdkit import Chem
+except ImportError:
+    Chem = None
 
 from ord_schema.proto import reaction_pb2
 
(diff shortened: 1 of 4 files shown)


connorcoley:
This makes sense to me. I've never used RDKit's binary format for saving/loading. I also like the idea of having our validation pipeline automatically add a compound identifier of type ```RDKIT_BINARY``` if it is able to successfully resolve the molecule (from SMILES or otherwise).

If we include RDKit as a dependency here, which I think is inevitable, it might be worth asking @n8kim1 to help containerize the entire project with Docker, rather than just the interface portion. We would want to preserve the option of using the schema w/o Docker using a separate RDKit installation though.

skearnes:
> This makes sense to me. I've never used RDKit's binary format for saving/loading. I also like the idea of having our validation pipeline automatically add a compound identifier of type `RDKIT_BINARY` if it is able to successfully resolve the molecule (from SMILES or otherwise).
> 
Yes, exactly!

> If we include RDKit as a dependency here, which I think is inevitable, it might be worth asking @n8kim1 to help containerize the entire project with Docker, rather than just the interface portion. We would want to preserve the option of using the schema w/o Docker using a separate RDKit installation though.

I think the simplest thing to do here would be to make sure that we have access to a docker image with RDKit pre-installed. Then you can simply clone this repo and do whatever you'd like. And of course we could create an image that has the ord_schema python package pre-installed as well. I can also experiment with the RDKit docker image stuff as part of the continuous integration support.

PR #69, opened 2020-04-09
Using warnings module for ValidationWarnings

connorcoley:
- ```raise``` interrupts the program like an Exception
- ```validate_binary_data``` was missing ```return message```

### DIFF
--- a/examples/example_template.ipynb
+++ b/examples/example_template.ipynb
@@ -20,14 +20,14 @@
     "id": "UuYMhx3e-xlw",
     "badge": true,
     "repo_name": "Open-Reaction-Database/ord-schema",
-    "branch": "skearnes-patch-3",
+    "branch": "binarydata-validation",
     "nb_path": "examples/example_template.ipynb",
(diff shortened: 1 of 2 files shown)


skearnes:
I was thinking that we wanted validation errors to be fatal? I'm not aware of a way to catch warnings; I'd suggest putting a try/except block in `validate_message` and adding a boolean argument `errors_are_fatal`.

connorcoley:
I agree that errors should be fatal, but I think that we should also have non-fatal warnings. For example, with a ```Percentage``` message with ```value=0.5``. You _could_ have 0.5% yield legitimately, but there is a high chance that they means 0.5 = 50% yield. 

There are ways to change the behavior of warnings (see [here](https://docs.python.org/3/library/warnings.html#the-warnings-filter)).

skearnes:
Oh I see. Sorry, it wasn't clear that these were warnings flagging potential issues that might be legitimate.

connorcoley:
That's my bad for not putting an informative comment on this PR, sorry

PR #77, opened 2020-04-13
Expanding validations, especially of compound identifiers

connorcoley:
- Numerous validations added for Workup; not tested, but validated on examples at least
- Replaced MolFile identifier with MolBlock to match RDKit function
- Using RDKit to resolve SMILES, InChI, and MolBlocks
- Resolved Compounds will try to add RDKIT_BINARY identifier if
  rdkit is installed
- In the absence of a structural identifier, use the PubChem API to
  resolve a NAME CompoundIdentifier
- Tests added for these new features

### DIFF
--- a/ord_schema/validations.py
+++ b/ord_schema/validations.py
@@ -3,9 +3,16 @@
 import math
 import re
 import warnings
+from urllib import request
+from urllib.error import HTTPError
 from dateutil import parser
 
 from ord_schema.proto import reaction_pb2
+
+try:
+    from rdkit import Chem, __version__ as RDKIT_VERSION
+except ImportError:
+    Chem = None
 
 
 def validate_message(message, recurse=True):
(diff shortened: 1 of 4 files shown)


ord_schema/validations.py
```diff
@@ -503,3 +606,11 @@ def validate_binary_data(message):
     reaction_pb2.Percentage: validate_percentage,
     reaction_pb2.BinaryData: validate_binary_data,
 }
+
+COMPOUND_STRUCTURAL_IDENTIFIERS_ = [
```
skearnes, line 610: A couple style comments:
* Please put constants at the top of the module, where possible (the validators dict is an exception since it depends on functions being defined above it).
* Indicate _PRIVATE constants with a leading (not trailing) underscore; see [here](http://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations).

ord_schema/validations.py
```diff
@@ -3,10 +3,17 @@
 import math
 import re
 import warnings
+from urllib import request
+from urllib.error import HTTPError
```
skearnes, line 7: Should be `from urllib import error` (and then `error.HTTPError`). If you don't like `error`, you could probably just do `urllib.error.HTTPError`.

ord_schema/validations.py
```diff
@@ -3,10 +3,17 @@
...
 from dateutil import parser
 
 from ord_schema.proto import reaction_pb2
 
+try:
+    from rdkit import Chem, __version__ as RDKIT_VERSION
```
skearnes, line 13: Please put these imports on separate lines, even though they're from the same package; see [here](http://google.github.io/styleguide/pyguide.html#313-imports-formatting).

ord_schema/validations.py
```diff
@@ -3,10 +3,17 @@
...
+except ImportError:
+    Chem = None
```
skearnes, line 15: Define `RDKIT_VERSION` here as well?

ord_schema/validations.py
```diff
@@ -120,16 +127,78 @@ def validate_reaction_identifier(message):
...
+                    new_identifier.type = new_identifier.SMILES
+                    new_identifier.value = smiles
+                    new_identifier.details = 'NAME resolved by PubChem'
+                    break
+                except HTTPError:
+                    pass
+
+    # Try to create an RDKit binary identifier
+    # TODO(ccoley) add more sources for RDKIT_BINARY
+    if Chem and not any(identifier.type == identifier.RDKIT_BINARY for
```
skearnes, line 164: Consider moving this branch out into a separate function to make the overall function more readable (and make for easier testing).

ord_schema/validations.py
```diff
@@ -120,16 +127,78 @@ def validate_reaction_identifier(message):
...
+                    raise ValueError(f'RDKit {RDKIT_VERSION} could not validate'
+                                     f' InChI identifier {identifier.value}')
+            elif identifier.type == identifier.MOLBLOCK:
+                this_mol = Chem.MolFromMolBlock(identifier.value)
+                if this_mol is None:
+                    raise ValueError(f'RDKit {RDKIT_VERSION} could not validate'
+                                     ' MolBlock identifier')
+
+            # If RDKit mol was not defined, define it with this resolved mol
+            mol = mol or this_mol
```
skearnes, line 187: I'm a bit confused by this; why not just `break` after getting any valid molecule and avoid the `this_mol` temporary variable?

ord_schema/validations.py
```diff
@@ -239,6 +315,32 @@ def validate_reaction_observation(message):
 
 def validate_reaction_workup(message):
     ensure_details_specified_if_type_custom(message)
+    if (message.type == reaction_pb2.ReactionWorkup.WAIT and
```
skearnes, line 318: Note that if you want to save space, `message.WAIT` should work as well.

ord_schema/validations.py
```diff
@@ -256,6 +358,10 @@ def validate_reaction_outcome(message):
                 if key not in analysis_keys:
                     raise ValueError(f'Undefined analysis key {key} '
                                      'in ReactionProduct')
+    if not message.products:
+        warnings.warn('No products specified for reaction outcome; this should'
+                      ' only be used when screening for reaction conversion',
```
skearnes, line 363: "this should only be used" is a bit confusing, since you really mean "this should only be empty", right?

proto/reaction.proto
```diff
@@ -712,9 +712,7 @@ message ReactionProvenance {
   }
   RecordEvent record_created = 7;
   repeated RecordEvent record_modified = 8;
-  // TODO(ccoley): is it useful to create a unique ID field 
-  // that the centralized database can write to? (I know uniqueness is not
-  // enforceable)
+  // This is a unique ID field that the centralized database will write to
```
skearnes, line 715: Super nit: please add a period :)

http://google.github.io/styleguide/pyguide.html#386-punctuation-spelling-and-grammar

ord_schema/validations.py
connorcoley: If a message has defined a SMILES and an InChI, we will want to validate both despite only needing one to define the ```RDKIT_BINARY``` identifier. Consider a case where the SMILES is valid but the InChI is valid; we don't want to break after resolving the valid SIMLES. We could even consider adding greater consistency checks by requiring all resolved molecules to be identical, but that is risky given the challenges of molecule canonicalization (charges, tautomers, etc.)

ord_schema/validations.py
connorcoley: I thought about that, but wanted to use the full ```reaction_pb2```-derived variable for clarity since ```message``` isn't terribly informative.

ord_schema/validations.py
connorcoley: Ah yes I see the confusion

ord_schema/validations.py
connorcoley: Done

ord_schema/validations.py
connorcoley: Done

ord_schema/validations.py
connorcoley: Done. I was worried about ```error``` but keeping ```urllib.error.HTTPError``` is good

ord_schema/validations.py
connorcoley: Done

proto/reaction.proto
connorcoley: Reluctantly done

ord_schema/validations.py
connorcoley: Could you clarify which part you would like to see separated out? This block has a dual role: validating defined identifiers and creating the ```RDKIT_BINARY``` identifier. The former role should be inside this function at least.

skearnes: Yes, I was referring to the creation of the RDKIT_BINARY identifier.

ord_schema/validations.py
skearnes: SGTM; possibly add a comment to clarify this.

ord_schema/validations.py
skearnes: SGTM

proto/reaction.proto
skearnes: ;)

ord_schema/validations.py
connorcoley: Since this ends up only being a few lines of code (beyond what is required to validate the identifiers), I've kept it as-is for now. When we implement more extensive canonicalization to resolve tautomers, salts, charges, etc., I will make separate functions for  ```canonicalize_mol(mol)``` and ```add_rdkit_binary(message, mol)```

PR #78, opened 2020-04-13
Remove the long list of default pylint ignores

skearnes:
These were provided as part of the default `pylintrc` generated by pylint. I don't think most of them are necessary...


connorcoley:
Colab badges seem to be interfering with this PR, but cleaning up ```pylintrc``` makes sense to me

skearnes:
FYI I also re-added the `pull_request` event to the test trigger, since I seem to have broken the triggers for PRs with #76.

PR #79, opened 2020-04-13
Add URL field to BinaryData

skearnes:
For large analytical data, I expect we'll want the flexibility to host image and other binary data elsewhere and point to it in the record. This will make it much faster to manipulate and search the database and help to avoid size capacity limitations.

__Proposal__: set a maximum size limit on `BinaryData.value` and automatically upload and redirect during processing of submissions.


skearnes:
(ignore the notebook updates; this will be fixed by #76)

connorcoley:
If we will be encouraging URLs (which I fully support), does the name ```BinaryData``` still make sense, rather than a more generic ```Data``` message with a ```oneof { string value; bytes bytes_value; string url }```? That would also let us more easily differentiate unstructured data as strings (```value```) and data of a specific file format (```bytes_value``` and ```format```).

skearnes:
> If we will be encouraging URLs (which I fully support), does the name `BinaryData` still make sense, rather than a more generic `Data` message with a `oneof { string value; bytes bytes_value; string url }`? That would also let us more easily differentiate unstructured data as strings (`value`) and data of a specific file format (`bytes_value` and `format`).

PTAL

skearnes:
I thought about making `description` required but that might be too restrictive if we have names in the maps?

connorcoley:
This change looks fantastic. Would it be too much to ping the ```url``` to make sure it's valid? I thought about something similar for the url in reaction provenance messages.

> I thought about making `description` required but that might be too restrictive if we have names in the maps?

I do think that would be too restrictive.

skearnes:
> This change looks fantastic. Would it be too much to ping the `url` to make sure it's valid? I thought about something similar for the url in reaction provenance messages.

I think this could be a good idea, but I want to be sure it's not too much of a performance hit. I'll add this as a TODO.

PR #81, opened 2020-04-17
Expanding Nielsen example to show preparation of Pandas DF

connorcoley:
- Adding pandas as dependency in GitHub action
- Added two message_helpers for iterating over compound identifiers


ord_schema/message_helpers_test.py
```diff
@@ -111,5 +111,14 @@ def test_vendor(self):
             'Sally')
 
 
+class GetCompoundSmilesTest(absltest.TestCase):
+
```
skearnes, line 115: add a test for `get_compound_mol`?

connorcoley: Done, thanks for the reminder

PR #82, opened 2020-04-17
Updates for writing Data values to disk

skearnes:
This is prep work for getting files that can be uploaded to another location and replaced with links.


connorcoley:
Three minor comments:
- The ```find_messages``` logic is reminiscent of the ```validate_message``` logic, so I wonder if there is an opportunity to avoid duplicate code with a shared recursive generator. I don't think it's worth worrying about this though.
- For testing file writing, perhaps it's worth using [tempfile](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory). I've used it for temporary files in the past, but it can also create temporary directories. One of their [examples](https://docs.python.org/3/library/tempfile.html#tempfile-examples) shows how to use it in combination with ```with``` -- it's very tidy.
- How do you envision using the ```write_data``` function? Something like...
```
filename = write_data(message, dirname)
if filename:
    del message.value
    del message.bytes_value
    url = upload_file(filename)
    message.url = url
```
? I'm just wondering about whether the steps of (a) checking sizes, (b) creating a BytesIO, and (c) all belong in one function.

skearnes:
> Three minor comments:
> 
> * The `find_messages` logic is reminiscent of the `validate_message` logic, so I wonder if there is an opportunity to avoid duplicate code with a shared recursive generator. I don't think it's worth worrying about this though.

Agreed that there might be an opportunity here, but I'm not too worried about trying to merge them right now.

> * For testing file writing, perhaps it's worth using [tempfile](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory). I've used it for temporary files in the past, but it can also create temporary directories. One of their [examples](https://docs.python.org/3/library/tempfile.html#tempfile-examples) shows how to use it in combination with `with` -- it's very tidy.

Thanks; I'm using `tempfile` to create a subdirectory in the tests now. Since write_data requires a directory and not a file I think this makes sense. Agreed that in another setting where you just need a temporary file the context manager seems nice. The `self.test_subdirectory` pattern is the recommended approach at Google---and hence the one I'm used to---since it uses the `test_srcdir` flag (which I think is mostly important for executing tests on remote servers where you don't necessarily have general filesystem access).

> * How do you envision using the `write_data` function? Something like...
> 
> ```
> filename = write_data(message, dirname)
> if filename:
>     del message.value
>     del message.bytes_value
>     url = upload_file(filename)
>     message.url = url
> ```
> 
> ? I'm just wondering about whether the steps of (a) checking sizes, (b) creating a BytesIO, and (c) all belong in one function.

One nice thing about `oneof` is that if you assign to one of its members, the others are automatically cleared. The pattern I'm working on now looks like:

```
data_filename = message_helpers.write_data(
    data_message,
    dirname,
    min_size=FLAGS.min_size,
    max_size=FLAGS.max_size)
if data_filename:
    # client.upload() uploads the file and returns its URL.
    data_message.url = client.upload(data_filename)
```

I wanted to separate the URL generation from the writing since that will give us more flexibility in choosing how we do the upload/URL generation.

PR #85, opened 2020-04-19
Add validate_reactions script for use in the submission pipeline

skearnes:


skearnes:
One thing we may want to do is have `validate_message` optionally return a list of all errors instead of raising an exception; then if there are multiple issues we won't need multiple passes to find out...

skearnes:
Yeah, I think that's important; please wait to review until I can add that change to this PR.

skearnes:
Nevermind; this is going to be a bigger change that will require replacing exceptions with warnings or some other mechanism. Let's get this in first and I'll start working on that in another branch.

ord_schema/validate_reactions.py
```diff
@@ -0,0 +1,55 @@
...
+flags.DEFINE_string('output', None, 'Output file for validation errors.')
+
+
+def main(argv):
+    del argv  # Only used by app.run().
+    filenames = glob.glob(FLAGS.input_pattern)
+    num_errors = 0
+    errors = []
+    for filename in filenames:
+        reaction = message_helpers.load_message(
```
connorcoley, line 32: Should this get wrapped in a try/except so that if a ```filename``` can't be parsed at all, that is written to ```FLAGS.output``` as an error?

skearnes: Yes, good idea.

skearnes:
Added checks for bad message inputs; they're not perfect and we may need to keep iterating, but they should catch obvious mistakes and garbage input (e.g. if someone tries to submit something that was not generated by `text_format.MessageToString()`).

PR #97, opened 2020-04-22
Add update_reactions.py

skearnes:
This is a script---designed to be run after validate_reactions.py---for updating reaction protos to include things like record IDs, canonicalized structures, etc.

I'm not super happy with the relatively complex format for `--input_file`, but if we go ahead with this we can standardize this for the other submission scripts and add a shell script to generate it.

### DIFF
--- a/ord_schema/update_reactions.py
+++ b/ord_schema/update_reactions.py
@@ -14,7 +14,6 @@ import uuid
 
 from absl import app
 from absl import flags
-from google.protobuf import text_format
 
 from ord_schema import message_helpers
 from ord_schema.proto import reaction_pb2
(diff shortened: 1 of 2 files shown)


connorcoley:
Would it be simpler to have the python script directly call `git diff` in a subprocess and parse the output?

skearnes:
> Would it be simpler to have the python script directly call `git diff` in a subprocess and parse the output?

I think that depends on whether we want to make it so these scripts only work in the context of a git repo. For this script, the answer will almost certainty be "yes", so that's a good idea. It will also allow me to remove the try/except block for the `git mv` operation...let me take another pass.

skearnes:
PTAL; in the end I decided to punt on the subprocess calls to git; it was getting too complicated with setting `base_ref` and needing multiple branches for the tests. In another PR I'll create a script that generates this file and adjust validate_reactions.py to take the same input format so we can use it everywhere in the submission pipeline.

PR #100, opened 2020-04-23
Stock solution helper

connorcoley:
build_solution helper function and tests
- Uses a defined solute and solvent (e.g., from build_compound)
- Returns the modified solute and solvent in a list, to make it easy
  to define stock solution inputs using components.MergeFrom
- Normalization of units could likely be made more elegant by another
  helper function designed to convert any units to any other units
  of the same type. Because this is the only use case currently in
  the codebase, I don't mind having it defined within this function.
- The `UnitResolver` class can now take different `unit_synonym`
  initializations so `build_solution` can make a concentration-only
  resolver via `units.CONCENTRATION_UNIT_SYNONYMS`
- Also added missing `Length` message types  to `UnitResolver`

### DIFF
--- a/ord_schema/message_helpers.py
+++ b/ord_schema/message_helpers.py
@@ -98,7 +98,7 @@ def build_compound(smiles=None, name=None, amount=None, role=None,
     return compound
 
 
-def build_solution(solute, solvent, concentration):
+def set_solute_moles(solute, solvent, concentration):
     """Helps define components for stock solution inputs with a single solute
     and a single solvent compound.
 
--- a/ord_schema/message_helpers.py
+++ b/ord_schema/message_helpers.py
@@ -98,6 +98,58 @@ def build_compound(smiles=None, name=None, amount=None, role=None,
     return compound
 
 
+def build_solution(solute, solvent, concentration):
+    """Helps define components for stock solution inputs with a single solute
+    and a single solvent compound.
+
+    Args:
+        solute: Compound with identifiers, roles, etc.; this argument is
+            modified in place to define an amount in moles.
+        solvent: Compound with identifiers, roles, etc. and defined volume.
+        concentration: string defining solute concentration.
+
+    Returns:
+        List of Compounds to assign to a repeated components field.
+    """
+    # Get solvent volume in liters.
+    if not solvent.HasField('volume') or not solvent.volume.value:
+        raise ValueError('solvent must have defined volume')
+    volume = solvent.volume.value
+    if solvent.volume.units == solvent.volume.LITER:
+        volume_liter = volume
+    elif solvent.volume.units == solvent.volume.MILLILITER:
+        volume_liter = volume * 1e-3
+    elif solvent.volume.units == solvent.volume.MICROLITER:
+        volume_liter = volume * 1e-6
+    else:
+        raise ValueError('solvent units not recognized by build_solution')
+    # Get solute concentration in molar.
+    resolver = units.UnitResolver(
+        unit_synonyms=units.CONCENTRATION_UNIT_SYNONYMS,
+        forbidden_units={},)
+    concentration_pb = resolver.resolve(concentration)
+    if concentration_pb.units == concentration_pb.MOLAR:
+        concentration_molar = concentration_pb.value
+    elif concentration_pb.units == concentration_pb.MILLIMOLAR:
+        concentration_molar = concentration_pb.value * 1e-3
+    elif concentration_pb.units == concentration_pb.MICROMOLAR:
+        concentration_molar = concentration_pb.value * 1e-6
+    # Assign moles amount and return.
+    moles = volume_liter * concentration_molar
+    if moles < 1e-6:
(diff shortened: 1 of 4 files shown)


ord_schema/units.py
```diff
@@ -70,15 +72,34 @@
...
+        reaction_pb2.Concentration.MICROMOLAR: ['uM', 'micromolar'],
+    },
+}
+
 
 class UnitResolver:
     """Resolver class for translating value+unit strings into messages."""
 
-    def __init__(self):
+    def __init__(self, unit_synonyms=None, forbidden_units=None):
```
skearnes, line 93: Please add a docstring for __init__

skearnes, line 93: Maybe use `forbidden_units=_FORBIDDEN_UNITS` as the default so `None` can mean that there are no forbidden units?

ord_schema/message_helpers.py
```diff
@@ -98,6 +98,58 @@ def build_compound(smiles=None, name=None, amount=None, role=None,
...
+    if concentration_pb.units == concentration_pb.MOLAR:
+        concentration_molar = concentration_pb.value
+    elif concentration_pb.units == concentration_pb.MILLIMOLAR:
+        concentration_molar = concentration_pb.value * 1e-3
+    elif concentration_pb.units == concentration_pb.MICROMOLAR:
+        concentration_molar = concentration_pb.value * 1e-6
+    # Assign moles amount and return.
+    moles = volume_liter * concentration_molar
+    if moles < 1e-6:
+        solute.moles.CopyFrom(reaction_pb2.Moles(units='NANOMOLES',
```
skearnes, line 140: I'd suggest making a copy of `solute` here before editing in case that solute is used elsewhere.

skearnes: Or possibly checking that solute doesn't already have a `moles` value set?

ord_schema/units.py
connorcoley: I had that originally, but pylint yelled at me for using a dangerous default value since dicts are mutable. Would you prefer I just disable that pylint warning?

ord_schema/message_helpers.py
connorcoley: What do you want the desired behavior to be here? I thought that an in-place modification of solute would actually be preferable.

Likewise, the intended behavior was to overwrite any previous value of `moles`. I could see one use case as something like...

```
for conc_string, yield_value in my_big_data_list:
  reaction.CopyFrom(template_reaction)
  message_helpers.build_solution(
    reaction.inputs['solution A'].components[0],
    reaction.inputs['solution A'].components[1],
    conc_string
  )
  reaction.outcomes[0].products[0].yield.value = yield_value
  my_reactions.append(reaction)
```

For the case where we do not do any in-place modifications, that example would look like

```
for conc_string, yield_value in my_big_data_list:
  reaction.CopyFrom(template_reaction)
  reaction.inputs['solution A'].components.MergeFrom(
    message_helpers.build_solution(
      reaction.inputs['solution A'].components[0],
      reaction.inputs['solution A'].components[1],
      conc_string
    )
  )
  reaction.outcomes[0].products[0].yield.value = yield_value
  my_reactions.append(reaction)
```

Thoughts?

connorcoley:
> High-level question: we don't use the `Concentration` message anywhere in the schema; should it be removed?

It's convenient to have it present for the sake of reusing the unit resolver helper function, but that's an interesting question.

ord_schema/units.py
skearnes: No, I agree with that warning; requiring the user to pass in an empty dict is fine (please indicate that in the docstring).

ord_schema/message_helpers.py
skearnes: I see; I guess I was expecting that most people would be using `Compound` messages directly rather than having copied them into `reaction.inputs` first. I'm okay with keeping it as-is as long as we add some checks that the `moles` field hasn't already been set to avoid accidental overwrites. Does that sound good?

This ambiguity is another reason I might want to resolve #108 before merging this PR.

connorcoley: Adding a keyword argument `overwrite` that defaults to `False` would be a good compromise here I think. In that case, an error would be raised if any of the `amount` fields of a `solute` are defined. Let's continue the broader discussion of solutions in #108

skearnes: SGTM

connorcoley:
@skearnes PTAL again. I still think there is value in having this helper function and envision the following use case:
```
reaction.inputs['solution A'].components.MergeFrom(
    message_helpers.set_solute_moles(
      message_helpers.build_compound(name='solute'),
      [
        message_helpers.build_compound(name='solvent 1', amount='5 mL'),
        message_helpers.build_compound(name='solvent 2', amount='1 mL')
      ],
     '1 molar',
    )
  )
```

skearnes:
Fixes #89

PR #106, opened 2020-04-23
Use binary format in validate_reactions

skearnes:
Changes:
* Adds support for binary proto format.
* Changes `--input_file` to match update_reactions.

### DIFF
--- a/ord_schema/validate_reactions_test.py
+++ b/ord_schema/validate_reactions_test.py
@@ -26,19 +26,19 @@ class ValidateReactionsTest(absltest.TestCase):
         dummy_component.mass.value = 1
         dummy_component.mass.units = reaction_pb2.Mass.GRAM
         reaction1.outcomes.add().conversion.value = 75
-        with open(os.path.join(self.test_subdirectory, 'reaction1.pbtxt'),
+        with open(os.path.join(self.test_subdirectory, 'reaction1.pb'),
                   'wb') as f:
             f.write(reaction1.SerializeToString())
         reaction2 = reaction_pb2.Reaction()
-        with open(os.path.join(self.test_subdirectory, 'reaction2.pbtxt'),
+        with open(os.path.join(self.test_subdirectory, 'reaction2.pb'),
                   'wb') as f:
             f.write(reaction2.SerializeToString())
-        with open(os.path.join(self.test_subdirectory, 'reaction3.pbtxt'),
+        with open(os.path.join(self.test_subdirectory, 'reaction3.pb'),
                   'wb') as f:
             f.write(b'garbage that is not a reaction proto')
 
     def test_main(self):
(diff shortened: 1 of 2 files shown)


ord_schema/validate_reactions_test.py
```diff
@@ -28,15 +27,15 @@ def setUp(self):
         dummy_component.mass.units = reaction_pb2.Mass.GRAM
         reaction1.outcomes.add().conversion.value = 75
         with open(os.path.join(self.test_subdirectory, 'reaction1.pbtxt'),
-                  'w') as f:
-            f.write(text_format.MessageToString(reaction1))
+                  'wb') as f:
+            f.write(reaction1.SerializeToString())
         reaction2 = reaction_pb2.Reaction()
         with open(os.path.join(self.test_subdirectory, 'reaction2.pbtxt'),
```
connorcoley, line 33: Is the file extension convention for binary files `*.pb`? It doesn't change the functionality, but these tests should probably be changed from `*.pbtxt` (a few cases of this)

skearnes: Oh good catch; fixed.

(Note that I'm not actually sure what the file extension should be. I picked `.pb` to match `.pbtxt`, which is the convention for `text_format` output. At Google we usually either keep the byte strings in memory and do RPCs or else write them to disk as part of another container---often [LevelDB](https://github.com/google/leveldb).)

I actually played around yesterday with using LevelDB to get around issues with disk block size ([example](http://www.alphaurax-computer.com/computer-tips/hard-drive-knowledge-blocks-vs-sectors)). We may want to leverage something like this in the future if we have trouble with many small files that each take up an entire block and lead to excessive disk usage...

skearnes: Note that this is what the TFRecords format is designed for: storage of multiple serialized protos. See [here](https://developers.google.com/protocol-buffers/docs/techniques#streaming) and [here](https://www.tensorflow.org/tutorials/load_data/tfrecord).

skearnes: But unlike LevelDB, TFRecords is not indexed or sorted by key, so it's not as useful for a database as LevelDB.

connorcoley: Interesting. Should we think about non-arbitrary ways of grouping serialized protos? For example, a batched submission from a single project or publication would all be stored in the same file?

skearnes: It sort of depends how it will be used. We could certainly provide such files for download, and one for the full dataset for fast lookup by record_id.

For storing in the repo, one TFRecords file per submission could actually be pretty handy. TFRecords has the advantage there because it is a single file with a nicely supported python interface. LevelDB is represented as a directory with multiple files and has only third-party python APIs (like [plyvel](https://github.com/wbolster/plyvel)).

Let me copy this discussion to a new issue.

PR #111, opened 2020-04-25
Consolidate submission scripts

skearnes:
* Combines validate_reactions.py and update_reactions.py into a single script (process_dataset.py).
* Uses `Dataset` as the input format. Multiple `Dataset` protos can be passed in with a glob pattern.
* Adds `write_message` to message_helpers and updates `load_message` to raise a consistent Exception type for easier use in scripts (like process_dataset.py).


ord_schema/message_helpers.py
```diff
@@ -316,17 +322,43 @@ def load_message(filename, message_type, input_format):
         ValueError: if the message cannot be parsed, or if `input_format` is not
             supported.
     """
-    if input_format == 'binary':
+    input_format = MessageFormats(input_format)
```
connorcoley, line 325: This is an interesting way to check allowable values -- is this a common use of `enum.Enum`? I've never used `enum`.

ord_schema/message_helpers.py
```diff
@@ -316,17 +322,43 @@ def load_message(filename, message_type, input_format):
...
-        if input_format == 'json':
-            return json_format.Parse(f.read(), message_type())
-        if input_format == 'pbtxt':
-            return text_format.Parse(f.read(), message_type())
-        if input_format == 'binary':
-            message = message_type()
-            message.ParseFromString(f.read())
+        try:
+            if input_format == MessageFormats.JSON:
+                message = json_format.Parse(f.read(), message_type())
```
connorcoley, line 333: Stylistically, is this preferred to `return` on every line? It is slightly more readable.

ord_schema/process_dataset.py
```diff
@@ -0,0 +1,143 @@
...
+        if errors:
+            validation_errors = True
+            if FLAGS.write_errors:
+                with open(f'{filename}.error', 'w') as f:
+                    for error in errors:
+                        f.write(f'{error}\n')
+    # NOTE(kearnes): Run validation for all datasets before
+    # exiting if there are errors.
+    if validation_errors:
+        raise ValueError('validation encountered errors')
```
connorcoley, line 70: Is it worth expanding this to something like "validation encountered errors; re-run with '--write_errors true' to save errors to file"? While this script should only be seen by superusers/developers, might make it easier to remember this option.

ord_schema/process_dataset.py
```diff
@@ -0,0 +1,143 @@
...
+        List of validation error messages.
+    """
+    basename = os.path.basename(filename)
+    errors = []
+    for reaction in dataset.reactions:
+        for error in validations.validate_message(
+                reaction, raise_on_error=False):
+            errors.append(error)
+            logging.warning('Validation error for %s: %s', basename, error)
+    logging.info('Validation summary for %s: %d/%d successful (%d failures)',
```
connorcoley, line 93: If we are validating two reactions and they generate 0 and 2 errors, wouldn't this report that 0/2 were successful (2 failures) instead of 1/2 successful?

ord_schema/process_dataset.py
```diff
@@ -0,0 +1,143 @@
...
+def _update_reaction(reaction):
+    """Updates a Reaction message.
+
+    Current updates:
+      * Sets record_id if not already set.
+
+    Args:
+        reaction: reaction_pb2.Reaction message.
+    """
+    if not reaction.provenance.record_id:
```
connorcoley, line 110: Do we want to only set the `record_id` if not already set? What if a submitter puts in a dummy placeholder value or a malicious value? 

I suppose we would ideally check if the `record_id` was previously set _by this validation script_. Perhaps you can drop a TODO in here, since I think the logic we will eventually want will be more complex than `if not reaction.provenance.record_id`.

ord_schema/message_helpers.py
skearnes: I was very surprised that I couldn't find anything in the `enum` documentation about testing for bad values; I found that this worked well and had the nice side effect of returning the Enum member (since `MessageFormats.BINARY != 'binary'`, but `MessageFormats.BINARY == MessageFormats('binary')`).

I like `enum` mostly for constants that are used in multiple places. Also this avoid needing to add a separate "unsupported type" error since `MessageFormats` will raise an exception if you pass in an invalid type.

ord_schema/message_helpers.py
skearnes: I know...but pylint was giving me an error about the `return` statements not all returning an expression so I just went with this since it wasn't obvious how to fix that error.

I'll just disable the error b/c I agree returns are more readable.

ord_schema/process_dataset.py
skearnes: By default all the errors will be written to the console, so they're not hidden. Do you still want that suggestion in that case?

ord_schema/process_dataset.py
skearnes: Good catch; will fix.

ord_schema/process_dataset.py
skearnes: Agreed there will be more complexity here; will add a TODO.

ord_schema/process_dataset.py
connorcoley: In that case, no. What you have makes sense.

skearnes: Added.

connorcoley:
LGTM

ord_schema/process_dataset.py
skearnes: One interesting note about boolean flags: if a flag defaults to False, you only need to pass the flag to turn it on: `--write_errors`. If a flag defaults to True, the flag module creates a special flag to turn it off: `--nowrite_errors` (just the flag name prefixed with `no`).

Issue #116, opened 2020-04-27
Include proto python wrappers in the repo?


skearnes:
The first step of `setup.py` currently requires the user to install and run `protoc`. How would you feel about including the generated `*_pb2.py` files as part of the repo so the entire thing would be pure python?

Pros:
* Easier to install for all users (no dependency on `protoc`)
* Simplifies automated actions (since the `protoc` step won't be needed)

Cons:
* Requires automated updates of the generated python files when the `*.proto` files change (I started working on this in #115)
* Risk of a mismatch if this automated workflow fails

connorcoley:
I think this would be a positive change as long as we can automatically update the files (#115).

Is there a way to have `setup.py` run `protoc` only if it detects that the compiled `*_pb2.py` does not match the source `*.proto`? I don't know if this is overly complex, but could we compare, for example, the hash of a `.proto` schema definition to an expected value that is recorded inside `*_pb2.py` after compilation.

skearnes:
Looking deeper into this, I found two things:

* It's hard to get any automated checks to use the updated python wrappers.
* `pip install protobuf` (one of our current deps) installs `protoc` automatically.

I think this means we should just keep the workflow as is, and instead I can try to remove the need for `python setup.py build_py` so that the normal `python setup.py install` works as a one-shot command.

skearnes:
I think this (no separate `protoc` install) only works with conda, but since we're using that for `rdkit` I think that's ok.

connorcoley:
SGTM. Good to consolidate dependencies to simple conda installations as long as we're using it.

PR #117, opened 2020-04-27
Update installation

skearnes:
Notable changes:
* `conda install protobuf` also installs `protoc`, so no additional step is needed.
* `python setup.py install` automatically calls `build_py`, so no additional step is needed.


skearnes:
Closes #116

PR #120, opened 2020-04-27
Add ReactionStep message

skearnes:
Adds a `ReactionStep` message to support multi-step and/or one-pot reactions.

* Only the first step would require any inputs (other steps could be for conditions)
* Only the last step would require any outputs


skearnes:
Closing since we are moving forward with a different option.

PR #122, opened 2020-04-28
Update colab badges action

skearnes:
After a lengthy battle with GitHub actions and some dangerous spelunking into the depths of forums, I have learned the following things:

* It is not possible to use the `pull_request` event to trigger a workflow that needs to commit changes to a branch in a forked repository.
* It is not possible to automatically trigger one workflow by an action caused by a workflow run in a forked repository.

The upshot for submissions:

* We must use `push` events to trigger validation/updates that will add commits to submission pull requests. This PR updates the colab badge action to use a `push` event trigger.
* Since it will not be possible to automatically trigger re-runs of the validation actions on the automatically generated commits, we can either disable required checks (and make sure reviewers check that that last non-automated commit passed the checks) or manually trigger the checks during the review process (with dummy commits, open/close cycles on the PR, or possibly third-party [apps](https://www.actionspanel.app/)).

In this repo we will continue to require checks; we can iterate in the submission repo.


skearnes:
Note that the colab notebook changes are mostly due to a change in JSON indentation that I added upstream.

PR #124, opened 2020-04-28
Updates for process_datasets.py

skearnes:
* Add `--input_file` argument
* Automatically generate dataset filenames using `uuid4()`
* Check suffixes when writing messages to disk

### DIFF
--- a/ord_schema/process_dataset.py
+++ b/ord_schema/process_dataset.py
@@ -145,6 +145,7 @@ def _get_filenames():
     raise ValueError('one of --input_pattern or --input_file is required')
 
 
+# pylint: disable=too-many-branches
 def main(argv):
     del argv  # Only used by app.run().
     filenames = sorted(_get_filenames())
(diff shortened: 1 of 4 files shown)


ord_schema/process_dataset.py
```diff
@@ -122,32 +127,72 @@ def _update_reaction(reaction):
...
             combined.reactions.extend(dataset.reactions)
     if FLAGS.output:
-        message_helpers.write_message(
-            combined, FLAGS.output, FLAGS.input_format)
+        output_filename = FLAGS.output
+    else:
+        suffix = message_helpers.get_suffix(FLAGS.input_format)
+        dataset_id = uuid.uuid4().hex
+        output_filename = os.path.join(
+            'data', dataset_id[:2], f'{dataset_id}{suffix}')
```
connorcoley, line 181: I think this is a good solution for organizing subfolders

skearnes: Yes; it was your idea :p

PR #137, opened 2020-05-06
Create data_storage module

skearnes:
Since the GCS process is moving slowly, and in an effort to have the data live close to the records that reference it, I'm moving forward with using Git LFS to store large `Data` values.

* Factors out `write_data` from the message_helpers module with minor changes.
* Adds `extract_data` function and integrates it into the process_datasets pipeline.

Feedback is welcome on my choice of value for the `url` field when using Git LFS.


ord_schema/data_storage.py
```diff
@@ -0,0 +1,114 @@
...
+        message, reaction_pb2.Data)
+    for data_message in data_messages:
+        data_filename = write_data(
+            data_message,
+            dirname,
+            min_size=min_size,
+            max_size=max_size)
+        if data_filename:
+            basename = os.path.basename(data_filename)
+            output_filename = os.path.join(
```
connorcoley, line 106: I don't see this as essential, but is it worth putting this 2-character subdirectory convention in a helper function? I imagine we'll encounter several cases where we need to go from an ID to its filepath.

skearnes: Yes, good idea; I'll put something together for a future PR.

PR #143, opened 2020-05-08
Distinguish between validations and updates

skearnes:
In the submission pipeline it will be helpful to differentiate between "validations" (checking messages for errors) and "updates" (adding or modifying message fields for things like canonicalization or resolving names to SMILES). The current validations include some update functionality; a message may be changed by the validation process.

This PR, in essence, makes all the validations `const` functions that do not change the underlying message. ~~After this PR is merged~~ I will re-add the pubchem resolver and binary RDKit addition as "updates" and refactor process_datasets.py to use them in the `--update` section of the submission workflow.

__Edit:__ Finished the "updates" module and added to this PR.

### DIFF
--- a/ord_schema/updates.py
+++ b/ord_schema/updates.py
@@ -1,0 +1,128 @@
+"""Automated updates for Reaction messages."""
+
+import datetime
+import urllib
+import uuid
+
+from rdkit import Chem
+
+from ord_schema import message_helpers
+from ord_schema.proto import reaction_pb2
+
+_COMPOUND_STRUCTURAL_IDENTIFIERS = [
+    reaction_pb2.CompoundIdentifier.SMILES,
+    reaction_pb2.CompoundIdentifier.INCHI,
+    reaction_pb2.CompoundIdentifier.MOLBLOCK,
+    reaction_pb2.CompoundIdentifier.CXSMILES,
+    reaction_pb2.CompoundIdentifier.XYZ,
+]
+
+
+def _pubchem_resolve(value_type, value):
+    """Resolves compound identifiers to SMILES via the PubChem REST API."""
+    response = urllib.request.urlopen(
+        'https://pubchem.ncbi.nlm.nih.gov/rest/pug/compound/'
+        f'{value_type}/{value}/property/IsomericSMILES/txt')
+    return response.read().decode().strip()
+
+
+def resolve_names(message):
+    """Attempts to resolve compound NAME identifiers to SMILES.
+
+    When a NAME identifier is resolved, a SMILES identifier is added to the list
+    of identifiers for that compound. Note that this function moves on to the
+    next Compound after the first successful name resolution.
+
+    Args:
+        message: Reaction proto.
+
+    Returns:
+        Boolean whether `message` was modified.
+    """
+    modified = False
+    compounds = message_helpers.find_submessages(message, reaction_pb2.Compound)
+    for compound in compounds:
+        if any(identifier.type in _COMPOUND_STRUCTURAL_IDENTIFIERS
+               for identifier in compound.identifiers):
+            continue  # Compound already has a structural identifier.
+        for identifier in compound.identifiers:
+            if identifier.type == identifier.NAME:
+                try:
+                    smiles = _pubchem_resolve('name', identifier.value)
+                    new_identifier = compound.identifiers.add()
+                    new_identifier.type = new_identifier.SMILES
+                    new_identifier.value = smiles
+                    new_identifier.details = 'NAME resolved by PubChem'
+                    modified = True
+                    break
+                except urllib.error.HTTPError:
+                    pass
+    return modified
+
+
+def add_binary_identifiers(message):
+    """Adds RDKIT_BINARY identifiers for compounds with valid structures.
+
+    Note that the RDKIT_BINARY representations are mostly useful in the context
+    of searching the database. Accordingly, this function is not included in the
+    standard set of Reaction updates in update_reaction().
+
+    Args:
+        message: Reaction proto.
+
+    Returns:
+        Boolean whether `message` was modified.
+    """
+    modified = False
+    compounds = message_helpers.find_submessages(message, reaction_pb2.Compound)
+    for compound in compounds:
+        if any(identifier.type == identifier.RDKIT_BINARY
+               for identifier in message.identifiers):
+            continue
+        for identifier in compound.identifiers:
+            mol = None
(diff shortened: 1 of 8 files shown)


skearnes:
Added the new "updates" module so you can see moves instead of deletes; sorry this PR got a bit bigger than usual.

ord_schema/validations.py
```diff
@@ -215,95 +201,48 @@ def validate_reaction_input(message):
...
-            # sanitization, etc., and separate this into its own function.
-            new_identifier = message.identifiers.add()
-            new_identifier.type = new_identifier.RDKIT_BINARY
-            new_identifier.bytes_value = mol.ToBinary()
-    # pylint: enable=too-many-branches
-
-    return message
+    if not Chem or any(identifier.type == identifier.RDKIT_BINARY
+                       for identifier in message.identifiers):
+        return
```
connorcoley, line 212: - If `Chem` is `None`, we should probably issue a `ValidationWarning` at the very least to indicate that it wasn't possible to validate the compound identifiers.
- If `identifier.type == identifier.RDKIT_BINARY`, I don't think we should assume we're in the clear. Maybe try to initialize a `Chem.Mol` and  `if mol is None: warnings.warn(..., ValidationError)`

ord_schema/validations.py
```diff
@@ -136,7 +126,6 @@ def ensure_details_specified_if_type_custom(message):
     if message.type == message.CUSTOM and not message.details:
         warnings.warn(f'Custom type defined for {type(message)}, '
                       'but details field is empty', ValidationError)
-    return message
```
connorcoley, line 139: I like the idea of separating validations and updates. Removing the `return message` from every validator is a nice bonus

ord_schema/validations.py
```diff
@@ -318,11 +257,10 @@ def validate_vessel(message):
             not message.preparation_details:
         warnings.warn('VesselPreparation custom, but no details provided',
                       ValidationError)
-    return message
 
 
 def validate_reaction_setup(message):
-    return message
+    del message  # Unused.
```
connorcoley, line 263: I don't understand what the intended behavior is here. Why `del message` and not, e.g., `pass`? Same comment for the other functions below.

ord_schema/validations.py
connorcoley: Also, what if the only identifier is a `NAME`? Should we issue a `ValidationWarning` here? It might be worth mentioning at this validation stage that we haven't yet tried to resolve the name.

ord_schema/updates.py
```diff
@@ -0,0 +1,128 @@
...
+            mol = None
+            if Chem and identifier.type == identifier.SMILES:
+                mol = Chem.MolFromSmiles(identifier.value)
+            elif identifier.type == identifier.INCHI:
+                mol = Chem.MolFromInchi(identifier.value)
+            elif identifier.type == identifier.MOLBLOCK:
+                mol = Chem.MolFromMolBlock(identifier.value)
+            if mol is not None:
+                compound.identifiers.add(
+                    bytes_value=mol.ToBinary(), type='RDKIT_BINARY')
```
connorcoley, line 92: Should we add `details` saying what the source of the `RDKIT_BINARY` was? This might not be necessary but thought I'd ask anyway. It looks like here you might end up with 3 `RDKIT_BINARY`s if there are identifiers for `SMILES`, `INCHI`, _and_ `MOLBLOCK` for whatever reason.

ord_schema/updates.py
```diff
@@ -0,0 +1,128 @@
...
+    for func in _UPDATES:
+        modified = modified or func(reaction)
+    if modified:
+        reaction.provenance.record_modified.add().time.value = (
+            datetime.datetime.utcnow().ctime())
+
+
+# Standard updates.
+_UPDATES = [
+    resolve_names,
```
connorcoley, line 127: `add_binary_identifiers` too?

ord_schema/validations.py
skearnes: * I think we should require rdkit for these; the conditional imports don't really make sense since we'll want to run all the validations.
* Agreed that RDKIT_BINARY isn't enough.
* Yes, I think every reaction should have at least one structural identifier, but I wanted to check with you since I remembered that some Compounds might only have NAMEs b/c they have no defined structure...a warning at the Compound level is a good idea.

ord_schema/updates.py
skearnes: Yes, that's a good idea.

ord_schema/updates.py
skearnes: I was thinking that maybe we'd leave this out and only put it in for uploading to BigQuery, etc. It adds a fair amount of noise to the submission to have commits that add a bunch of unreadable bytes and I don't think there's much benefit to having the RDKIT_BINARY reps in typical offline ML settings. (There's a comment about this in the docstring for add_binary_identifiers.)

What do you think?

ord_schema/validations.py
skearnes: pylint complains if `message` is unused; I had `pass` originally but changed to this so we could avoid changing the function signature. I could replace `message` with `_` in the signature if you prefer.

ord_schema/updates.py
skearnes: Also I'll update this so it breaks after the first RDKIT_BINARY.

skearnes:
Thanks for your comments; PTAL.

ord_schema/validations.py
connorcoley: Ah, it makes sense that pylint would complain. `del` seems like an odd placeholder still, but that's okay

ord_schema/updates.py
connorcoley: I actually like the idea of `RDKIT_BINARY` being present, since it makes it very convenient to use the data for ML in Python if you reload the compounds from binary (and are using RDKit, of course). 

This might not matter as much if we are primarily working with SMILES and MolBlocks. I also understand the drawback of having unreadable bytes. But having a pre-parsed RDKit object is quite convenient

ord_schema/validations.py
connorcoley: The warning for compounds w/o structural identifier is perfect, thanks for adding that

ord_schema/updates.py
skearnes: I'll suggest adding this post-review in a separate PR if we want it as part of the official record. That way we don't need to include it as part of the submission review.

Issue #152, opened 2020-05-13
RSC chemical methods ontology for AnalysisType


dcambie:
Congratulations on this project :)
I have a small consideration.

Currently, in the `ReactionAnalysis` message, the `AnalysisType` enum lists 17 values.
https://github.com/Open-Reaction-Database/ord-schema/blob/b8dfae2e3ea3d4a011cfcfdbcab79b501d6398af/proto/reaction.proto#L678-L695

However, many different analytical techniques can be used to characterize a reaction. To this end, what about using the RSC ontology for chemical methods[1]?
e.g. a new CHMO ID field could be adopted instead of adding new values to the AnalysisType enum.

[1] https://github.com/rsc-ontologies/rsc-cmo
See for example [LC-DAD](https://www.ebi.ac.uk/ols/ontologies/chmo/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FCHMO_0001738&viewMode=All&siblings=true)

skearnes:
Thanks for the suggestion! I'll let @connorcoley weigh in on this since he's got the best perspective on the diversity of methods here. My main concerns are that it adds another level of complexity for users to have to find their approach in the system and downstream applications may not be able to profitably distinguish between methods at that resolution.

A possible alternative would be to modify the schema to make this an optional specification; for example adding a `int32 chmo_id` field (possibly as a `oneof` with AnalysisType unless we allow the user to specify both).

connorcoley:
I agree that our `AnalysisType` enum covers a very limited list compared to the CHMO ontology. I like the suggestion of having an optional `int32 chmo_id` field to allow this greater degree of flexibility. I also agree with Steve that users may not want to deal with that increased level of complexity, so allowing both to be defined would be find with me. I'll submit a quick PR

PR #155, opened 2020-05-14
Adding CHMO ID field for analysis (addresses #152)

connorcoley:
Creates the `chmo_id` field that can be defined in addition to the `AnalysisType` to specify the analytical method used with greater specificity, see https://github.com/rsc-ontologies/rsc-cmo

addresses #152

### DIFF
--- a/proto/reaction.proto
+++ b/proto/reaction.proto
@@ -695,17 +695,20 @@ message ReactionAnalysis {
     SFC = 17;  // Supercritical fluid chromatography.
   }
   AnalysisType type = 1;
-  // Any details about analysis (e.g., NMR type, columns, gradients, conditions)
-  string details = 2;
+  // RSC Chemical Methods Ontology ID to define the analytical method with
+  // greater specificity. Defined at https://github.com/rsc-ontologies/rsc-cmo.
+  int32 chmo_id = 2;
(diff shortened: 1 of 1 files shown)


proto/reaction.proto
```diff
@@ -695,17 +695,20 @@ message ReactionAnalysis {
     SFC = 17;  // Supercritical fluid chromatography.
   }
   AnalysisType type = 1;
-  // Any details about analysis (e.g., NMR type, columns, gradients, conditions)
-  string details = 2;
+  // RSC Chemical Methods Ontology ID to define the analytical method with
+  // greater specificity. Defined at https://github.com/rsc-ontologies/rsc-cmo.
+  int32 chmo_id = 2;
```
skearnes, line 700: Should we use a `string` here and require the CHMO prefix? For example, the full ID for 1H NMR is `CHMO:0000593`.

connorcoley: I'm ambivalent, but my thought was that the field name is already specific to CHMO so it isn't necessary. Do you want me to change it?

skearnes: No, int is fine; I think if we were encoding PubChem CIDs we wouldn't prefix them with `CID` either, right?

connorcoley: Correct

PR #158, opened 2020-05-15
Replacing bools in proto with enum

connorcoley:
- All booleans in proto are replaced with Boolean.BooleanValue enums
- Default value is UNSPECIFIED = 0; then TRUE = 1; FALSE = 2;
- Usage is somewhat complicated by the fact that you can still use
  the assignment operator with "True" and "False" in Python, since
  these will automatically be cast to integers as 1 or 0. With these
  values for enum indices, setting the boolean to True will resolve to 1,
  which is TRUE. However, setting the boolean to False will resolve to
  0, which is UNSPECIFIED not FALSE.
- Existing code/tests have been updated to explicitly use
  `Boolean.TRUE` instead of True even though these are
  functionally equivalent.
- Filters in `visualization/filters.py` required modification as well,
  because `if bool_value` will resolve to False if UNSPECIFIED==0,
  True if TRUE==1, and True if FALSE==2.
- An additional change is the addition of 
  `Boolean.BooleanValue volume_includes_solutes` in `Compound` messages


connorcoley:
P.S. `volume_includes_solutes` isn't the most elegant name, but I wanted to keep it as descriptive as possible.

ord_schema/visualization/filters.py
```diff
@@ -3,9 +3,14 @@
 
 from ord_schema import units
 from ord_schema import message_helpers
+from ord_schema.proto import reaction_pb2
 from ord_schema.visualization import drawing
 
 
+def _is_true(boolean):
+    return boolean == reaction_pb2.Boolean.BooleanValue.TRUE
```
skearnes, line 11: Just `reaction_pb2.Boolean.TRUE`?

proto/reaction.proto
```diff
@@ -141,6 +141,11 @@ message Compound {
     Moles moles = 3;
     Volume volume = 4;
   }
+  // Whether the volume measurement refers to the pure substance or to the
+  // total volume of the reaction input. An example of when this field should
+  // be TRUE is when stock solutions are prepared by adding solvent to a
+  // volumetric flask already containing solute(s).
+  Boolean.BooleanValue volume_includes_solutes = 14;
```
skearnes, line 148: I'm ok with this, but some possible alternatives:

* `is_final_volume`
* `is_total_volume`

ord_schema/visualization/filters.py
connorcoley: Oops, fixed

proto/reaction.proto
connorcoley: For clarity over brevity, let's stick to `volume_includes_solutes`. This also reinforces this field's relevance to stock solution preparation

PR #159, opened 2020-05-15
Start validations.md

skearnes:
It seems like it might be helpful to document the required fields someplace. I considered adding docstrings to all of the validation functions, but in this case writing the documentation separately will probably be better than trying to make users read generated Python docs.

Just a draft so far to get your thoughts: @connorcoley


skearnes:
The best alternative that would avoid the sync issue is to move the per-message validations to a separate module and generate the pydoc from that. It would have more boilerplate (method names and signatures) but we could put this info into the docstrings so it's next to the code.

I think I still prefer keeping them separate for readability (and slightly easier use of the webform---people can look things up by the MessageName rather than `validate_message_name`).

Nearly done with my first pass through validations.py; will send out for review at that point.

skearnes:
(not for this PR): what would you think about pasting in the definition for each message here? Then the schema and the validations would be in one place and it would be easy to get permalinks to message definitions...could be really nice?

connorcoley:
I don't feel strongly about pasting the definition for each message there _until_ we reach a point where the schema is 100% finalized and versioned.

PR #161, opened 2020-05-16
Add reaction_ids to Dataset

### LINKED ISSUE #150: Can multiple datasets use the same reactions?
Suppose we have a handful of publications, where each has a `Dataset` with all of the reactions contained in that publication. Now suppose someone comes along and wants to create an "All C-N cross couplings reactions from 2019" `Dataset`. Is that an appropriate use case? Do we allow the creation of these aggregated datasets?

If no, let's make sure we aren't neglecting this possible use case of curating data subsets.

If yes, let's think about how to avoid storing redundant information and how we might link both datasets to the same reaction IDs.

skearnes:
Fixes #150 

Question: should we rename `Reaction.provenance.record_id` to `reaction_id`? I'm using `reaction_ids` instead of `record_ids` here for parity with the `reactions` field.


connorcoley:
Two thoughts:
1. Yes, we should rename `record_id` to `reaction_ids`
1. Do we need to expand the comment for `Dataset.reaction_ids` to clarify when/how it should be defined?

skearnes:
> Two thoughts:
> 
> 1. Yes, we should rename `record_id` to `reaction_ids`

OK, will do. Follow-up question: I think it would be helpful to move `reaction_id` out of `ReactionProvenance` and into `Reaction`. It's not really provenance information, and it will be helpful to have it at the top level. Do you agreee?

> 2. Do we need to expand the comment for `Dataset.reaction_ids` to clarify when/how it should be defined?

I assume you're referring to the use case of slicing and dicing from existing datasets? I'll add some more context.

skearnes:
Pushed some changes; PTAL.

michaelmaser:
Quick question: would `reaction_id` refer to a specific data entry or the reaction (i.e. transformation) itself? Thinking it could be important to distinguish a `ReactionProvenance` type of `reaction_id` (almost like a `record_id`) from a general `Reaction` type `reaction_id` to accommodate multiple entries of the same chemical transformation from different users as results could be different.

connorcoley:
Those changes look good to me.

Re: @michaelmaser 's questions: We aren't intending to follow the Reaxys convention of having "reactions" and "details", where the former are defined in terms of reactants and products and the latter are defined in terms of the conditions and outcome. It would be very challenging to test for equivalence between two entries of "the same chemical transformation" without extensive manual curation. Instead, I expect that multiple entries of the same reaction type would be retrieved together when the database is searched by reaction substructure or reaction similarity.

michaelmaser:
Got it. Yes, makes good sense.

PR #168, opened 2020-05-21
Update reaction.proto schema

michaelmaser:
Extends enums, adds messages, restructures ReactionWorkup, restructures some repeated messages, adds comments and details


connorcoley:
Thanks, @michaelmaser . I'll provide a detailed review on the proto shortly. We'll want to update many of the existing examples in the repo (which I can help with) and the `validations.py` script.

proto/reaction.proto
```diff
@@ -124,6 +126,24 @@ message ReactionInput {
...
+      UNSPECIFIED = 0;
+      CUSTOM = 1;
+      NONE = 2;
+      SYRINGE = 3;
+      CANNULA = 4;
+      ADDITION_FUNNEL = 5;
+    }
+    AdditionDeviceType type = 1;
+    // Device specifications, e.g., "gas-tight" for SYRINGE.
+    string details = 2;
```
connorcoley, line 140: I'd suggest we remove `details` from inside this `AdditionDevice` message and add a new field below, `string addition_details = 9;` which will be a catch-all for details. Examples of what could be specified include "gas-tight" for a SYRINGE `addition_device`, information about the portion sizes for a PORTIONWISE `addition_speed`, and _how_ the reaction input was heated/cooled if `addition_temperature` is specified.

proto/reaction.proto
```diff
@@ -212,9 +233,9 @@ message CompoundPreparation {
     REPURIFIED = 3;  // Compound repurified (e.g., recrystallized).
     SPARGED = 4;  // Compound sparged, most likely to be the case with solvents.
     DRIED = 5;  // Moisture removed, e.g., using molecular sieves.
-    SYNTHESIZED = 6;  // Compound synthesized in-house
+    SYNTHESIZED = 6;  // Compound synthesized in-house.
   }
-  PreparationType type = 1;
+  repeated PreparationType type = 1;
```
connorcoley, line 238: Rather than making the `PreparationType type` repeated, let's make `CompoundPreparation preparation = 7;` in the `Compound` message repeated (and rename to `preparations`). That way, each preparation type will get its own `details` field and a `CompoundPreparation` message is just a single preparative step.

proto/reaction.proto
```diff
@@ -304,11 +333,42 @@ message Vessel {
       CUSTOM = 1;
       NONE = 2;
       OVEN_DRIED = 3;
+      FLAME_DRIED = 4;
+      EVACUATED_BACKFILLED = 5;
+      PURGED = 6;
     }
   }
-  VesselPreparation.VesselPreparationType preparation = 5;
+  repeated VesselPreparation.VesselPreparationType preparation = 5;
```
connorcoley, line 341: @skearnes what is the right norm for singular/plural with repeated fields? Should it be `preparation` or `preparations`? The [examples](https://developers.google.com/protocol-buffers/docs/proto3) all use plural names for a repeated field, so I'm leaning that way.

proto/reaction.proto
```diff
@@ -304,11 +333,42 @@ message Vessel {
...
+  message VesselAttachment {
+    enum VesselAttachmentType {
+      UNSPECIFIED = 0;
+      NONE = 1;
+      CUSTOM = 2;
+      SEPTUM = 3;
+      CAP = 4;
+      REFLUX_CONDENSER = 6;
+      VENT_NEEDLE = 7;
+      DEAN-STARK = 8;
```
connorcoley, line 352: I don't believe you can use hyphens here. One way to check is to see if the protobuf compiles. If you run `python setup.py install` from the `ord-schema` folder, it will try to compile the protobuf.

Either `DEANSTARK` or `DEAN_STARK` should be fine here probably

proto/reaction.proto
```diff
@@ -319,6 +379,21 @@ message ReactionSetup {
...
+    CUSTOM = 1;
+    FUME_HOOD = 2;
+    BENCH_TOP = 3;
+    GLOVE_BOX = 4;
+    GLOVE_BAG = 5;
+    }
+    // Specify reaction environment.
+    ReactionEnvironmentType type = 1;
+    // Specify environment details, e.g., "nitrogen-filled" for GLOVE_BOX.
+    string details = 2;
```
connorcoley, line 394: Since `environment` is not repeated, we should follow the conventions used above (see `vessel_material` and `vessel_material_details` for a good example) like so:

```
message ReactionEnvironment {
  enum ReactionEnvironmentType {
    UNSPECIFIED = 0;
    CUSTOM = 1;
    FUME_HOOD = 2;
    BENCH_TOP = 3;
    GLOVE_BOX = 4;
    GLOVE_BAG = 5;
    }
}
ReactionEnvironment.ReactionEnvironmentType environment = 5;
string environment_details = 6;

```

proto/reaction.proto
```diff
@@ -375,19 +452,19 @@ message TemperatureConditions {
...
     enum PressureControlType {
       UNSPECIFIED = 0;
       CUSTOM = 1;
       AMBIENT = 2;
-      BALLOON = 3;
-      SEALED = 4; // Fully sealed vessel (e.g., microwave vial).
-      SEPTUM_WITH_NEEDLE = 5;  // Slight positive pressure maintained
-      RELEASEVALVE = 6;
-      BPR = 7; // Back pressure regulator, as used in flow synthesis.
+      CONSTANT = 3;  // Slight positive pressure, e.g., with a gas line inlet.
```
connorcoley, line 461: Does `CONSTANT` imply `SLIGHT_POSITIVE`? The latter might provide a clearer distinction with `PRESSURIZED`, but I feel like high pressure reactions are often also constant pressure.

proto/reaction.proto
```diff
@@ -541,62 +642,84 @@ message ReactionObservation {
...
+    string step_details = 2;
+    Time duration = 3;
+    repeated Compound components = 4;
+    TemperatureConditions temperature = 5;
+    string keep_phase = 6;
+    StirringConditions stirring = 7;
+    float target_ph = 8;
+  }  
+  // Specify individual steps 
+  repeated WorkupStep step = 1;
```
connorcoley, line 717: In the highest-level `Reaction` message definition, we actually already use `repeated ReactionWorkup workup`. So inside the `ReactionWorkup` message, we don't need to have a `repeated` field with a nested `WorkupStep` inner message.

proto/reaction.proto
```diff
@@ -541,62 +642,84 @@ message ReactionObservation {
...
+      // Specify temperature in "temperature".
+      // Specify pressure and apparatus notes in "details".
+      DISTILLATION = 17;
+    }
+    WorkupStepType type = 1;
+    // Include step-specific details such as purpose of an addition, e.g.,
+    // "crystallization". Include method details, e.g., "layer diffusion".
+    string step_details = 2;
+    Time duration = 3;
+    repeated Compound components = 4;
```
connorcoley, line 710: We also discussed changing `repeated Compound components = 4;` to a `ReactionInput input = 4;` so that we can include information about addition speed and addition temperature.

connorcoley: When we make that change, we should also update the comments to refer to `input` not `components`

proto/reaction.proto
```diff
@@ -662,7 +791,9 @@ message Selectivity {
     CUSTOM = 1;
     EE = 2;  // Enantiomeric excess as a percentage.
     ER = 3;  // Enantiomeric ratio. (x:1)
-    DE = 4;  // Diasteromeric ratio (x:1)
+    DR = 4;  // Diasteromeric ratio. (x:1)
+    EZ = 5;  // Alkene geometry ratio. (x:1)
+    ZE = 6;  // Alkene geometry ratio. (x:1)
```
connorcoley, line 796: Great addition. Maybe in the comments for these selectivity types, we should specify that `EE`, `ER`, and `DR` follow the convention that any product labeled as desired is the one we treat as desired (duh), but `EZ` and `ZE` follow an absolute definition where `EZ` is always the ratio of trans to cis. Either `EZ` or `ZE` can be specified.

proto/reaction.proto
```diff
@@ -716,6 +852,9 @@ message ReactionAnalysis {
   // Whether an internal standard was used with this analytical technique for
   // quantification, e.g., of yield.
   Boolean.BooleanValue uses_internal_standard = 8;
+  // Whether an authentic sample of a material was used for callibrating
+  // an analytical technique, e.g., for product identification by LCMS.
+  Boolean.BooleanValue uses_authentic_standard = 9;
```
connorcoley, line 857: I still think this is a terrific addition

proto/reaction.proto
```diff
@@ -681,12 +812,13 @@ message DateTime {
...
   // TODO(ccoley): Solicit more feedback from experimentalists
+  // TODO: Consider adding 1H/13C/OTHER_NMR fields
   enum AnalysisType {
     UNSPECIFIED = 0;
     CUSTOM = 1;
     LC = 2;  // Liquid chromatography, including HPLC and UPLC.
     GC = 3;  // Gas chromatography.
-    IR = 4;  // Infrared spectroscopy.
+    IR = 4;  // Infrared spectroscopy, including ReactIR.
     NMR = 5;  // NMR spectroscopy.
```
connorcoley, line 822: We discussed replacing `NMR` with `NMR_1H`, `NMR_13C`, and `NMR_OTHER` here since these are _so_ common. Thoughts? The comment for `NMR_OTHER` could say "specify details in `details`"

proto/reaction.proto
michaelmaser: Agreed. That sounds great. I like the comment as a whole as well.

michaelmaser: Are you making the edits or should I?

proto/reaction.proto
michaelmaser: Agreed, sounds good.

proto/reaction.proto
michaelmaser: Done.

proto/reaction.proto
michaelmaser: Done.

proto/reaction.proto
michaelmaser: Got it, thanks!

proto/reaction.proto
michaelmaser: Yes, went back and forth on this. `SLIGHT_POSITIVE` is more clear. Change made.

proto/reaction.proto
michaelmaser: Right, thanks. I like it.

proto/reaction.proto
michaelmaser: Yes, thanks for the clarification. Fixed.

proto/reaction.proto
michaelmaser: Agreed. I added definitions to the comments. Feel free to review and change as you see fit.

proto/reaction.proto
michaelmaser: Agreed. Made the change and deleted the `TODO`.

michaelmaser:
Made and pushed all the above changes from your reviews. Looks great! Let me know if there are any other comments, or feel free to edit yourself too. Thanks!

connorcoley:
There are slight inconsistencies I would like to get @skearnes to weigh in on:
- `CompoundPreparation`s contain `type` and `details`, so `repeated CompoundPreparation preparations` lets details be specified for _each_; `VesselPreparation` is simply used for the enum, so there is only one `preparation_details` shared by all.
- The `Vessel` generally follows the pattern of having one enum field `foo` and a string field `foo_details`, rather than having a submessage `Foo foo` with fields `type` and `string details`.
- Please check if singular/plural naming convention is consistent and suggest a standard here. For example, we have `repeated VesselPreparation.VesselPreparationType preparations`, `string preparation_details`, `repeated VesselAttachment.VesselAttachmentType attachments`, and `string attachments_details`

Although it means more nested messages, I actually think I prefer having a submessage with `type` and `details` for these enums-with-details. This makes it easier to use the construction...
```
message.foos.add(type='ENUM_OPTION', details='my_details')
```
rather than
```
message.foos.append(message.FooValues.ENUM_OPTION)
message.foos_details = 'my_details'
```

If you agree, I can update all of the fields as part of this PR to match that construction.

proto/reaction.proto
```diff
@@ -318,11 +349,40 @@ message Vessel {
...
+      FLAME_DRIED = 4;
+      EVACUATED_BACKFILLED = 5;
+      PURGED = 6;
     }
   }
-  VesselPreparation.VesselPreparationType preparation = 5;
+  repeated VesselPreparation.VesselPreparationType preparations = 5;
   string preparation_details = 6;
-  Volume volume = 7;  // Size (volume) of the vessel.
+  message VesselAttachment {
```
skearnes, line 359: For long-ish messages like this, I suggest moving them to top-level and adding a message-level comment describing the proto.

proto/reaction.proto
```diff
@@ -318,11 +349,40 @@ message Vessel {
...
+      GAS_ADAPTER = 15;  // E.g. a ground-glass adapter with a gas line.
+      PRESSURE_REGULATOR = 16;
+      RELEASE_VALVE = 17;
+    }   
+  }
+  repeated VesselAttachment.VesselAttachmentType attachments = 7;
+  // Include attachment specifications, e.g., "rubber" for SEPTUM,
+  // "Teflon" for CAP, "water jacket" for REFLUX_CONDENSER, etc.
+  // Also include sealing details, e.g., "electrical tape", "parafilm", etc.
+  string attachments_details = 8;
```
skearnes, line 384: I'd prefer:

```proto
message VesselAttachment {
  enum VesselAttachmentType {
    ...
  }
  VesselAttachmentType type = 1;
  string details = 2;
}
repeated VesselAttachment attachments = 7;
```

proto/reaction.proto
```diff
@@ -389,19 +465,19 @@ message TemperatureConditions {
...
     enum PressureControlType {
       UNSPECIFIED = 0;
       CUSTOM = 1;
       AMBIENT = 2;
-      BALLOON = 3;
-      SEALED = 4; // Fully sealed vessel (e.g., microwave vial).
-      SEPTUM_WITH_NEEDLE = 5;  // Slight positive pressure maintained
-      RELEASEVALVE = 6;
-      BPR = 7; // Back pressure regulator, as used in flow synthesis.
+      SLIGHT_POSITIVE = 3;  // E.g., with a gas line inlet.
```
skearnes, line 474: Just `POSITIVE`? Maybe group closer to `PRESSURIZED` and add a comment to help distinguish between these options?

proto/reaction.proto
```diff
@@ -730,6 +859,9 @@ message ReactionAnalysis {
   // Whether an internal standard was used with this analytical technique for
   // quantification, e.g., of yield.
   Boolean.BooleanValue uses_internal_standard = 8;
+  // Whether an authentic sample of a material was used for callibrating
```
skearnes, line 862: What does "authentic" mean in this context?

proto/reaction.proto
michaelmaser: I agree, I like this logic. This is actually how I have it in my fork...I may have missed a change? So based on the conversation, should _all_ sub-messages be changed to this structure?

proto/reaction.proto
michaelmaser: This is really referring to a single (extremely common) setup, written in the comment, where a gas line with a needle punctures a septum or cap on the vessel. So, while there is a "slight-positive" pressure, there's typically no intention of actually "pressurizing" the flask.

proto/reaction.proto
michaelmaser: Authentic refers to a sample of a compound whose structure has been proven by the accepted means (1H NMR, 13C NMR, IR, HRMS), and is now being used as a comparison point for other analytical methods. This way one can determine if that compound is present in a reaction in the future by quicker methods (e.g., TLC or LCMS) that are not acceptable "proof" of a structure by themselves.

michaelmaser:
@connorcoley 

> If you agree, I can update all of the fields as part of this PR to match that construction.

Does this refer to _all_ messages/submessages? I do prefer that construction. If you've started updating, feel free to apply to all.

proto/reaction.proto
michaelmaser: Done.

proto/reaction.proto
connorcoley: That's on me, Mike, as I had changed it to be consistent with our non-repeated fields. In light of our discussion, all _repeated_ fields should use the structure that Steve commented here.

proto/reaction.proto
connorcoley: +1 that `SLIGHT_POSITIVE` means something distinct from `POSITIVE`. In the common case that Mike describes, the setpoint would not be specified. `SLIGHT_POSITIVE` usually implies 1-5 psi

proto/reaction.proto
connorcoley: Let's expand the comment to be a little more explicit. Even if synthetic chemists understand, it would be useful for the informaticians to have this explanation

connorcoley:
> @connorcoley
> 
> > If you agree, I can update all of the fields as part of this PR to match that construction.
> 
> Does this refer to _all_ messages/submessages? I do prefer that construction. If you've started updating, feel free to apply to all.

I have not started updating. I will plan to work on this more starting tonight, but if you have time during the day feel free to say so.

Make sure you pull first to get an updated version of the repo; I changed a lot of files to use the new schema. I'll need to update everything again once we change how `type`/`details` are recorded.

proto/reaction.proto
michaelmaser: No worries, thanks. Updating some other ones in the specific conditions messages now.

proto/reaction.proto
michaelmaser: Done.

michaelmaser:
Okay, pushed latest changes to the proto. Think I caught all the appropriate messages to restructure. Updated some more in the specific conditions section too, e.g., `StirringConditions`, `ElectrochemistryConditions` and `FlowConditions`. Check these out and make sure they look okay. Extended some of the comments based on the discussion, and shifted the long message per @skearnes suggestion. Thanks!

proto/reaction.proto
```diff
@@ -360,6 +360,7 @@ message Vessel {
     VesselPreparationType type = 1;
     string details = 2;
   }
+<<<<<<< HEAD
```
connorcoley, line 363: @michaelmaser it looks like you've pushed conflicted merges. This is because I updated the `reaction.proto` code (and several other files) after your initial PR. The correct workflow in this case would be for you to `git pull` to update your version, make any modifications, then commit, test, and push. I assume that you started making your modifications before you pulled, and when you pushed you got a message saying your version is out of date. When you _then_ pulled, you might have gotten a warning that the merge could not be completed automatically. In this case, you need to carefully look at where the conflicts occur and manually resolve them. I'll fix all of these, but something to look out for in the future!

connorcoley:
By compartmentalizing `type`/`details` into separate messages, we've managed to clean up some of the validation functions in `validations.py`. All of the example notebooks now use the updated schema. It doesn't seem overly burdensome to have these additional fields, and I do like the consistency. In particular, the beginning of the OrgSyn example looks quite nice now:

```
reaction.setup.vessel.CopyFrom(
    reaction_pb2.Vessel(
        type=dict(type='ROUND_BOTTOM_FLASK', details='single-necked'), 
        material=dict(type='GLASS'),
        volume=unit_resolver.resolve('500 mL'),
    )
)
reaction.setup.vessel.preparations.add(type='OVEN_DRIED')
reaction.setup.vessel.preparations.add(type='PURGED', details='with nitrogen')
reaction.setup.vessel.attachments.add(type='SEPTUM', details='rubber')
```

proto/reaction.proto
michaelmaser: That's exactly what happened. I had the proto open for editing not realizing there were changes made. Then thought I caught them after I got the error, but clearly didn't. Apologize for the confusion. Thanks.

michaelmaser:
That's great, it looks awesome. Glad it worked out.

skearnes:
This is great; I'm really happy with the move to consistent `type`/`details` messages.

connorcoley:
@michaelmaser would you like to do the honors and merge?

michaelmaser:
Yes, I'd be happy to. I think I need write access, though?

PR #175, opened 2020-05-29
Add integer_value to Data

skearnes:
Mostly for completeness.

Question: Is there any reason we might want the float/integer values to be `repeated`? We could ditch the oneof and enforce mutual exclusivity with validations.

### DIFF
--- a/proto/reaction.proto
+++ b/proto/reaction.proto
@@ -1100,18 +1100,20 @@ message Percentage {
   optional float precision = 2;
 }
 
-// Data is a container for arbitrary string or bytes data.
+// Data is a container for arbitrary data.
 message Data {
   oneof kind {
-    string string_value = 1;
-    float float_value = 2;
+    float float_value = 1;
+    int32 integer_value = 2;
     bytes bytes_value = 3;
-    string url = 4;  // URL for data stored elsewhere.
+    string string_value = 4;
+    string url = 5;  // URL for data stored elsewhere.
   }
-  string description = 5;
+  // Description of these data (but avoid redundancy with map keys).
+  string description = 6;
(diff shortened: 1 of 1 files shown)


connorcoley:
I would support the change because it makes `Data` a more flexible message. 

Analytical data are going to be the majority of the `Data` entries we'll get. It could be nice to record processed 13C NMR data, IR peaks, etc. as arrays of floats. This provides just a little bit more structure. It's not that important, since most processed analytical data will have to be stored as semi-structured text (e.g., 1H NMR data).

For uniformity, would you make all of the `*_value` fields `repeated`?

skearnes:
> For uniformity, would you make all of the `*_value` fields `repeated`?

Hmm...that's a bit tricky. For instance, if we allow repeated `bytes`, it's not clear how to do the export to LFS/GCS for offloading. I think I'd prefer to make only the numeric fields repeated. 

We could also do something potentially confusing like:

```proto
message Data {
  float float_value = 1;
  repeated float float_list = 2;
  ...
}
```

What do you think?

connorcoley:
I'd prefer to avoid having that construction. With that logic, you could think about having a `map<string,float> float_map` field too.

Perhaps we stick to non-repeated fields, since we're never going to have good coverage of structured data without something like PR#51.

skearnes:
> I'd prefer to avoid having that construction. With that logic, you could think about having a `map<string,float> float_map` field too.
> 
> Perhaps we stick to non-repeated fields, since we're never going to have good coverage of structured data without something like PR#51.

SGTM, thanks.

skearnes:
(Note that this is going into the new `dev` branch.)

Issue #156, opened 2020-05-14
Ambiguous defaults in the schema


skearnes:
Proto3 changed a lot of things, including the ability to [check that non-message fields are set to their default values](https://developers.google.com/protocol-buffers/docs/proto3#default) and [set custom defaults](https://stackoverflow.com/a/33229024).

However, I don't think this will be a problem for us since most (all?) of the ambiguous fields are contained in submessages that *do* indicate when they have been set to their defaults. For example:

```python
message = reaction_pb2.ReactionOutcome()
print(message.HasField('conversion')  # --> False
message.conversion.value = 0.0
print(message.HasField('conversion')  # --> True
```

So even though the `value` is not serialized because it matches the default, the `conversion` field *is* serialized as an empty (default) message:

```python
print(message)  # --> conversion {}

**Comment by connorcoley** [comment]
That's good to know. And this behavior (the ability to check with `HasField`) is retained after serialization/deserialization? 

The other question would be whether this significantly detract from the readability of the pbtxts and if, upon exporting the data to BigQuery for visualization/search, BigQuery can distinguish these two cases

**Comment by skearnes** [comment]
> That's good to know. And this behavior (the ability to check with `HasField`) is retained after serialization/deserialization?
> 

Yes:

```python
new = reaction_pb2.ReactionOutcome()
print(new.HasField('conversion'))  # --> False
text_format.Parse(text_format.MessageToString(message), new)
print(new.HasField('conversion'))  # --> True
```

> The other question would be whether this significantly detract from the readability of the pbtxts and if, upon exporting the data to BigQuery for visualization/search, BigQuery can distinguish these two cases

Good questions. I don't think the readability is bad enough to justify moving to JSON...do you agree? I'll need to check for BigQuery; worst case we can manually fill in the default value when generating the BigQuery input JSONL file.

connorcoley:
> I don't think the readability is bad enough to justify moving to JSON...do you agree?

I agree, as long as we can provide a sufficiently clear explanation/documentation for users. I think for a casual user looking to scrape data from a `Reaction`, their instinct will be to look at `outcome.conversion.value` rather than first check `if outcome.HasField('conversion')`

How averse are you to defining a default values in general? Just because yield/conversion are going to be so frequently used, and because we would love to capture 0% conversion and 0% yield cases, I might still suggest a default value of -100 for a Percentage value.

michaelmaser:
[...]

PR #177, opened 2020-05-31
Add FrozenMessage

skearnes:
Adds the `FrozenMessage` class, which does not allow for assignment to message fields. When used with `optional` field labels (see #174), this avoids many of the issues with ambiguous defaults raised in #156.

Question: Raising exceptions is relatively safe but also noisy. Will this match your target use case or would you prefer something that will require fewer try/except blocks? (See my note in the FrozenMessage docstring about returning None values.)

### DIFF
--- a/ord_schema/frozen_message.py
+++ b/ord_schema/frozen_message.py
@@ -19,11 +19,13 @@ from typing import Union
 
 import google.protobuf.message
 
-_MESSAGE_TYPES = (collections.abc.MutableMapping,
-                  google.protobuf.message.Message)
+_MESSAGE_TYPES = (
+    collections.abc.MutableMapping,  # Proto map.
+    google.protobuf.message.Message,  # Generic submessage.
+)
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass(eq=True, frozen=True)
 class FrozenMessage(collections.abc.Mapping):
     """Container for a protocol buffer that does not allow edits.
 
@@ -71,6 +73,8 @@ class FrozenMessage(collections.abc.Mapping):
         value = getattr(self._message, name)
         if isinstance(value, _MESSAGE_TYPES):
             return FrozenMessage(value)
+        if hasattr(value, 'append'):
+            return tuple(value)  # Make repeated fields immutable.
         return value
 
     def __iter__(self):
--- a/ord_schema/frozen_message.py
+++ b/ord_schema/frozen_message.py
@@ -74,7 +74,14 @@ class FrozenMessage(collections.abc.Mapping):
         if isinstance(value, _MESSAGE_TYPES):
             return FrozenMessage(value)
         if hasattr(value, 'append'):
-            return tuple(value)  # Make repeated fields immutable.
+            # Make repeated fields (and their elements) immutable.
+            values = []
+            for v in value:
+                if isinstance(v, _MESSAGE_TYPES):
+                    values.append(FrozenMessage(v))
+                else:
+                    values.append(v)
+            return tuple(values)
         return value
 
     def __iter__(self):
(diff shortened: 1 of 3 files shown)


skearnes:
Hold off on this; I'm not sure I'm handling `repeated` fields properly.

connorcoley:
Sounds good. I was also trying to think about the None v. error behavior
and what would be best.

skearnes:
PTAL

ord_schema/frozen_message_test.py
```diff
@@ -0,0 +1,116 @@
...
+from ord_schema import frozen_message
+from ord_schema.proto import reaction_pb2
+
+
+class FrozenMessageTest(absltest.TestCase):
+    def test_scalar_without_optional(self):
+        """This is what happens if scalar fields are not labeled `optional`."""
+        frozen = frozen_message.FrozenMessage(
+            reaction_pb2.StirringConditions.StirringRate())
+        self.assertTrue(hasattr(frozen, 'rpm'))  # Wrong!
```
connorcoley, line 29: Confused by `# Wrong!` here. This seems like it's the expected (but undesired) behavior when we're not using the `optional` field modifier or proto2. Is this just a note so we know to revise this  test later?

ord_schema/frozen_message.py
```diff
@@ -0,0 +1,111 @@
...
+        try:
+            if not self._message.HasField(name):
+                raise AttributeError(f'attribute "{name}" has not been set')
+        except ValueError:
+            pass  # The request attribute is not a submessage.
+        value = getattr(self._message, name)
+        if isinstance(value, _MESSAGE_TYPES):
+            return FrozenMessage(value)
+        if hasattr(value, 'append'):
+            return tuple(value)  # Make repeated fields immutable.
```
connorcoley, line 77: Does this need to recurse and return `tuple(map(lambda v: FrozenMessage(v) if isinstance(v, _MESSAGE_TYPES) else v, values))` or `tuple(FrozenMessage(v) if isinstance(v, _MESSAGE_TYPES) else v for v in values)`?

ord_schema/frozen_message_test.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def test_set_submessage(self):
+        frozen = frozen_message.FrozenMessage(reaction_pb2.Reaction())
+        with self.assertRaises(AttributeError):
+            frozen.setup.automation_platform = 'test'
+        with self.assertRaises(AttributeError):
+            frozen.setup.CopyFrom(
+                reaction_pb2.ReactionSetup(automation_platform='test'))
+
+    def test_set_repeated_scalar(self):
```
connorcoley, line 96: Function name doesn't match test

ord_schema/frozen_message.py
connorcoley:
```
>>> message = reaction_pb2.Reaction()
>>> message.workup.add(type=2)
type: ADDITION

>>> frozen = frozen_message.FrozenMessage(message)
>>> frozen
FrozenMessage(_message=workup {
  type: ADDITION
}
)
>>> frozen.workup[0].type = 3
>>> frozen
FrozenMessage(_message=workup {
  type: TEMPERATURE
}
)
```

ord_schema/frozen_message_test.py
skearnes: It's just a disclaimer that fields not marked `optional` will have inappropriate behavior. I thought it might be nice to illustrate the problem we are trying to solve.

ord_schema/frozen_message.py
skearnes: Good catch; I'll revise.

ord_schema/frozen_message_test.py
skearnes: Changed to `test_add_repeated_scalar`; I was going for consistency with the other names, but this is better.

ord_schema/frozen_message_test.py
skearnes: I can remove or expand the comment; what would you like me to do here?

connorcoley: Expanding the comment would be helpful, it's good to have it in there

skearnes:
PTAL

PR #180, opened 2020-05-31
ORD web editor

antonkast-google:
Adds a top-level folder called "edit". Does not modify existing files.

A README explains how to build, run and test.


skearnes:
Awesome! Thanks Anton, this is going to be great. A couple of minor suggestions:

* Perhaps change the top-level folder name to `editor`?
* Consider renaming the `db` folder to something like `testdata` to make its purpose clearer?
* For JS files that are mapped to specific message types, consider renaming to include the full message name and pluralization; e.g. `reaction_setup.js` instead of `setups.js`?

antonkast-google:
"edit" -> "editor": Done.

"db" -> "testdata":
This directory does hold test data, but it is not only for tests. It holds the user's work at runtime, and the server writes temp files there like uploads and locks. The server uses files, and I'd like to avoid requiring it to distinguish test and production paths.

Does "db" seem like a poor name for this? Maybe it's more like unix "var"? Or "work"? Or "tmp"?

"setups.js" -> "reaction_setup.js" etc.
I am not sure that suggesting a correspondence between proto messages and source files would be helpful. The JS disaggregation is based on file sizes more than any other rationale, and most of the files represent data rather than messages (e.g. inputs.js handles a map field in Reaction).

If the goal is to make it easier to navigate from schema to code, then maybe the most effective way would be through DOM class names. Every symbol in the schema appears in DOM classes, and it is easy to go from a DOM class to its UI elements and to code that touches those elements in JS. Could that achieve the same end?

skearnes:
> "edit" -> "editor": Done.
> 
> "db" -> "testdata":
> This directory does hold test data, but it is not only for tests. It holds the user's work at runtime, and the server writes temp files there like uploads and locks. The server uses files, and I'd like to avoid requiring it to distinguish test and production paths.

SGTM
> 
> Does "db" seem like a poor name for this? Maybe it's more like unix "var"? Or "work"? Or "tmp"?

No, I think `db` is fine for this.
> 
> "setups.js" -> "reaction_setup.js" etc.
> I am not sure that suggesting a correspondence between proto messages and source files would be helpful. The JS disaggregation is based on file sizes more than any other rationale, and most of the files represent data rather than messages (e.g. inputs.js handles a map field in Reaction).
> 
> If the goal is to make it easier to navigate from schema to code, then maybe the most effective way would be through DOM class names. Every symbol in the schema appears in DOM classes, and it is easy to go from a DOM class to its UI elements and to code that touches those elements in JS. Could that achieve the same end?

OK, I trust your judgment here. Using the DOM to navigate to code sounds reasonable, especially if that's a typical workflow for editing JS (I've never done any real JS development---the comment about files was mostly driven by my experience with backend development).

antonkast-google:
Hi Connor: I imagine this PR is scroll-blinding, so I'd be pleased to meet and talk about it if that's of interest to you.

connorcoley:
This looks great. I'm still digesting the organization of the code, but it's easy to follow.

For some reason, I had to change the makefile to just use `protoc` (installed through homebrew) rather than `$(protobuf)/src/protoc` with the unpacked director inside `/editor`. to successfully `make`.

skearnes:
> This looks great. I'm still digesting the organization of the code, but it's easy to follow.
> 
> For some reason, I had to change the makefile to just use `protoc` (installed through homebrew) rather than `$(protobuf)/src/protoc` with the unpacked director inside `/editor`. to successfully `make`.

Ah good catch; yes please just use whatever protoc is in the PATH if possible...

antonkast-google:
I agree the protoc binary is not critical, but I believe the full protobuf distribution is still necessary for its JS compilation dependencies.

skearnes:
> I agree the protoc binary is not critical, but I believe the full protobuf distribution is still necessary for its JS compilation dependencies.

Note that if people run `pip install -r requirements.txt` they will get `protoc` via protoc-wheel-0.

antonkast-google:
Added README to build protobuf, which would have saved Connor some work, and explained it's also fine to use a pre-installed protoc so long as the protobuf distribution is still available to provide the needed JS common files.

skearnes:
Thanks for expanding the README; this LGTM.

connorcoley:
> Added README to build protobuf, which would have saved Connor some work

Ah, that makes sense.

PR #182, opened 2020-06-01
Add tests for editor to CI

skearnes:


skearnes:
Probably don't merge this until we have a fix for #183

Issue #183, opened 2020-06-02
Simplify editor build


skearnes:
* I think we only need a protoc that supports both `py_out` and `js_out`; we might consider simply downloading a binary from GitHub or putting a wheel on pypi instead of building from scratch. This would also avoid the issue of having incompatible `protobuf` installations...
* A one-touch script would be nice, if possible. It could handle the protoc version/output checking, downloading closure, and building the editor.

@antonkast-google

antonkast-google:
To build the JS resources used by the editor, we need more than "protoc --js_out ...". The generated JS declares dependencies on shared protobuf JS namespaces. These namespaces must be resolved by the Closure JS compiler at a later stage, and the only way I have seen to get them is through the protobuf source distribution.

Is it clear that a one-step build is worthwhile? It seems to me this repo supports multiple use cases, and I could imagine that a user intending data entry could be better served by a completed artifact like a web site or desktop app.

I feel different about conflicting protobuf versions. That is just a trap, and I would fix it with build logic to detect the condition and emit an explanatory error.

antonkast-google:
Google also distributes these shared JS dependencies through npm: https://www.npmjs.com/package/google-protobuf

If we are prepared to make the build depend on npm, then that's another way to resolve things.

skearnes:
Ah interesting; I don't have any problem depending on npm for the editor (the tests already do).

antonkast-google:
Amazingly, the protobuf common JS in the npm "google-protobuf" package is different from the common JS in the protobuf distribution itself.

The former supports both the Closure style of dependency declarations, "goog.provide() / goog.require()" and also the CommonJS pattern, "require() / module.exports". The latter is documented to support both but in fact comes only with CommonJS declarations.

My understanding is that CommonJS has been dead as a project for five years and I don't know how to compile it.

skearnes:
Largely fixed in #211

PR #185, opened 2020-06-03
Update the editor Makefile for easier builds.

antonkast-google:
See Issue #183.

* Use pip's protoc for proto compilation to avoid compiling protobuf.

* Reference protobuf version 3.12.0 to match the current pip version.

* Assert that these versions match in a prerequisite make target.

* Prevent Closure from generating empty object files at compiler errors.


skearnes:
Thanks Anton!

PR #187, opened 2020-06-04
Add run_tests.sh

skearnes:
* Adds a script for running all python tests under `ord_schema/`. The script filters out `stdout` and reduces the logging verbosity so the console output is easier to read.
* Spot fixes to remove warning messages to further clean up `stderr` for test outputs.


connorcoley:
Nice, the script is a convenient addition

PR #195, opened 2020-06-10
Add terms of use

skearnes:
* Adds terms of use to CONTRIBUTING.
* The lawyers who wrote the terms recommended the "ORD Project" to distinguish from the database itself, so I changed the license headers to match.


CONTRIBUTING.md
```diff
@@ -51,3 +51,79 @@ Excellent! There are a few steps you'll need to follow to get ready to submit ch
...
+<p align="center">Open Reaction Database Project License Agreement</p> 
+
+In order to clarify the intellectual property license granted with Contributions
+to the Open Access Reaction Database from any person or entity, the Open
+Reaction Database Project ("ORDP") must indicate agreement to the license terms
+below. 
+
+This version of the Agreement allows an entity (the "Corporation") to
+submit Contributions to ORDP, to authorize Contributions submitted by its
+designated employees to ORDP, and to grant copyright and patent licenses
```
connorcoley, line 69: Does this mean that when somebody submits, they cannot do it on behalf of themselves but rather on behalf of their employer?

skearnes: PTAL

PR #198, opened 2020-06-12
Revert "Revert 'Add optional labels to float fields'"

### LINKED ISSUE #156: Ambiguous defaults in the schema

skearnes:
Reverts #176 and fixes #156 

This will break the editor tests; Anton has given us the go-ahead to unblock field presence for programmatic submissions.


skearnes:
Thanks; I'll add a commit that disables that test in the CI and create an issue to track re-enabling it.

PR #202, opened 2020-06-12
Merge dev into master

skearnes:


proto/reaction.proto
```diff
@@ -1100,18 +1100,20 @@ message Percentage {
...
 }
 
-// Data is a container for arbitrary string or bytes data.
+// Data is a container for arbitrary data.
 message Data {
   oneof kind {
-    string string_value = 1;
-    float float_value = 2;
+    float float_value = 1;
+    int32 integer_value = 2;
```
antonkast-google, line 1107: Do we have no persisted data yet?

skearnes: There's no persistent data in `ord-submissions-test` yet, if that's what you're asking. We'll blow away the stuff there and regenerate after these changes go in just to make sure everything is clean.

PR #204, opened 2020-06-12
Remove references to 'dev' branch

skearnes:
After chatting with Anton, we felt like having two shared branches (`master` and `dev`) was a bit of a headache. Instead, we propose that PRs be merged into `master` and that we use the GitHub Releases mechanism for versioning.

If we need to patch a release off of the `master` branch, we can create a new branch and a new release that points to that branch. This should simplify PRs and merges in the future since there will only be one "trunk" instead of two.

WDYT?


antonkast-google:
It seems fine to me to make shirt-lived branches for individual PRs, if
that's easier than using forks.

I just thought that a shared, long-loved branch getting reused all the time
by multiple people would take a lot of bookkeeping.

Anton

PR #207, opened 2020-06-15
Update editor for proto3 "optional".

antonkast-google:
The editor requires HEAD protobuf to get experimental JS support for proto3
"optional" declarations. As a stopgap until this support is released, we are
adding README instructions to build protobuf from HEAD. To permit automated
tests, we are also adding protoc binaries for Mac and Linux with matching JS
runtime files. These were built at 1dae8fdd.

I do not have a way to compile for Windows, ARM, or other environments. Some
users may still have to compile protobuf themselves.

This change also ends the redundant compilation of protos to python. The editor Makefile now pulls generated python modules from the setup.py build in the root directory.


editor/README.md
```diff
@@ -25,11 +25,22 @@ Serving depends on:
...
+$ cd protobuf
+$ ./autogen.sh && ./configure && make
+````
+
+The editor has been tested with [Closure
+v20200517](https://github.com/google/closure-library/releases/).
+
+Unpack both protobuf and closure-library in this directory so that make can
+find them. For the sake of automated testing, protobuf at 1dae8fdd has
+already been built for MacOS and its JS runtime and statically linked protoc
```
skearnes, line 42: I think we'll need a linux binary for the GitHub actions testing.

editor/protobuf/js/binary/constants.js
```diff
@@ -0,0 +1,376 @@
+// Protocol Buffers - Google's data interchange format
```
skearnes, line 1: Is there any reason to check these in rather than requiring the user to download the source? I'm not super excited about bloating the repo with vendored third party deps or worrying about keeping them in sync with HEAD.

editor/README.md
antonkast-google: Sorry, the text was not correct. There is a Linux binary, but I couldn't do Windows.

editor/protobuf/js/binary/constants.js
antonkast-google: The only reason to check these in is to permit the automated test.

If it could be sufficient for the user to download the protobuf JS runtime, then they could download a protoc binary at the same time and we could abandon this PR.

Could the need for this JS runtime complicate automated testing even after presence detection is released? Are we barking up the wrong tree?

skearnes: We can download arbitrary files during the automated tests (and we already download quite a lot of things in the current tests), so I think I'd prefer not including "test-only" files in the repo if there will be any associated maintenance burden. 

That said, if checking in this small set of extra files will make it much easier for _users_ to spin up the editor, that might be worth it. In that case, should we consider creating an "editor/third_party" folder to clearly separate these files?

antonkast-google: As discussed, I have reverted the added protobuf dependencies and published them instead at https://storage.cloud.google.com/open-reaction-database/editor/test/protobuf_1dae8fdd.tgz.

Now hopefully someone else can update the GitHub CI testing to unpack that archive in the "editor" directory before running "make test".

PR #214, opened 2020-06-16
Replace BooleanValue with boolean values

skearnes:
* Removes `Boolean` message in favor of `optional bool` fields.
* Updates tests and colab notebooks accordingly.

FYI the diff is a lot easier to see if you turn off whitespace diffs (use the "gear" in the `Files changed` tab). I updated the colab badge script so we shouldn't have to deal with this in future PRs.


skearnes:
FYI @antonkast-google this breaks the editor since it removes the `Boolean` message. Should we merge now and fix later? Or feel free to add a commit to this PR or tell me what to change :)

antonkast-google:
I'm glad to add a commit, but I can't do it today, and I don't think I can describe the fix without looking more closely.

Its OK with me to merge the breakage if that would be useful. Otherwise I'll get to this in a day or so and report back.

skearnes:
I think I'll merge this now so I can update `ord-submissions-test` and start playing with more BigQuery stuff.

PR #215, opened 2020-06-18
Restore the editor test.

antonkast-google:
This adds expression in the editor for the new proto3 "optional" field semantics.

In the case of "optional bool", the fields are represented as before by a three-way popup, "TRUE / FALSE / UNSPECIFIED". In the case of "optional float", the load logic first checks whether the field is defined before initializing its text, and the unload logic treats the empty string as an absent field.

The editor test is again passing.

### DIFF
--- a/editor/.gitignore
+++ b/editor/.gitignore
@@ -7,3 +7,4 @@ gen
 node_modules
 package
 package-lock.json
+protobuf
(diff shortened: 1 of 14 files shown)


skearnes:
Thanks Anton!

PR #218, opened 2020-06-19
Multistep reaction capabilities

connorcoley:
As a result of our [discussions](https://docs.google.com/document/d/1eO5-2E4RXCpG2bdMGo9R-JnEC92uppPBlswAscVrtlE/edit#) about how to best handle multi-step/stage reactions, this adds that capability.

Use cases are as follows:
1. For a multi-step reaction sequence with complete intermediate isolations and characterizations:
- A string reaction_id field is added to the CompoundPreparation message. This is to be used when the preparation type is SYNTHESIZED and the preparation of that compound is also recorded in the database. Note that this is not to be used when the preparation merely follows a previously described process; the isolated product of the previous reaction should be what is used, physically, in this reaction.
2. For a one-pot reaction sequence with no intermediate isolation:
- For all but the final step (i.e., reaction steps without analysis/characterization of the crude product):
Reactions have a single ReactionOutcome defined at the appropriate reaction_time.
- A putative product structure (key intermediate) can be defined as a ReactionProduct, with no supporting analytical data used to confirm its identity (i.e., analysis_identity is empty); yield, purity, and selectivity would be undefined as well.
3. For all but the first step (i.e., reaction steps that use the output of a previous reaction step without complete isolation and characterization):
- A new repeated CrudeComponent crude_components field is added to the ReactionInput message to indicate that crude (or partially worked up) was used in the reaction.

The key additions in this PR are as follows:
- Addition of a `CrudeComponent` message to the schema and a `repeated CrudeComponent crude_components` field for a `ReactionInput`
- Addition of a `reaction_id` to `CompoundPreparation` to be used when `SYNTHESIZED`
- Validations and validation tests to ensure that cross-referenced reaction IDs are all defined within a single dataset
- Addition of a `validate_ids` keyword argument to `validate_message` so that we can enforce well-formed IDs only after assigning them through `ord_schema.updates`
- Addition of an `update_dataset` function in `updates.py` that is used by `process_dataset.py`, which ensures that when reaction IDs are assigned, we still preserve any cross-references that existed within the `Dataset`.
- Addition of a test for `update_dataset` which confirms that cross-references are preserved.

### DIFF
--- a/ord_schema/updates.py
+++ b/ord_schema/updates.py
@@ -16,6 +16,7 @@
 import datetime
 import urllib
 import uuid
+import re
 
 from absl import logging
 from rdkit import Chem
@@ -120,21 +121,31 @@ def update_reaction(reaction):
 
     Current updates:
       * Sets reaction_id if not already set.
+      * Adds a record modification event to the provenance.
+      * Resolves compound identifier names if no structural identifiers are
+        defined for a given compound.
 
     Args:
         reaction: reaction_pb2.Reaction message.
+
+    Returns:
+        A dictionary mapping placeholder reaction_ids to newly-assigned
+            reaction_ids.
     """
     modified = False
+    id_substitutions = {}
     if not reaction.provenance.HasField('record_created'):
         reaction.provenance.record_created.time.value = (
             datetime.datetime.utcnow().ctime())
         modified = True
-    if not reaction.reaction_id:
+    if not re.fullmatch('^ord-[0-9a-f]{32}$', reaction.reaction_id):
         # NOTE(kearnes): This does not check for the case where a Dataset is
         # edited and reaction_id values are changed inappropriately. This will
         # need to be either (1) caught in review or (2) found by a complex
         # check of the diff.
         reaction_id = f'ord-{uuid.uuid4().hex}'
+        if reaction.reaction_id:
+            id_substitutions[reaction.reaction_id] = reaction_id
         reaction.reaction_id = reaction_id
         modified = True
     for func in _UPDATES:
@@ -146,6 +157,38 @@ def update_reaction(reaction):
         event = reaction.provenance.record_modified.add()
         event.time.value = datetime.datetime.utcnow().ctime()
         event.details = 'Automatic updates from the submission pipeline.'
+    return id_substitutions
+
+
+def update_dataset(dataset):
+    """Updates a Dataset message.
+
+    Current updates:
+      * All reaction-level updates in update_reaction.
+      * reaction_id cross-references between Reactions in the dataset.
+
+    Args:
+        dataset: dataset_pb2.Dataset message.
+    """
+    # Reaction-level updates
+    id_substitutions = {}
+    for reaction in dataset.reactions:
+        id_substitutions.update(update_reaction(reaction))
+    # Dataset-level updates of cross-references
+    # Note: if the Dataset has been validated, then we know that all
+    # cross-referened reaction_ids correspond to placeholder reaction_ids of
+    # another Reaction in the Dataset. If not validated, this update might
+    # raise a KeyError exception.
+    for reaction in dataset.reactions:
+        for reaction_input in reaction.inputs:
+            for component in reaction_input.components:
+                for preparation in component.preparations:
+                    if preparation.reaction_id:
+                        preparation.reaction_id = id_substitutions[
+                            preparation.reaction_id]
+            for crude_component in reaction_input.crude_components:
+                crude_component.reaction_id = id_substitutions[
+                    crude_component.reaction_id]
 
 
 # Standard updates.
--- a/ord_schema/updates.py
+++ b/ord_schema/updates.py
@@ -180,7 +180,7 @@ def update_dataset(dataset):
     # another Reaction in the Dataset. If not validated, this update might
     # raise a KeyError exception.
     for reaction in dataset.reactions:
-        for reaction_input in reaction.inputs:
+        for reaction_input in reaction.inputs.values():
             for component in reaction_input.components:
                 for preparation in component.preparations:
                     if preparation.reaction_id:
--- a/ord_schema/updates.py
+++ b/ord_schema/updates.py
@@ -169,16 +169,17 @@ def update_dataset(dataset):
 
     Args:
         dataset: dataset_pb2.Dataset message.
+
+    Raises:
+        KeyError: if the dataset has not been validated and there exists a
+            cross-referenced reaction_id in any Reaction that is not defined
+            elsewhere in the Dataset.
     """
     # Reaction-level updates
     id_substitutions = {}
     for reaction in dataset.reactions:
         id_substitutions.update(update_reaction(reaction))
     # Dataset-level updates of cross-references
-    # Note: if the Dataset has been validated, then we know that all
-    # cross-referened reaction_ids correspond to placeholder reaction_ids of
-    # another Reaction in the Dataset. If not validated, this update might
-    # raise a KeyError exception.
     for reaction in dataset.reactions:
         for reaction_input in reaction.inputs.values():
             for component in reaction_input.components:
--- a/ord_schema/updates_test.py
+++ b/ord_schema/updates_test.py
@@ -57,7 +57,7 @@ class UpdateReactionTest(absltest.TestCase):
     def test_with_no_updates(self):
         message = reaction_pb2.Reaction()
         message.provenance.record_created.time.value = '2020-05-08'
-        message.reaction_id = 'ord-test'
+        message.reaction_id = 'ord-c0bbd41f095a44a78b6221135961d809'
         copied = reaction_pb2.Reaction()
         copied.CopyFrom(message)
         updates.update_reaction(copied)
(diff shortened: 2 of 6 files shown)


ord_schema/updates_test.py
```diff
@@ -57,7 +58,7 @@ def test_with_updates_simple(self):
     def test_with_no_updates(self):
         message = reaction_pb2.Reaction()
         message.provenance.record_created.time.value = '2020-05-08'
-        message.reaction_id = 'ord-test'
+        message.reaction_id = 'ord-c0bbd41f095a44a78b6221135961d809'
```
connorcoley, line 61: FYI, this was changed because the update function will only preserve an existing `reaction_id` if it does not appear malformed

ord_schema/validations.py
```diff
@@ -151,7 +171,11 @@ def validate_message(message, recurse=True, raise_on_error=True):
             f"Don't know how to validate {type(message)}")
 
     with warnings.catch_warnings(record=True) as tape:
-        _VALIDATOR_SWITCH[type(message)](message)
+        if validate_ids and isinstance(
+                message, (reaction_pb2.Reaction, dataset_pb2.Dataset)):
+            _VALIDATOR_SWITCH[type(message)](message, validate_id=True)
```
connorcoley, line 176: I don't _love_ how this turned out, but it was the most elegant way I could think of to have an easy switch between...
1. Validating where we don't care if IDs are malformed
1. Validating where we want to raise an error if IDs are malformed

ord_schema/validations.py
```diff
@@ -261,10 +286,39 @@ def validate_dataset(message):
         for reaction_id in message.reaction_ids:
             if not re.fullmatch('^ord-[0-9a-f]{32}$', reaction_id):
                 warnings.warn('Reaction ID is malformed', ValidationError)
-    if message.dataset_id:
+    if validate_id:
         # The dataset_id is a 32-character uuid4 hex string.
         if not re.fullmatch('^ord_dataset-[0-9a-f]{32}$', message.dataset_id):
             warnings.warn('Dataset ID is malformed', ValidationError)
+    # Check cross-references
```
connorcoley, line 293: Let me know if there is a more concise way to do these checks. I realize this is an absurd level of nesting, but I find it legible at least.

ord_schema/validations.py
```diff
@@ -562,14 +640,21 @@ def validate_reaction_outcome(message):
                         f'Undefined analysis key {key} '
                         'in ReactionProduct', ValidationError)
     if not message.products and not message.HasField('conversion'):
-        # TODO(kearnes): Should products and conversion be mutually exclusive?
         warnings.warn(
             'No products or conversion are specified for reaction;'
-            ' at least one must be specified', ValidationError)
+            ' this is permissible only for multistep reactions',
+            ValidationWarning)
```
connorcoley, line 646: Note: while I changed this to a warning, we _don't_ check whether this reaction ID is used in a CrudeComponent of another reaction in the dataset. I felt like that was an unnecessarily complicated check

ord_schema/updates.py
```diff
@@ -146,6 +157,38 @@ def update_reaction(reaction):
...
+    """
+    # Reaction-level updates
+    id_substitutions = {}
+    for reaction in dataset.reactions:
+        id_substitutions.update(update_reaction(reaction))
+    # Dataset-level updates of cross-references
+    # Note: if the Dataset has been validated, then we know that all
+    # cross-referened reaction_ids correspond to placeholder reaction_ids of
+    # another Reaction in the Dataset. If not validated, this update might
+    # raise a KeyError exception.
```
skearnes, line 181: Please add a `Raises` section to the docstring; http://google.github.io/styleguide/pyguide.html#doc-function-raises

ord_schema/updates_test.py
skearnes: SGTM

ord_schema/validations.py
```diff
@@ -27,7 +27,7 @@
 from ord_schema.proto import reaction_pb2
 
 
-def validate_datasets(datasets, write_errors=False):
+def validate_datasets(datasets, write_errors=False, validate_ids=False):
```
skearnes, line 30: Please add `validate_ids` to the docstring.

ord_schema/validations.py
```diff
@@ -55,40 +58,51 @@ def validate_datasets(datasets, write_errors=False):
             f'validation encountered errors:\n{error_string}')
 
 
-def _validate_dataset(filename, dataset):
-    """Validates Reaction messages in a Dataset.
-
-    Note that validation may change the message. For example, NAME
-    identifiers will be resolved to structures.
+def _validate_dataset(dataset, label='dataset', validate_ids=False):
```
skearnes, line 61: Please add `validate_ids` to the docstring.

ord_schema/validations.py
skearnes: LGTM, although I have a slight preference for the default being (2) and needing to explicitly disable it for (1). This matches the default strictness of `raise_on_error=True`.

ord_schema/validations.py
```diff
@@ -261,10 +286,39 @@ def validate_dataset(message):
...
+    for reaction in message.reactions:
+        if reaction.reaction_id:
+            if reaction.reaction_id in defined_ids:
+                warnings.warn(
+                    'Multiple Reactions should never have the same '
+                    'IDs', ValidationError)
+            defined_ids.add(reaction.reaction_id)
+        for reaction_input in reaction.inputs.values():
+            for component in reaction_input.components:
+                for preparation in component.preparations:
```
skearnes, line 305: Alternatively, use `message_helpers.find_submessages` instead of all this nesting.

ord_schema/validations.py
```diff
@@ -261,10 +286,39 @@ def validate_dataset(message):
...
+                            warnings.warn(
+                                'A Reaction should not reference '
+                                'its own ID', ValidationError)
+            for crude_component in reaction_input.crude_components:
+                referenced_ids.add(crude_component.reaction_id)
+                if crude_component.reaction_id == reaction.reaction_id:
+                    warnings.warn(
+                        'A Reaction should not reference '
+                        'its own ID', ValidationError)
+    if len(referenced_ids - defined_ids) > 0:
```
skearnes, line 318: Prefer to test without `len`; `if referenced_ids - defined_ids` should be enough.

ord_schema/validations.py
skearnes: This is good; maybe factor out as a separate method just to keep `validate_dataset` sane?

ord_schema/validations.py
```diff
@@ -326,6 +380,23 @@ def validate_addition_speed(message):
     del message  # Unused.
 
 
+def validate_crude_component(message):
+    if not message.reaction_id:
+        warnings.warn('CrudeComponents must specify a reaction_id',
+                      ValidationError)
+    if (message.HasField('has_derived_amount') and message.has_derived_amount
```
skearnes, line 387: It's a bit sloppy, but it should be safe to do `if message.has_derived_amount` since you'll never get `True` unless that was set explicilty.

ord_schema/validations.py
```diff
@@ -326,6 +380,23 @@ def validate_addition_speed(message):
...
+            and message.HasField('amount')):
+        warnings.warn(
+            'CrudeComponents with derived amounts cannot have their'
+            ' mass or volume specified explicitly', ValidationError)
+    if ((not message.HasField('has_derived_amount')
+         or message.has_derived_amount is False)
```
skearnes, line 393: Prefer `not message.has_derived_amount` to comparing to `False`; you've already covered the unset case.

skearnes: If we use FrozenMessage throughout the validations we can clean a lot of this up...not in this PR though :)

ord_schema/validations.py
skearnes: Maybe let's add a TODO here? I think these kinds of "there should be a link" checks will be valuable just for hardening the submission pipeline. This might be better placed as part of `validate_dataset`.

ord_schema/validations.py
```diff
@@ -562,14 +637,21 @@ def validate_reaction_outcome(message):
...
             'No products or conversion are specified for reaction;'
-            ' at least one must be specified', ValidationError)
+            ' this is permissible only for multistep reactions',
+            ValidationWarning)
 
 
 def validate_reaction_product(message):
-    del message  # Unused.
+    # pylint: disable=too-many-boolean-expressions
+    if (message.compound.HasField('volume_includes_solutes')
```
skearnes, line 648: Someone will need to go through and update the validation docs; not in this PR :)

ord_schema/validations_test.py
```diff
@@ -104,6 +104,31 @@ def test_orcid_should_fail(self):
         with self.assertRaisesRegex(validations.ValidationError, 'Invalid'):
             self._run_validation(message)
 
+    def test_synthesized_compound(self):
+        message = reaction_pb2.CompoundPreparation(
+            type='SYNTHESIZED', reaction_id='dummy_reaction_id')
+        self.assertEmpty(self._run_validation(message))
+        message = reaction_pb2.CompoundPreparation(
+            type='NONE', reaction_id='dummy_reaction_id')
+        with self.assertRaisesRegex(validations.ValidationError, 'IDs'):
```
skearnes, line 113: Maybe be a bit more descriptive with this regex? I'm not sure what the error is from reading just this.

ord_schema/validations.py
connorcoley: I would advocate for a default of False, since a normal use case will be people checking their programmatic definitions for validity before submission. Having the default be True might encourage users to assign their own IDs so the validation passes, which I don't think we want to do

ord_schema/validations.py
connorcoley: I think using `HasField` explicitly is a nice reminder that we have to be careful with optional Booleans, but this makes sense. I'll simplify for now, even though we might revert if we switch to using FrozenMessages

ord_schema/validations.py
connorcoley: I made a separate `get_referenced_reaction_ids` that makes this a little cleaner

connorcoley:
FYI @antonkast-google this should not break the editor, but simply introduces additional fields that will not be captured by the current form

antonkast-google:
No problem. I will add the new fields in the editor.

PR #221, opened 2020-06-25
Updates to reaction HTML

skearnes:
* Clean up borders and group components by their input
* Stable sort order so the page looks the same every time
* Here's an example: [test-reaction.pdf](https://github.com/Open-Reaction-Database/ord-schema/files/4833320/test-reaction.pdf)

TODOs:
* Click on the compound name to copy the SMILES to the clipboard?
* Calculate and show the concentration for solutes?


ord_schema/visualization/template.html
```diff
@@ -14,101 +14,153 @@
...
+        {% for key, input in reaction.inputs|sort_addition_order %}
+            {% for compound, border in input.components|get_input_borders %}
+                <td class="{{ border }}">{{ compound|compound_role }}</td>
             {% endfor %}
         {% endfor %}
-        <td></td>
+    </tr>
+
+    {# LAST ROW = addition order groups #}
+    <tr><td class="clean"><br></td></tr>
```
connorcoley, line 160: Do we need to wrap this in an `if` so that it doesn't get displayed if the addition order is not defined?

skearnes: Done.

skearnes:
Changed to sans-serif :)

skearnes:
Added TODOs to the [ORD UI Requirements](https://docs.google.com/document/d/1kinvTzbyCM3YVUqZoSbhFKePGoYNPElhrGs7PILDPWo/edit#heading=h.cp0yh4eww8qr) doc.

ord_schema/units.py
```diff
@@ -60,8 +60,8 @@
         reaction_pb2.Pressure.KILOPASCAL: ['kPa', 'kilopascals', 'kPas'],
     },
     reaction_pb2.Temperature: {
-        reaction_pb2.Temperature.CELSIUS: ['C', 'degC', 'celsius'],
-        reaction_pb2.Temperature.FAHRENHEIT: ['F', 'degF', 'fahrenheit'],
+        reaction_pb2.Temperature.CELSIUS: ['°C', 'C', 'degC', 'celsius'],
```
antonkast-google, line 63: µL? µg?

ord_schema/visualization/drawing.py
```diff
@@ -80,7 +80,10 @@ def mol_to_png(mol, max_size=1000, bond_length=25, png_quality=70):
     molecule."""
```
antonkast-google, line 80: update the comment too?

ord_schema/visualization/filters.py
```diff
@@ -23,9 +24,52 @@ def _is_true(boolean):
...
+
 def _sort_addition_order(inputs):
-    for k in sorted(inputs.keys(), key=lambda k: inputs[k].addition_order):
-        yield k, inputs[k]
+    """Sort inputs by addition order, sorting again within stages/steps.
+
+    Args:
+        inputs: Map of ReactionInput messages.
+
+    Returns:
```
antonkast-google, line 49: "Yields:"?

ord_schema/visualization/filters.py
```diff
@@ -159,12 +200,10 @@ def _temperature_conditions_html(temperature):
...
     if (temperature.control.type == temperature.control.UNSPECIFIED
             or temperature.control.type == temperature.control.AMBIENT):
-        return 'ambient temperature<br>'
+        return 'ambient temperature'
     setpoint = units.format_message(temperature.setpoint)
     if setpoint:
         txt += f'{setpoint}'
-    if txt:
-        txt += '<br>'
     return txt
```
antonkast-google, line 209: In _analysis_format() below, we're still returning strings with angle brackets. Will those need escaping for HTML injection? Or is Jinja taking care of that?

connorcoley: Jinja [does not auto-escape by default](https://jinja.palletsprojects.com/en/2.11.x/templates/#html-escaping), but it can. The need to handle HTML injections is an excellent point. If possible, it would be better for these filters to never return HTML tags so we could

skearnes: Flask escapes by default; I removed HTML tags from the bits of the template that were touched by my example, but clearly I missed some!

ord_schema/units.py
skearnes: Done (in a new PR).

ord_schema/visualization/drawing.py
skearnes: Updating the docstrings for all visualization modules (in another PR).

ord_schema/visualization/filters.py
skearnes: Done (in another PR).

ord_schema/visualization/filters.py
skearnes: Fixed (in another PR).

PR #224, opened 2020-06-27
editor Dockerfile and instructions for use

antonkast-google:
See the README for simple instructions to bundle the editor into a Docker image.

### DIFF
--- a/editor/serve.sh
+++ b/editor/serve.sh
@@ -18,6 +18,4 @@ export PYTHONPATH=py:../build/lib
 export FLASK_APP=serve.py
 export FLASK_ENV=development
 
-python -m flask run
-# Caution.
-# python -m flask run --host=0.0.0.0 --port=80
+python -m flask run --host=0.0.0.0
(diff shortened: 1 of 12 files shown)


editor/serve.sh
```diff
@@ -18,6 +18,4 @@ export PYTHONPATH=py:../build/lib
 export FLASK_APP=serve.py
 export FLASK_ENV=development
 
-python -m flask run
-# Caution.
-# python -m flask run --host=0.0.0.0 --port=80
+python -m flask run --host=0.0.0.0
```
skearnes, line 21: Is the --host=0.0.0.0 required? Don't we only want this in the context of the docker container---maybe put this arg there instead?

From https://flask.palletsprojects.com/en/1.1.x/quickstart/
> If you have the debugger disabled or trust the users on your network, you can make the server publicly available simply by adding --host=0.0.0.0 to the command line

antonkast-google: Good thinking! Done.

PR #225, opened 2020-06-28
Add a user identifier for sharding.

antonkast-google:
The identifier is set in a cookie. It defines a directory, "db/{user}" where
the protos get written.

We use a cookie instead of a URL param so the identifier doesn't need to be
copied into all the links and XHRs (save, clone, upload, delete, compare).

However, we copy the identifier into URLs anyway when a page is viewed:
"?user={user}". Then if the URL identifier is different from the cookie
identifier, the URL wins and the cookie gets clobbered. This lets link sharing
continue to work.


editor/py/serve.py
```diff
@@ -375,11 +386,52 @@ def resolve_tokens(proto):
...
+
+def url_for_user(user):
+  parts = list(urllib.parse.urlparse(flask.request.url))
+  parts[4] = urllib.parse.urlencode({'user': user})
+  return urllib.parse.urlunparse(parts)
+
+
+def next_user():
+  """Return a user identifier that is not present in the db/ directory."""
+  user = ''.join(random.choice(string.ascii_letters) for i in range(4))
```
skearnes, line 434: Can we use a full `uuid.uuid4().hex` here? I want to reduce the risk of guessing a valid user ID and messing with their data.

antonkast-google: Done.

skearnes:
Thanks Anton!

PR #231, opened 2020-07-15
Add scripts for managing a persistent database image

skearnes:
Adds a Dockerfile and supporting scripts for creating and distributing a PostgreSQL database containing ORD data. There's no obviously simple way to distribute volumes, so I'm using the container data layer instead.

* Use update_image.sh to build and push an updated image to [Docker Hub](https://hub.docker.com/repository/docker/openreactiondatabase/ord-postgres).
* Use serve.sh to start the PostgreSQL server (exposed to the host on port 5432).

I've tested that the data persists in the container and that the rdkit extension works. Later PRs will add rdkit-specific functionality.

### DIFF
--- a/interface/docker/Dockerfile
+++ b/interface/docker/Dockerfile
@@ -1,0 +1,39 @@
+# Copyright 2020 Open Reaction Database Project Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM conda/miniconda3
+
+RUN apt-get update && apt-get upgrade -y && apt-get install -y \
+    git \
+    libpq-dev \
+    wget \
+    build-essential
+
+RUN conda update conda --yes && conda install -c rdkit rdkit-postgresql
+
+ENV ORD_ROOT=/opt
+WORKDIR "${ORD_ROOT}"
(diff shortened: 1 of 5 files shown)


interface/docker/postgres.conf
```diff
@@ -0,0 +1,750 @@
+# -----------------------------
```
antonkast-google, line 1: This looks like it came from an upstream distribution. Is it verbatim from upstream?

If not and you have changed some things, it might be nice to commit the upstream one first so your changes are in the history and an update would not be practical.

interface/docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN git clone https://github.com/Open-Reaction-Database/ord-schema.git
+RUN git clone https://github.com/Open-Reaction-Database/ord-submissions-test.git
+WORKDIR "${ORD_ROOT}/ord-schema"
+RUN pip install -r requirements.txt
+RUN python setup.py install
+
+# NOTE(kearnes): Use non-standard PGDATA so the database persists; see
+# https://nickjanetakis.com/blog/docker-tip-79-saving-a-postgres-database-in-a-docker-image.
+WORKDIR "${ORD_ROOT}"
+ENV PGDATA=/data
```
connorcoley, line 36: This relies on the existence of a `/data` folder. Modifications to the database are not preserved after stopping the container and restarting it when I try on OS X. That shouldn't matter too much in our case, I guess

skearnes: Modifications are only saved if you `docker commit` like I do in `update_image.sh`. The `/data` folder is stored in the container's storage layer while the container is running.

skearnes:
> The conf file looks like it came from an upstream distribution. Is it verbatim from upstream?
> 
> If not and you have changed some things, it might be nice to commit the upstream one first so your changes are in the history and an update would be practical.

It's an edited version of the sample config provided by the postgres image; I'll check in the vanilla config here and move the changes to #232.

interface/docker/Dockerfile
skearnes: (I have tested this on OS X.)

PR #232, opened 2020-07-16
Add RDKit cartridge functionality

skearnes:
Adds RDKit cartridge functionality to the ORD postgres backend.

* Edits to postgres.conf are recommended [here](https://www.rdkit.org/docs/Cartridge.html#configuration).

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -347,7 +347,9 @@ good-names=i,
            ex,
            Run,
            _,
-           f
+           f,
+           db,
+           df
 
 # Include a hint for the correct naming format with invalid-name.
 include-naming-hint=no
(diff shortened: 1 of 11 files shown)


skearnes:
@greglandrum PTAL if you have a chance; are we doing anything crazy here?

greglandrum:
I don't see anything in the way the rdkit is being used that sets off any alarm bells.
Since it looks like you're heading towards doing structure searches on reactions, it's worth keeping in mind that substructure searches on reactions don't always behave exactly the way that people might expect.

Of course, it's been a while since I worked with that code so I don't remember exactly what the searches do. I would need to dig back into that.

skearnes:
> I don't see anything in the way the rdkit is being used that sets off any alarm bells.
> Since it looks like you're heading towards doing structure searches on reactions, it's worth keeping in mind that substructure searches on reactions don't always behave exactly the way that people might expect.
> 
> Of course, it's been a while since I worked with that code so I don't remember exactly what the searches do. I would need to dig back into that.

Thanks! We'll focus on the vanilla substructure searching in the inputs/outputs tables first; let's chat sometime about the subtleties around reaction matching.

PR #234, opened 2020-07-17
Update update_image.sh

skearnes:


interface/docker/update_image.sh
```diff
@@ -15,8 +15,13 @@
 
 set -ex
 
-docker build -t ord-postgres:empty .
-docker run --rm --name ord-postgres -d -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=postgres ord-postgres:empty
+# NOTE(kearnes): Disable caching so the latest version of the ORD is used.
```
antonkast-google, line 18: Could a makefile accomplish the same effect but with the benefit of caching?

skearnes: Possibly, but the cost of cloning the latest repos from github is about the same as the cost of all the other RUN steps, so I don't think it's worth investing much effort into improving this.

skearnes: There are additional mechanisms we could use to break caching just for the clone steps, but again I'm not too concerned about this.

PR #237, opened 2020-07-21
Add support for stereochemistry in substructure searches

skearnes:
There's no stereochemistry in the current docker image to test on; let's chat about building a set of test data for the interface backend.

(Removing the `Cursor` class since cursor() can already be used as a context manager; https://www.psycopg.org/docs/usage.html#with-statement)

### DIFF
--- a/ord_schema/interface/query.py
+++ b/ord_schema/interface/query.py
@@ -69,11 +69,6 @@ class OrdPostgres:
                             use_stereochemistry=False):
         """Performs a substructure search.
 
-        Note that matches will ignore stereochemistry by default unless
-        rdkit.do_chiral_sss is set.
-
-        TODO(kearnes): Set stereochemistry flag on a per-query basis.
-
         Args:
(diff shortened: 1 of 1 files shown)


connorcoley:
It would be good to have some way of setting that flag only for the current query. Maybe there is an undocumented option for doing that

skearnes:
> It would be good to have some way of setting that flag only for the current query. Maybe there is an undocumented option for doing that

Note that this resets the flag for each query, which should accomplish the same thing.

connorcoley:
> Note that this resets the flag for each query, which should accomplish the same thing.

You're forgetting about when this database is wildly popular and we have to worry about race conditions!

skearnes:
> > Note that this resets the flag for each query, which should accomplish the same thing.
> 
> You're forgetting about when this database is wildly popular and we have to worry about race conditions!

I think it's still safe; each `with` on the connection is a separate transaction. See #238 for an improved option.

PR #244, opened 2020-07-23
Add functions for getting Mol and SMILES from Compound messages

skearnes:


ord_schema/message_helpers.py
```diff
@@ -253,36 +253,89 @@ def find_submessages(message, submessage_type):
...
+            if return_identifier:
+                return mol, identifier
+            return mol
+
+
+# pylint: enable=inconsistent-return-statements
+
+
+def check_compound_identifiers(compound):
+    """Verifies that structural compound identifiers are consistent.
```
connorcoley, line 316: We might have issues with tautomers here. Roundtripping SMILES -> InChI -> SMILES can change the string. So if someone is programmatically defining a compound and uses both a `MolToSmiles` and `MolToInchi`, this check might fail. This might be worth a quick email to Greg to ask for advice on checking consistency (or not)

ord_schema/updates.py
```diff
@@ -97,22 +96,17 @@ def add_binary_identifiers(message):
...
-                                         type='RDKIT_BINARY',
-                                         details=f'Generated from {source}')
-                modified = True
-                break  # Only add one RDKIT_BINARY per Compound.
+        try:
+            mol, identifier = message_helpers.mol_from_compound(
+                compound, return_identifier=True)
+            source = reaction_pb2.CompoundIdentifier.IdentifierType.Name(
+                identifier.type)
+            compound.identifiers.add(bytes_value=mol.ToBinary(),
```
connorcoley, line 104: Is it intentional that we are still excluding `add_binary_identifiers` from the default `_UPDATES`?

ord_schema/visualization/generate_text_test.py
```diff
@@ -37,8 +37,8 @@ def setUp(self):
             ))
         reaction.inputs['dummy_input'].components.add().CopyFrom(
             message_helpers.build_compound(
-                name='C1OCCC1',
-                smiles='THF',
+                name='THF',
```
connorcoley, line 40: oops

ord_schema/message_helpers.py
skearnes: Agreed; we don't necessarily have to use this anywhere until it's doing what we want---would you prefer that I remove it for now?

ord_schema/updates.py
skearnes: Yes, we didn't want to store those in the official repo. And we don't need them for Postgres since we are already computing columns that contain Mol objects.

We might even consider removing RDKIT_BINARY as a type and dropping the bytes_value field...

ord_schema/visualization/generate_text_test.py
skearnes: Yeah, I'm not quite sure why the test was passing before.

ord_schema/updates.py
connorcoley: I was considering suggesting removing RDKIT_BINARY for both CompoundIdentifiers and ReactionIdentifiers. It would be nice to remove the bytes_value field from these, too. Would we still allow people to upload files on the editor but just read them as a string and insert them into the value field?

ord_schema/message_helpers.py
connorcoley: Where were you planning to use it? I think it would be appropriate to use it when validating a compound to issue a warning, but not an error.

skearnes: I was thinking of putting it into the validation for Compound; we could issue a warning in that case.

ord_schema/updates.py
skearnes: Yes, that would make sense to me.

ord_schema/message_helpers.py
skearnes: Added a commit to do this.

PR #248, opened 2020-07-24
Add RDFILE, remove ATOM_MAPPED_SMILES, add is_mapped

connorcoley:


ord_schema/proto/reaction.proto
```diff
@@ -76,6 +76,10 @@ message ReactionIdentifier {
     string value = 3;
     bytes bytes_value = 4;
   }
+  // Whether identifier contains atom-to-atom mapping information. When True,
+  // we encourage users to specify how that mapping was obtained in the
+  // details field (e.g., manually, using NameRXN, using ChemDraw).
+  optional bool is_mapped = 5;
```
skearnes, line 82: Do we also want a "details" field here for the mapping algo?

skearnes: Oh sorry, there's one already.

connorcoley: I don't think there necessarily needs to be a separate `mapping_details` field. There will rarely be any details to provide about a `ReactionIdentifier`, so I imagine `details` will be sufficient. However, this would provide more structure...might as well add it. I'm going with `how_mapped` to be consistent with `how_computed` for compound features.

skearnes: LGTM; your choice if you prefer adding it vs. not.

PR #250, opened 2020-07-24
Add get_reaction_smiles

skearnes:
Fairly simple reaction SMILES generator.

### DIFF
--- a/ord_schema/message_helpers.py
+++ b/ord_schema/message_helpers.py
@@ -342,12 +342,16 @@ def check_compound_identifiers(compound):
         raise ValueError(f'structural identifiers are inconsistent: {smiles}')
 
 
-def get_reaction_smiles(message, validate=True):
+def get_reaction_smiles(message, allow_incomplete=True, validate=True):
     """Fetches or generates a reaction SMILES.
 
     Args:
         message: reaction_pb2.Reaction message.
+        allow_incomplete: Boolean whether to allow "incomplete" reaction SMILES
+            that do not include all components (e.g. if a component does not
+            have a structural identifier).
         validate: Boolean whether to validate the reaction SMILES with rdkit.
+            Only used if allow_incomplete is False.
 
     Returns:
         Text reaction SMILES.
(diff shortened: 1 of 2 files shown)


ord_schema/message_helpers.py
```diff
@@ -338,6 +342,72 @@ def check_compound_identifiers(compound):
...
+    reactants, agents, products = set(), set(), set()
+    roles = reaction_pb2.Compound.ReactionRole()
+    for key in sorted(message.inputs):
+        for compound in message.inputs[key].components:
+            smiles = smiles_from_compound(compound)
+            if compound.reaction_role in [
+                    roles.REAGENT, roles.SOLVENT, roles.CATALYST
+            ]:
+                agents.add(smiles)
+            else:
```
connorcoley, line 370: This should have an `elif compound.reaction_role == roles.INTERNAL_STANDARD: continue` too

ord_schema/message_helpers.py
```diff
@@ -338,6 +342,72 @@ def check_compound_identifiers(compound):
...
+                reactants.add(smiles)
+    for outcome in message.outcomes:
+        for product in outcome.products:
+            smiles = smiles_from_compound(product.compound)
+            products.add(smiles)
+    if not reactants or not products:
+        raise ValueError(
```
connorcoley, line 377: How will this function be used / when will this ValueError be caught? With representing certain multi-step reactions w/o intermediate isolation, we'll have `Reaction` messages without any products. If the only input is a crude component, theoretically a reaction could also have no well-defined reactants.

ord_schema/message_helpers.py
```diff
@@ -338,6 +342,72 @@ def check_compound_identifiers(compound):
...
+    Returns:
+        Updated reaction SMILES.
+
+    Raises:
+        ValueError: If the reaction contains errors.
+    """
+    try:
+        reaction = rdChemReactions.ReactionFromSmarts(reaction_smiles,
+                                                      useSmiles=True)
+        rdChemReactions.SanitizeRxn(reaction)
```
connorcoley, line 401: I'm not too familiar with what this method does; any chance of unintended consequences?

ord_schema/message_helpers.py
```diff
@@ -338,6 +342,72 @@ def check_compound_identifiers(compound):
...
+        ValueError: If the reaction contains errors.
+    """
+    for identifier in message.identifiers:
+        if identifier.type == reaction_pb2.ReactionIdentifier.REACTION_SMILES:
+            return identifier.value
+    reactants, agents, products = set(), set(), set()
+    roles = reaction_pb2.Compound.ReactionRole()
+    for key in sorted(message.inputs):
+        for compound in message.inputs[key].components:
+            smiles = smiles_from_compound(compound)
```
connorcoley, line 365: I'm interpreting this as follows: a ValueError will be raised if _any_ compound cannot be resolved to a SMILES string. Suppose a reaction takes place in "hexanes" (mixture of isomers) as a solvent. The reactants, reagents, etc. can be well-defined but the solvent might be defined by name only. Do we want to still raise an error? What if a catalyst is "Pd/C" (super common) or a biological enzyme so there is no SMILES string and it is only defined by name?

ord_schema/message_helpers.py
```diff
@@ -338,6 +342,72 @@ def check_compound_identifiers(compound):
...
+            else:
+                reactants.add(smiles)
+    for outcome in message.outcomes:
+        for product in outcome.products:
+            smiles = smiles_from_compound(product.compound)
+            products.add(smiles)
+    if not reactants or not products:
+        raise ValueError(
+            'reaction must contain at least one reactant and one product')
+    reaction_smiles = '%s>%s>%s' % ('.'.join(reactants), '.'.join(agents),
```
connorcoley, line 379: For a minor type of "canonicalization", is it worth sorting these?

ord_schema/message_helpers.py
connorcoley: If I'm correct in that `smiles_from_compound` will raise an error, perhaps we should have some kwarg that is akin to `reactants_only` or `include_agents`? Alternatively, we could have some wrapper that can issue warnings instead of errors

skearnes: My plan was to catch the error as part of build_database.py and just not generate a reaction SMILES in those cases. I think this is ok since it's just for postgres---users could still search by the inputs/outputs of a reaction without searching based on the entire thing. I'm not sure what to do about `NAME`d components yet other than adding more columns for these that people can search against.

That being said, if you'd prefer to generate "incomplete" reaction SMILES that would be fine with me...what do you think? (This affects several other comments as well.)

connorcoley: I do think it would be worth allowing incomplete reaction SMILES in the database

ord_schema/message_helpers.py
skearnes: Possibly, but it's now not used by default since allow_incomplete defaults to True.

ord_schema/message_helpers.py
skearnes: Done. PTAL.

ord_schema/message_helpers.py
skearnes: Done.

ord_schema/message_helpers.py
skearnes: Done.

ord_schema/message_helpers.py
skearnes: I expect that any errors (which will be uncommon now with allow_incomplete) would be caught in build_database.py and simply skipped.

PR #258, opened 2020-08-03
Add change_stats.py for GitHub actions

skearnes:
Calculates changes in reaction IDs and publishes comments to PRs. See https://github.com/skearnes/ord-submissions-test/pull/1.


connorcoley:
The summary is very useful here; the `git diff` splits the file in an awkward place

connorcoley:
Could `git diff` every cause an identical line to be added/removed if it is part of a larger segment of code that is moved? For example, what if you just switched the order of two reactions in a dataset?

I wonder if these stats should be calculated from the two parsed Dataset messages, rather than relying on `git diff`.

skearnes:
> I wonder if these stats should be calculated from the two parsed Dataset messages, rather than relying on `git diff`.

You're probably right about that; what we really want is the set difference between Datasets A and B and that's a cleaner approach. I'll do that.

Issue #247, opened 2020-07-24
Remove RDKIT_BINARY from identifier options


connorcoley:
We decided not to store serialized molecules in the ground truth data, making this an unnecessary way of defining a compound or reaction.

- Remove RDKIT_BINARY as an identifier type
- Remove bytes_value field
- Change editor (webform) to populate the string value field upon file upload of, e.g., a MolFile

PR #260, opened 2020-08-03
Simplify identifier upload options

connorcoley:
- Does not allow users to upload their own binary data, but allows
  for any existing binary data to be retrieved still
- Upload button by text box inserts file contents into the value
  box directly

### DIFF
--- a/editor/js/reaction.js
+++ b/editor/js/reaction.js
@@ -447,6 +447,22 @@ function writeMetric(prefix, proto, node) {
   }
 }
 
+// Prompt user to upload a file and return its content as a string.
+function setTextFromFile(identifierNode, valueClass) {
+  var input = document.createElement('input');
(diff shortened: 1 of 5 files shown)


editor/js/reaction.js
```diff
@@ -462,6 +478,10 @@ function initSelector(node) {
     if (options[i] == 'UNSPECIFIED') {
       option.attr('selected', 'selected');
     }
+    // Don't let users directly select/enter RDKIT_BINARY identifiers.
+    if (options[i] == 'RDKIT_BINARY') {
```
skearnes, line 482: I assume this will go away after #247 is fixed?

connorcoley: Yes; #247 is slightly more involved, but this line would be removed

PR #261, opened 2020-08-03
Name resolve backups

connorcoley:
PubChem's name resolver API is currently down! 

```
Status: 500
Code: PUGREST.ServerError
Message: Can't see oe_license.txt
```

Highlighted the need for a backup option. Added NIH's cactus resolver. At least our tests will pass now.

### DIFF
--- a/ord_schema/updates_test.py
+++ b/ord_schema/updates_test.py
@@ -77,7 +77,8 @@ class UpdateReactionTest(absltest.TestCase):
         self.assertEqual(
             component.identifiers[1],
             reaction_pb2.CompoundIdentifier(
-                type='SMILES', value='CCN',
+                type='SMILES',
+                value='CCN',
                 details='NAME resolved automatically'))
 
     def test_add_reaction_id(self):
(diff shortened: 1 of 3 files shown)


ord_schema/updates_test.py
```diff
@@ -23,16 +23,20 @@
...
             type='NAME', value='aspirin')
         self.assertTrue(updates.resolve_names(message))
+        resolved_smi = roundtrip_smi(
+            message.inputs['test'].components[0].identifiers[1].value)
+        self.assertEqual(resolved_smi, roundtrip_smi('CC(=O)Oc1ccccc1C(O)=O'))
         self.assertEqual(
-            message.inputs['test'].components[0].identifiers[1],
-            reaction_pb2.CompoundIdentifier(
-                type='SMILES',
-                value='CC(=O)OC1=CC=CC=C1C(=O)O',
```
skearnes, line 34: Since these can differ, is it important to have the details specify which service was used?

connorcoley: I was wondering if you thought that would be worthwhile; I think it's less important to note _how_ it was automatically resolved (assuming it was a somewhat reputable service). The intent is for the user to manually check whether the structure makes sense when using the webform

PR #262, opened 2020-08-03
Dataset-level change statistics

skearnes:
Replaces #258 

Adds new features to `process_dataset.py`:

* The sets of added and removed reaction IDs are computed by comparing to the branch specified by `--base`. Specifically, we load the `Dataset` messages from the base branch and compare to the `Dataset` messages loaded from the current HEAD.
* Tests now use a new branch instead of just a new commit to better simulate real-world use.


skearnes:
Refactored a bit for pylint and added a new dep on `pygithub`; PTAL.

PR #264, opened 2020-08-04
Remove rdkit binary

connorcoley:
Addresses #247 

- No more bytes_value fields in proto for identifiers
- Validation updated to require that "string value" be defined
- Tests, helper functions updated
- Update function to assign binary now irrelevant
- Editor changed to remove upload boxes around compound and
  reaction identifiers
- load/unload of identifiers in webform simplified
- `full.pbtxt` dataset example updated to not use bytes_value in reaction identifier


skearnes:
Please update the description to link to the issue.

PR #265, opened 2020-08-04
Use Ketcher molecular drawer in editor

n8kim1:
Adds Ketcher, an open-source molecular drawer/editor, as a method of drawing compound identifiers. (More info about Ketcher: [their website](https://lifescience.opensource.epam.com/ketcher/index.html), and [their repo](https://github.com/epam/ketcher).)

Summary:
- When adding identifiers to a component, there's a button to "draw compound", which pulls up Ketcher in a modal
    - If we can derive the component's molecular structure (as a Molblock), then do so, and load this structure in Ketcher
        - Implemented methods to convert some identifiers to Molblock
    - While waiting for this loading process to happen, show a spinner
- When done drawing, the SMILES string and Molblock representation of what's drawn are added as identifiers. 
    - (Overwrites previous SMILES/Molblocks, if present)
- Implemented adding Ketcher to the Makefile:
    - Clone the ORD/ketcher repo if not present; pull to ensure we're up-to-date
    - Build Ketcher as necessary on changes to source code

We also created [our own fork of Ketcher](https://github.com/Open-Reaction-Database/ketcher), to make some of our own changes:
- Ketcher originally loaded through window.onload; now we load it with a custom, more controllable method 
- Bump some npm requirements to avoid recently-found vulnerabilities
- Skip a check that only allowed Ketcher's organization to commit


n8kim1:
@connorcoley please add/revise the description as you see fit!

editor/js/compounds.js
```diff
@@ -288,7 +288,76 @@ ord.compounds.addNameIdentifier = function (node) {
...
+      }
+    });
+    // Create new identifiers.
+    const identifier = new proto.ord.CompoundIdentifier();
+    identifier.setType(proto.ord.CompoundIdentifier.IdentifierType.SMILES);
+    identifier.setValue(ketcher.getSmiles());
+    identifier.setDetails('Drawn with Ketcher');
+    ord.compounds.loadIdentifier(node, identifier);
+    identifier.setType(proto.ord.CompoundIdentifier.IdentifierType.MOLBLOCK);
+    identifier.setValue(ketcher.getMolfile());
```
skearnes, line 358: Should there be a `setDetails` call here as well?

editor/Makefile
```diff
@@ -19,7 +19,7 @@ endif
 # https://github.com/google/closure-library/releases/
 closure = closure-library-20200517
 
-all : prep py js
+all : prep py js ketcher/dist/.touch
```
antonkast-google, line 22: The editor already has third-party dependencies: protobufs, closure, and the ORD proto classes. These are not set up with "make". Instructions are supplied in the README instead.

I think that way makes sense for such complex initializations where file modification times are not the best way to ensure consistency.

editor/README.md
```diff
@@ -19,15 +19,16 @@ Everything requires Python 3.
...
+* the Closure Library for Javascript; and
+* Node.js and npm.
 
 Serving depends on:
 * the Flask python web framework.
 
 To build the ord-schema python code, follow the instructions [here](https://github.com/Open-Reaction-Database/ord-schema/blob/main/README.md).
 
-The editor currently requires HEAD protobuf (to get experimental "optional" declarations) for the protobuf compiler and Javascript runtime libraries.
+For the protobuf compiler and Javascript runtime libraries, the editor currently requires HEAD protobuf (to get experimental "optional" declarations).
```
antonkast-google, line 31: Thanks for cleaning this up!

No need to fix this now, but this file a good example of a change that is unrelated to the rest of the PR and would have been good to send as a separate one.

skearnes, line 31: Note that this will go away after #266 is done.

editor/css/ketcher-extras.css
```diff
@@ -0,0 +1,344 @@
...
+    height: -webkit-min-content;
+    height: -moz-min-content;
+    height: min-content;
+  }
+  .modal-sm {
+    max-width: 300px;
+  }
+}
+
+@media (min-width: 992px) {
```
antonkast-google, line 243: As far as I know, none of our UI has been optimized for mobile yet.

editor/html/ketcher_iframe.html
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
```
antonkast-google, line 1: This HTML does not look well-formed.

editor/css/ketcher-extras.css
```diff
@@ -0,0 +1,344 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Styling for properly displaying Ketcher in the reaction page. 
+ * Mainly derived from Bootstrap v4.5.0 (https://getbootstrap.com/).
+ */
```
antonkast-google, line 20: Can you say something about why you're using Bootstrap here?

Is there something tricky about displaying Ketcher? Could there be a simpler way?

editor/css/ketcher-extras.css
n8kim1: Yes -- I just directly copy-pasted Bootstrap code for consistency. Would it be better to remove these?

editor/css/ketcher-extras.css
n8kim1: We decided on displaying Ketcher through a modal. Showing an iframe in the middle of the page, hiding the background, etc. is doable with our own CSS, but we figured it'd be okay to use what Bootstrap has.

(Less importantly, using Bootstrap also allowed us to use their spinners too.)

editor/html/ketcher_iframe.html
n8kim1: Whoops, will move the script tags into body. I can add a head too.

Outside of this, would you suggest any other changes? Because this is only meant to show Ketcher, I'd prefer to keep it smaller, if possible.

editor/README.md
n8kim1: Yup! My bad.

editor/Makefile
n8kim1: While this makes sense to me, I just have one concern: if we make changes that require the newest version of the Ketcher repo, what's the best way to ensure that users pull and use this newest version? Should we add some versioning/version checking? 

(At the moment, the makefile pulls every time it's run, and only rebuilds when necessary.)

editor/js/compounds.js
connorcoley: We're re-using the same `identifier` so the details are already set; we just overwrite the type and value.

editor/css/ketcher-extras.css
antonkast-google: Generally if we need a big block of mysterious code that belongs to someone else, we handle it as a third-party dependency rather than copying from their repo into ours.

Maybe we can discuss in the meeting today the best way to handle this.

n8kim1: I see. I'd like to include it as a dependency, I don't know how best to do so though... Available versions of the Bootstrap CSS library consist of a ton of stylings which override ours. I decided to pick out only some parts.

(I can wait til the meeting to discuss too; just didn't want to forget this.)

editor/Makefile
antonkast-google: If a dependency is something we control, then we can have an automated test that will block its changes if they break something downstream (which is common). We have no way to do that for third parties though, so usually it's better to have a manual step in the process.

editor/css/ketcher-extras.css
antonkast-google: I'm sorry, misunderstanding: I thought you had written this yourself and I was trying to save you some trouble.

Since this is third party code, any changes we make will require merging when we update it. We need to decide if this is a dependency or if we're forking it.

editor/html/ketcher_iframe.html
n8kim1: (BTW, pushed changes. It's having trouble showing up for me; just in case, [here's a link](https://github.com/Open-Reaction-Database/ord-schema/blob/editor-ketcher/editor/html/ketcher_iframe.html).)

antonkast-google: I have no other suggestions, but I can't say I understand why an iFrame is helpful.

The only time I expect to see an iFrame is when someone needs a separate JS context for code from a different origin. Since we are hosting the JS, I would need to know the purpose.

editor/css/ketcher-extras.css
antonkast-google: I see. This could be fine. Maybe in the meeting you can just explain the decision.

editor/html/ketcher_iframe.html
n8kim1: iFrames are EPAM's (the organization behind Ketcher) recommended way to add Ketcher to another page. I tried to use Ketcher in a regular div, but ran into difficulties that were tough to fix and required our own changes.

We wanted to avoid our own changes as much as possible, and also to save time -- we figured it would be better to stick with an iFrame.

editor/Makefile
n8kim1: Ah I see. We're using our own fork of Ketcher, so maybe we do have this control. Automated tests would take a while though (maybe for later?); I can make this a manual step for now and put instructions in the readme.

editor/css/ketcher-extras.css
n8kim1: TODO (from mtng): add the link to where we derived this css from; add a link to where Bootstrap says it's ok to use parts. Also might have to update our licenses, noting that we're using some stuff under MIT license.

editor/Makefile
n8kim1: TODO (from mtng): tag our own version of Ketcher; in make, check that we're using this tag, and fail fast if we're not.

editor/css/ketcher-extras.css
n8kim1: @skearnes added some more info to the css file. Are there any changes we need to make to our own readmes? (Sorry for the potential headache)

editor/css/ketcher-extras.css
skearnes: I think you need to create a separate file for the boostrap-only things, and put their license header at the top:

https://github.com/twbs/bootstrap/blob/89dc9750ec029a1479cba87f27ab936df6edb681/js/src/alert.js#L1-L6

Then put your own CSS in another file with our usual ORD header.

n8kim1: Sounds good, thanks. Should I also remove the comments I wrote? (these: 

image

skearnes: I think it's fine to put a link to the source file, but I would leave out the other comments. Make sure you clearly indicate any changes you make to their CSS. If it's just copy-paste of select elements (and no changes), you should indicate that in the comments below the header as well.

skearnes: I'd use the header from that source file directly:

/*!
 * Bootstrap v4.5.0 (https://getbootstrap.com/)
 * Copyright 2011-2020 The Bootstrap Authors
 * Copyright 2011-2020 Twitter, Inc.
 * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
 */

n8kim1: How does this look?

image

skearnes:
Can you use the original header from that file?

editor/css/ketcher-extras.css
skearnes: Can you use the header from the original file?

n8kim1: @skearnes just saw your comment about the original header (not sure why I didn't earlier, sorry). Switched to that; anything else? 

image

skearnes: LGTM; thanks.

editor/Makefile
n8kim1: ^ in progress, reopening

n8kim1: Finished -- added this check, added another check requiring Ketcher to be fully built, and added the installation process to GitHub CI.

@antonkast-google @skearnes would you guys be able to double-check the Makefile and CI? Will rebase and merge once everything's good!

skearnes: LGTM for the updates to `run_tests.yml`. It would be nice to have a one-shot script to install these dependencies at some point, but that's not blocking for this PR.

editor/Makefile
```diff
@@ -67,6 +74,12 @@ gen/js/proto/ord/%.js : ../ord_schema/proto/%.proto
 ../build/lib/ord_schema/proto/%_pb2.py : ../ord_schema/proto/%.proto
 	$(error $< not found--run setup?)
 
+# Checks for the presense of the last file produced in Ketcher's build. 
+# (hacky; may want to revisit later)
+ketcher/dist/doc/help.html : 
+	# Ensure that ketcher has been built.
+	$(error built ketcher not found, see editor readme?)
```
antonkast-google, line 81: I think you can omit this target. The tag check is sufficient.

editor/Makefile
```diff
@@ -90,4 +106,6 @@ test : all
 
 clean :
 	rm -rf gen
+	rm -rf ketcher/node_modules
+	rm -rf ketcher/dist
```
antonkast-google, line 110: If ketcher/dist was not created by make, then I would not delete it with "make clean".

editor/Makefile
n8kim1: I was worried about the Ketcher build failing, but users still trying to proceed anyways. Is this worth checking for?

antonkast-google: But the ketcher build is not part of this build.

I wouldn't use make to ensure that a human did a job.

n8kim1: Makes sense, sounds good.

Thanks for the checks!

PR #271, opened 2020-08-05
Stricter provenance validation

skearnes:
* A `Person` message in a `RecordEvent` must pass the following new validations (note that this does not apply to `ReactionProvenance.experimenter`):
  * At least one of `username`, `name`, or `orcid` must be set.
  * `email` must be set.
* The submission workflow scripts no longer create a missing `record_created` event. This must be created by the user.
* `record_modified` events added automatically during submission use the GitHub Actions username and email.


skearnes:
@connorcoley PTAL before I go through and update all the notebooks...

skearnes:
> L(mostly)GTM. One question is that if we have a record modification added automatically in `updates.py`, do we encourage users to have added _their own_ in advance? If not, there is no way to specify their name/ORCID
> 
> Perhaps we need to do the following:
> 
> * Check if the reaction has been modified
> * If yes, check if the current record has at least one more `reaction.provenance.record_modified` item than the previous version of the record
> * If yes, add the GitHub username is `username` is empty; GitHub email if `email` is empty to the most recent `record_modified` message
> * If no, add a new event (as the current code does)

Here's what happens now (in this PR):

* A `record_created` event must be provided by the user.
* `record_modified` events added by `updates.py` have the GitHub Actions bot username and email, since these modifications were not made by the user.
* I guess we're not expecting users to add `record_modified` events for manual changes they make during the review process, but they are free to do this if they want.

Does that answer your question/resolve your concerns?

connorcoley:
Right, the automatic modification record is only when there are changes automatically made by the update script. I had somehow forgotten the context for the `if modified`. 

My question stands, but is not related to this PR then: when validating submissions, if we detect that a user has modified a Reaction but they haven't added an event_modified RecordEvent, we should raise an error.

PR #276, opened 2020-08-06
Move DatasetExample out of Dataset

skearnes:
It's awkward to edit a `Dataset` to add a `DatasetExample`; instead we should just have a `dataset_id` pointer in `DatasetExample` and take it out of `Dataset`.


skearnes:
PTAL; removed DatasetExample from editor.

PR #278, opened 2020-08-07
Use best practices for SQL queries

skearnes:
From the [psycopg2 documenation](https://www.psycopg.org/docs/usage.html#the-problem-with-the-query-parameters):

> Never, __never__, __NEVER__ use Python string concatenation (+) or string parameters interpolation (%) to pass variables to a SQL query string. Not even at gunpoint.

* Updates queries to use `psycopg2.sql` classes. The changes to `build_database.py` are probably not strictly necessary since that script doesn't take any untrusted inputs.
* Updates tests to clean up after themselves and provide better logging.

### DIFF
--- a/ord_schema/interface/build_database.py
+++ b/ord_schema/interface/build_database.py
@@ -26,6 +26,7 @@ from absl import app
 from absl import flags
 from absl import logging
 import psycopg2
+from psycopg2 import sql
 
 from ord_schema import message_helpers
 from ord_schema.proto import dataset_pb2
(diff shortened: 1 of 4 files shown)


ord_schema/interface/query.py
```diff
@@ -72,19 +75,29 @@ def substructure_search(self,
...
-            INNER JOIN {table} B ON A.reaction_id = B.reaction_id
-            WHERE B.{column}@>'{pattern}'{'::qmol' if use_smarts else ''}
-            {f'LIMIT {limit}' if limit else ''};"""
-        logging.info(query)
+        components = [
+            sql.SQL("""
+                SELECT DISTINCT A.reaction_id, A.serialized 
+                FROM reactions A 
+                INNER JOIN {} B ON A.reaction_id = B.reaction_id
+                WHERE B.{}@>%s""").format(sql.Identifier(*table.split('.')),
```
connorcoley, line 83: Why does `table` need to be split like this?

skearnes: In the tests, `table` comes in as something like `rdk.inputs` and using sql.Identifier(table) fails since "rdk.inputs" isn't a valid identifier but "rdk"."inputs" is --- see https://www.psycopg.org/docs/sql.html#psycopg2.sql.Identifier

ord_schema/interface/build_database.py
```diff
@@ -170,20 +171,35 @@ def create_database():
...
     if FLAGS.overwrite:
+        logging.info('Removing existing tables')
         for table in _TABLES:
-            cursor.execute(f'DROP TABLE IF EXISTS {table};')
-    cursor.execute('CREATE EXTENSION IF NOT EXISTS rdkit;')
-    cursor.execute('CREATE SCHEMA rdk;')
+            command = sql.SQL('DROP TABLE IF EXISTS {}')
+            cursor.execute(command.format(sql.Identifier(table)))
+    cursor.execute(sql.SQL('CREATE EXTENSION IF NOT EXISTS rdkit'))
+    cursor.execute(sql.SQL('CREATE SCHEMA rdk'))
```
connorcoley, line 179: Minor point, but perhaps the schema name rdk should be defined by an internal constant `_SCHEMA` in this file and not hard-coded here and in subsequent queries, e.g., ` 'rdk.{table}'`

connorcoley:
It looks like all unsanitized calls were replaced properly. A stylistic comment is that these queries would be more clear if you stick to using f-strings rather than `.format`

skearnes:
> It looks like all unsanitized calls were replaced properly. A stylistic comment is that these queries would be more clear if you stick to using f-strings rather than `.format`

Note that the `.format` call is happening on the `SQL` object and not the string---I was confused by this at first as well. It means we can't use f-strings...

ord_schema/interface/build_database.py
skearnes: Good idea; I'll push a fix.

connorcoley:
> > It looks like all unsanitized calls were replaced properly. A stylistic comment is that these queries would be more clear if you stick to using f-strings rather than `.format`
> 
> Note that the `.format` call is happening on the `SQL` object and not the string---I was confused by this at first as well. It means we can't use f-strings...

So it is! I noticed that and immediately forgot it before writing my comment.

ord_schema/interface/query.py
skearnes: Added a comment.

PR #279, opened 2020-08-07
Render compounds upon validation

connorcoley:
When compounds are validated, try to retrieve an RDKit molecule and then render a png depiction in the editor. For example, 

image


connorcoley:
@skearnes having trouble with long tests it seems; just re-triggered to see if that fixes it

skearnes:
> @skearnes having trouble with long tests it seems; just re-triggered to see if that fixes it

Sometimes pushing an empty commit resets things; idk why but you might try that.

skearnes:
Can you try pushing a commit that sets `verbosity=0` in `run_long_tests.sh`?

PR #283, opened 2020-08-07
Reaction-level visualization upon reaction validation

### LINKED ISSUE #269: Show visual summary of reaction on editor page

connorcoley:
Closes #269

- Displays a visual summary of the reaction at the top of the editor using `ord_schema.visualization.generate_text`
    - Note that flag definitions in this file had to be moved so multiple imports didn't lead to conflicting absl flag definitions
- Reaction-level validation is triggered after loading the form

Example at `http://localhost:5000/dataset/ord-nielsen-example/reaction/0?user=test` (I know this isn't a real link, but would make it faster for someone to test on their local deployment) using the new Nielsen example:

image


ord_schema/visualization/generate_text.py
```diff
@@ -101,5 +95,12 @@ def main(argv):
 
 
 if __name__ == '__main__':
+    FLAGS = flags.FLAGS
```
skearnes, line 98: Not necessarily required for this PR, but I think it would be good to convert this module to a library rather than a script. These flags can be converted to function args, for instance, and then removed.

skearnes:
This is awesome!

PR #286, opened 2020-08-07
Add a search interface to the Postgres DB.

antonkast-google:
First start Postgres, e.g. with docker/serve.sh. Then start the web server with ./serve.sh. A demo is currently running at http://104.154.95.96:8080/

The user interface design is not final; it is just meant to demonstrate the query semantics. Those semantics are modeled in a machine-friendly form with a new class in query.py called "Predicate".

The full query structure is maintained in the URL so results can be hyperlinked.

This introduces a new dependency on jQuery-UI to make the little similarity threshold slider.


ord_schema/interface/query.py
```diff
@@ -25,6 +27,155 @@
...
+        @classmethod
+        def from_name(cls, name):
+            """Take a matching criterion from a URL param."""
+            return Predicate.MatchMode[name.upper()]
+
+    def __init__(self):
+        # List of pairs, (SMILES/SMARTS, MatchMode).
+        self.inputs = []
+        # A pair, (SMILES/SMARTS, MatchMode).
+        self.output = None
```
skearnes, line 49: This should support multiple outputs.

ord_schema/interface/query.py
```diff
@@ -25,6 +27,155 @@
...
+        self.tanimoto_threshold = threshold
+
+    def use_stereochemistry(self):
+        self.do_chiral_ss = True
+
+    def _inputs_predicate(self):
+        exprs = []
+        for pattern, mode in self.inputs:
+            if mode == Predicate.MatchMode.EXACT:
+                expr = f"inputs.smiles = '{pattern}'"
```
skearnes, line 81: This should use the `%s` and `Identifier` syntax in psycopg2; see https://www.psycopg.org/docs/sql.html and https://www.psycopg.org/docs/usage.html#passing-parameters-to-sql-queries

ord_schema/interface/query.py
```diff
@@ -25,6 +27,155 @@
...
+            elif mode == Predicate.MatchMode.SIMILAR:
+                expr = f"rdki.mfp2%morganbv_fp('{pattern}')"
+            elif mode == Predicate.MatchMode.SUBSTRUCTURE:
+                expr = f"rdki.m@>'{pattern}'"
+            elif mode == Predicate.MatchMode.SMARTS:
+                expr = f"rdki.m@>'{pattern}'::qmol"
+            exprs.append(expr)
+        return ' AND '.join(exprs) or 'TRUE'
+
+    def _output_predicate(self):
```
skearnes, line 91: I think input and output queries are similar enough that they shouldn't duplicate code (they only differ in the table name).

ord_schema/interface/query.py
```diff
@@ -25,6 +27,155 @@
...
+            SELECT DISTINCT reactions.serialized
+            FROM reactions
+            INNER JOIN inputs ON inputs.reaction_id = reactions.reaction_id
+            INNER JOIN outputs ON outputs.reaction_id = reactions.reaction_id
+            INNER JOIN rdk.inputs rdki ON rdki.reaction_id = reactions.reaction_id
+            INNER JOIN rdk.outputs rdko ON rdko.reaction_id = reactions.reaction_id
+            WHERE
+                {self._inputs_predicate()}
+                AND {self._output_predicate()}
+                AND {self._reactions_predicate()}
```
skearnes, line 134: I think reaction_smiles should be mutually exclusive with input/output matchers; they are redundant. @connorcoley do you agree?

ord_schema/interface/query.py
```diff
@@ -25,6 +27,155 @@
...
+            LIMIT 100;
+        """
+
+    def query_ids(self):
```
skearnes, line 138: This query should just be a simple lookup in the `reactions` table. Probably better to add this to the `OrdPostgres` class.

ord_schema/interface/query.py
```diff
@@ -48,6 +199,47 @@ def __init__(self, dbname, user, password, host, port):
     def cursor(self):
         return self._connection.cursor()
 
+    def predicate_search(self, predicate):
```
skearnes, line 202: The naming here is a bit ambiguous; `predicate` is actually a set of predicates.

ord_schema/interface/query.py
```diff
@@ -25,6 +27,155 @@
...
+
+    def use_stereochemistry(self):
+        self.do_chiral_ss = True
+
+    def _inputs_predicate(self):
+        exprs = []
+        for pattern, mode in self.inputs:
+            if mode == Predicate.MatchMode.EXACT:
+                expr = f"inputs.smiles = '{pattern}'"
+            elif mode == Predicate.MatchMode.SIMILAR:
```
skearnes, line 82: It seems like we could avoid a lot of code duplication if we adjusted/added `OrdPostgres` methods to return `psycopg2.sql.Composable` objects that would match these predicates. I'm happy to work on that if you'd like.

https://www.psycopg.org/docs/sql.html#psycopg2.sql.Composable

skearnes:
In general I think the interface could be simplified by mirroring the predicate structure (which I like) in the UI. For instance, the top-level choice would be (1) reaction ID, (2) reaction SMARTS, or (3) inputs/outputs. 

The inputs/outputs tab/section would have a list of predicates (w/ add/remove options)---each input/output predicate would be independent, representing a single molecule, and have selectors for (a) input/output, (b) exact/substructure/similarity, (c) similarity threshold.

ord_schema/interface/search.py
```diff
@@ -0,0 +1,126 @@
...
+        smiles, mode_name = output.split(';', 1)
+        mode = query.Predicate.MatchMode.from_name(mode_name)
+        predicate.set_output(smiles, mode)
+
+    if similarity is not None:
+        predicate.set_tanimoto_threshold(float(similarity))
+
+    if reaction_ids or reaction_smiles or inputs or output:
+        db = connect()
+        reaction_ids = db.predicate_search_ids(predicate)
```
n8kim1, line 101: Would it be a good idea to add error handling here? (I typo'd a SMILES string and caused a crash)

ord_schema/interface/search.html
```diff
@@ -0,0 +1,387 @@
...
+        let path = '/?'
+        let hasReagent = false;
+
+        // Inputs.
+        const inputs = $('.input.edittext');
+        inputs.each((index, node) => {
+          const input = $(node);
+          if (input.is(':hidden')) {
+            // The template.
+            return;
```
n8kim1, line 255: When the "reactions" tab is shown (so reagents are hidden), this also excludes all inputs in the form -- not just the template.

ord_schema/interface/query.py
connorcoley: Yes, I agree

skearnes:
I'd like to take another pass on the predicate logic on the backend; @antonkast-google and I discussed and it might be simplest to merge this and edit rather than trying to merge PRs.

Anton: Please pull the latest changes into your branch so this can be merged into `main` and then I'll merge. Thanks!

PR #285, opened 2020-08-07
Properly compile and lint Javascript in CI

n8kim1:
Fixes #277. In short, emulates the proper build/compile process found in the Makefile, and only shows warnings for our own written code.

Works how (I think) we want it to, see results of `lint_javascript` [in this CI run](https://github.com/Open-Reaction-Database/ord-schema/runs/960097580). However, the commands look horrific right now -- will clean up.

PR #292, opened 2020-08-10
Make formatters a check rather than a fix

### LINKED ISSUE #288: linting a pre-merge check, not a code bot
If the lint checks were a merge condition rather than a file-modifying daemon, then I could run yapf when I feel like it instead of facing a more time consuming and repetitive merge chore when I push, and there would be no effect on conformance in the repo.

skearnes:
Fixes #288 

* Moves linters and formatters to `cleanup.yml` (now `run_tests.yml` only runs actual tests)
* Runs linters and formatters in "check" mode; actions will fail if files are not formatted properly
* Adds `format.sh` script for formatting files locally (which can then be committed)

I'm happy to block on #285 to make merging easier for @n8kim1 if you'd like.


connorcoley:
`format.sh` isn't very cross-platform friendly (e.g., uses `find`), and doesn't provide any checks that things are installed (e.g., Go). I'm having trouble getting `addlicense` to run locally. Having to set everything up to run `format.sh` locally will slightly raise the barrier to contributing. 

Would a compromise to @antonkast-google 's comments be to have the _option_ to run this script locally but still include it as an action?

skearnes:
> `format.sh` isn't very cross-platform friendly (e.g., uses `find`), and doesn't provide any checks that things are installed (e.g., Go). I'm having trouble getting `addlicense` to run locally. Having to set everything up to run `format.sh` locally will slightly raise the barrier to contributing.
> 

This works on mac and linux, but I agree that not checking for things being installed is annoying. I did include the installation commands in comments, but they only work on linux.

> Would a compromise to @antonkast-google 's comments be to have the _option_ to run this script locally but still include it as an action?

In general I agree with Anton that auto-commits are a bit disruptive and non-standard (we started doing this as part of the submission pipeline to keep the directory layout sane), but I don't feel strongly about this and think your compromise is reasonable.

skearnes:
Whatever we decide, I'm pretty happy with the refactoring I did here and it's small changes to go back to auto-commits.

skearnes:
PTAL and try `format.sh` again...it's a bit friendlier now

connorcoley:
> PTAL and try `format.sh` again...it's a bit friendlier now

Thanks for these changes

n8kim1:
Having some trouble running format.sh -- adding licenses isn't working... The script is trying to add licenses to all the files I have (including the gitignore'd third-party code), and so it's spitting out "too many open files" for each of these files.

Is there a workaround? I don't want to temporarily delete the third-party files, but can't think of much better.

skearnes:
Can you just comment out that line in format.sh temporarily? I'll look into it.

n8kim1:
Yep, thanks!

PR #295, opened 2020-08-10
Fix JS linter semicolons

n8kim1:
Fixes some issues regarding semicolons that the linter has with our code. (yes, misleading branch name -- whoops)

TODO in other PRs: use goog.module; add JSDocs; change linter to throw errors rather than warnings


n8kim1:
I'd rather get these changes in main first; would save some headache later on. Other PRs still need to address JSDocs and goog.module

PR #297, opened 2020-08-11
Use goog.module for Closure modules

n8kim1:
`goog.provide` is deprecated , and the linter prefers us to use `goog.module`. This requires a lot of changes to how we import/export methods... (see [here](https://github.com/google/closure-library/wiki/goog.module:-an-ES6-module-like-alternative-to-goog.provide) for more details)

### DIFF
--- a/editor/html/dataset.html
+++ b/editor/html/dataset.html
@@ -66,14 +66,14 @@ limitations under the License.
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
     <script src="/js/dataset.js"></script>
   </head>
-  <body onload="init('{{ file_name }}');">
+  <body onload="ord.dataset.init('{{ file_name }}');">
 
     <center>
       <h1><a href="/">Dataset</a>: {{ file_name }}.pbtxt</h1>
     </center>
     <div id="top_buttons">
-      <div id="download" onclick="download();">download</div>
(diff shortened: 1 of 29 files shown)


antonkast-google:
If it were just me, I wouldn't do this. Linter warnings are not expensive. And I haven't heard about any material benefit. 

If you still want to do this, then I would look out for things affected by the namespacing. In particular, I would check manually that all the radio button groups still respect mutual exclusivity.

skearnes:
I think it might be ok to be more relaxed in a more constrained software development environment, but in the open source world where literally anyone can contribute I think it pays to lock things down early on. Plus it's not clear how we could promote some warnings to errors and leave others alone---the documentation one is something I definitely want to see fixed.

skearnes:
Hi Nathan, what's the status on this PR?

n8kim1:
@skearnes a little slow, but still going! Sorry I forgot to push commits.

skearnes:
This LGTM. I prefer the new scoping/definition syntax to the old. The `export` logic is a bit annoying but it's also not crazy---it allows you to explicitly define the public API and not expose internal methods. It would be nice if it that were a bit simpler, but this isn't terrible and reminds me of Python's `__all__`: https://stackoverflow.com/questions/44834

n8kim1:
Interesting! And thanks for the check -- glad that the export logic is acceptable. I'm open to it being changed down the road.

PR #303, opened 2020-08-11
Force Flask redirects to use SSL.

antonkast-google:


antonkast-google:
This is the wrong solution. I will fix this using a second GCP load balancer instead.
https://cloud.google.com/load-balancing/docs/https/setting-up-http-https-redirect

PR #308, opened 2020-08-12
Updates to submission workflow

skearnes:
* Reverts changes `process_dataset.py` so we can post comments on PRs again (from the same repo)
* Doesn't count empty IDs in the statistics
* Removes the documentation about enabling GitHub Actions in forks since this is no longer necessary


skearnes:
Added some additional edits to the submission workflow docs; PTAL

PR #312, opened 2020-08-12
Add docs on instance groups

skearnes:


docs/developer/editor.rst
```diff
@@ -0,0 +1,85 @@
...
+Create a new VM instance
+************************
+
+.. IMPORTANT::
+   If you are only updating the docker image, **you only need to reset the
+   existing VM rather than create a new one!**
+
+.. NOTE::
+   Only one instance at a time can access a persistent disk. Make sure to shut
+   down (and delete) any existing instance before proceeding.
```
antonkast-google, line 44: In addition to shut-down and delete, is any cleanup necessary for the instance group?

skearnes:
Note that I removed the note about restarting to pick up the new image since I've seen the opposite lately.

docs/developer/editor.rst
skearnes: Yes, updated.

PR #316, opened 2020-08-13
Updates to editor Dockerfile

skearnes:
* Rearranges some logic in the docker file to make iterative builds faster
* Adds the `procps` package to the image so we can run things like `ps` and `top` in the container
* Turns off Flask development mode (I have verified that we still see exceptions in the error logs in production mode)
* Removes the ORD_EDITOR_PORT variable since our load balancer uses the Flask default (port 5000)

### DIFF
--- a/editor/Dockerfile
+++ b/editor/Dockerfile
@@ -15,7 +15,7 @@
 FROM continuumio/miniconda3
 WORKDIR /usr/src/app
 
-RUN apt-get update && apt-get install --yes build-essential unzip
+RUN apt-get update && apt-get install --yes build-essential procps unzip
 RUN conda install -c rdkit rdkit
 RUN conda install flask nodejs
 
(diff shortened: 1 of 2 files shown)


editor/Dockerfile
```diff
@@ -29,18 +29,20 @@ RUN tar -xzf v20200517.tar.gz
...
 RUN mv closure-library-20200517 ord-schema/editor/
 RUN mv ketcher ord-schema/editor/
 RUN mv protobuf ord-schema/editor/
 
-# Build the editor.
 WORKDIR ord-schema
+# Bust the cache to get the latest commits; see
+# https://stackoverflow.com/a/39278224.
+ADD https://api.github.com/repos/Open-Reaction-Database/ord-schema/git/refs/heads/main ord-schema-version.json
+RUN git pull
```
antonkast-google, line 41: Just curious, is there a difference for the cache-bust between "clone" and "pull"?

skearnes: No, but the clone takes longer and I wanted to cache moving those deps like ketcher since that can take a long time.

Issue #315, opened 2020-08-13
Changing "Amount" selectors for one component affects others


connorcoley:
On the web editor, when you choose between mass/moles/volume for one component in an input, the fields in all other components also change (e.g., showing mass units when mass is selected).

PR #317, opened 2020-08-13
Autohide options

connorcoley:
Showing/hiding indicators for one component was affecting other components (see #315). The selectors for setting limiting reactors, unit selectors, and whether the volume includes solutes were improperly scoped.

The delay for validation is to ensure that all the data is loaded onto the form and all fields are shown. It would be ideal to not have a fixed time delay, but we don't have callbacks set up for all of these operations.

### DIFF
--- a/editor/js/reaction.js
+++ b/editor/js/reaction.js
@@ -65,7 +65,7 @@ async function init(fileName, index) {
   loadReaction(reaction);
   clean();
   // Trigger reaction-level validation.
-  validateReaction();
+  setTimeout(validateReaction, 1000);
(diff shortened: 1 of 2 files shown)


editor/js/reaction.js
```diff
@@ -65,7 +65,7 @@ async function init(fileName, index) {
   loadReaction(reaction);
   clean();
   // Trigger reaction-level validation.
-  validateReaction();
+  setTimeout(validateReaction, 1000);
```
skearnes, line 68: This only happens once, on startup, right?

connorcoley: Yes

PR #321, opened 2020-08-14
Add check_path to serve.py

### LINKED ISSUE #249: Add time limit on editor tests / error handling in `make test`
Running the editor test with a dataset.pbtxt file that cannot be parsed properly (e.g., invalid enum option) causes the action to hang indefinitely, rather than fail

skearnes:
Changes to identify and prevent directory traversal attacks on the editor. Please let me know if you see a better way to handle this.

Also fixes #249


editor/py/serve.py
```diff
@@ -38,8 +38,37 @@
...
+
+    Raises:
+        ValueError: If the path is not deemed safe.
+    """
+    if root is None:
+        root = _DB_ROOT
+    if user is None:
+        user = flask.g.user
+    abspath = os.path.abspath(path)
+    if not abspath.startswith(os.path.abspath(os.path.join(root, user))):
```
connorcoley, line 69: Is `flask.g.user` necessarily safe? I feel like `os.path.join(root, user)` isn't the safest to check against. Since we have no concept of permissions, I would probably suggest that you just check if the path starts with `_DB_ROOT`

editor/py/serve.py
```diff
@@ -38,8 +38,37 @@
...
+        raise ValueError(f'path is not safe: {path}')
```
connorcoley, line 70: I don't know what best practices are, but this seems like a time when using something other than a `ValueError` makes sense. `PermissionError`? Or a custom `AccessError`?

editor/py/serve.py
skearnes: That's fair; will do.

editor/py/serve.py
skearnes: Good idea; will do.

skearnes: Changed to `PermissionError`.

editor/py/serve.py
skearnes: Changed to only check for `_DB_ROOT`. I left the checks in the user section that don't rely on user input just for consistency.

connorcoley: LGTM

test_requirements.txt
```diff
@@ -1 +1,2 @@
 docker>=4.2.2
+flask>=1.1.2
```
antonkast-google, line 2: Should this be in requirements.txt?

skearnes:
> As a style concern, what do you think about adding a function, e.g. "get_path(file_name)" that would do the os.path.join() in just one place, do the check_path() validation, and return the PermissionError if necessary using flask.abort()?
> 
> Hopefully that would eliminate many lines of code and prevent the accidental introduction of such vulnerabilities in the future by reducing mandatory boilerplate.

Excellent idea. flask.abort() will avoid all the extra try/except boilerplate.

test_requirements.txt
skearnes: Done.

skearnes:
PTAL

PR #323, opened 2020-08-19
Report python test coverage

skearnes:


skearnes:
> Is the goal 100% or is this to just make sure we don't forget to write tests?

Mostly the latter. There's diminishing returns as you approach 100%, but it will help us be more aware of what we're missing. The CodeCov app also provides an overlay tool so we can see exactly which *lines* are covered.

PR #326, opened 2020-08-21
Updates to serve.py

skearnes:
* Alphabetize imports and move non-routed methods to the bottom with the others.
* Use StringIO instead of temporary files (and adjust dataset templating functions to allow file-like objects).
* Use StringIO instead of BytesIO; we're using `MessageToString`, not `MessageToBytes`. Also change input/output `open` calls to use string format.
* Use `send_from_directory` rather than `send_file` when not using a buffer.
* Use status code 307 for redirects to preserve the request method (the default 302 converts all requests to GET).

### DIFF
--- a/editor/py/serve.py
+++ b/editor/py/serve.py
@@ -15,56 +15,28 @@
 
 import contextlib
 import fcntl
+import io
+import json
 import os
 import re
-import json
-import uuid
-import io
-import tempfile
 import urllib
+import uuid
 
 import flask
 from google.protobuf import text_format
 
+from ord_schema import dataset_templating
+from ord_schema import message_helpers
+from ord_schema import updates
+from ord_schema import validations
 from ord_schema.proto import dataset_pb2
 from ord_schema.proto import reaction_pb2
-from ord_schema import message_helpers
-from ord_schema import validations
-from ord_schema import dataset_templating
-from ord_schema import updates
 from ord_schema.visualization import generate_text
 from ord_schema.visualization import drawing
 
 # pylint: disable=invalid-name,no-member,inconsistent-return-statements
 app = flask.Flask(__name__, template_folder='../html')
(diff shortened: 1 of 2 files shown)


ord_schema/dataset_templating.py
```diff
@@ -52,12 +52,13 @@
 flags.DEFINE_boolean('validate', True, 'If True, validate Reaction protos.')
 
 
-def read_spreadsheet(file_name):
+def read_spreadsheet(file_name_or_buffer, suffix=None):
```
connorcoley, line 55: Have you tested this with a buffer? I don't know how the calls to `os.path.splitext` work in that case.

editor/py/serve.py
```diff
@@ -151,12 +121,9 @@ def enumerate_dataset():
...
     data = flask.request.get_json(force=True)
     basename, suffix = os.path.splitext(data['spreadsheet_name'])
-    temp_dir = get_user_path()
-    with tempfile.NamedTemporaryFile(mode='w', suffix=suffix,
-                                     dir=temp_dir) as f:
-        f.write(data['spreadsheet_data'].lstrip('ï»¿'))
-        f.seek(0)
-        dataframe = dataset_templating.read_spreadsheet(f.name)
+    spreadsheet_data = io.StringIO(data['spreadsheet_data'].lstrip('ï»¿'))
+    dataframe = dataset_templating.read_spreadsheet(spreadsheet_data,
```
connorcoley, line 125: Did you test this at least through the GUI (if not an automated test)?

ord_schema/dataset_templating.py
skearnes: Yes; see the tests in #327. If you try to call this with a buffer and don't provide `suffix` you should get an exception.

editor/py/serve.py
skearnes: It's tested programmatically in #327 and I tested it in the GUI on my `editor-tests` branch that combines #326 and #327.

PR #331, opened 2020-08-23
Improve validation buttons storing methods as attributes

n8kim1:
validateReaction didn't work, because one of the button's onclick methods was missing a parenthesis. Dang...

Also, onclick methods had been stored as custom `button-onclick` attributes in the HTML, and dynamically turned into `onclick` attributes when the editor loaded. This led to my IDE interpreting the attribute method as a generic block of text, not JS, and so it never noticed the error. Will change the code to represent JS as real JS.

### DIFF
--- a/editor/js/reaction.js
+++ b/editor/js/reaction.js
@@ -91,7 +91,7 @@ async function init(fileName, index) {
   loadReaction(reaction);
   clean();
   // Trigger reaction-level validation.
-  setTimeout(validateReaction, 1000);
+  validateReaction();
   // Signal to tests that the DOM is initialized.
   ready();
 }
--- a/editor/js/reaction.js
+++ b/editor/js/reaction.js
@@ -602,9 +602,12 @@ function initCollapse(node) {
 // inserting contents into a div in reaction.html
 function initValidateNode(oldNode) {
   let newNode = $('#validate_template').clone();
-  // Add attributes necessary for validation functions.
+  // Add attributes necessary for validation functions:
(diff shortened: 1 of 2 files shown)


editor/js/reaction.js
```diff
@@ -65,7 +65,7 @@ async function init(fileName, index) {
   loadReaction(reaction);
   clean();
   // Trigger reaction-level validation.
-  setTimeout(validateReaction, 1000);
```
skearnes, line 68: Why is this removed? I thought there was some startup lag that this was intended to control for?

n8kim1: Didn't solve the problem we were originally trying to solve... I figured the timeout wasn't necessary. 

After further thought, even though I can't notice any issues in my own testing, I think the timeout is still safer (perhaps if startup takes a while for example). I'll put it back.

n8kim1: After even _more_ testing, validateReaction should be okay without a timeout. Loading, adding, and starting to show the validation divs all take place before validateReaction is called. Even though we have a "show" transition, the divs are still visible before the transition actually finishes.

I'll stick with no timeout for now...

n8kim1:
@skearnes added a little bit more, could you take another look? Sorry I left it hanging.

skearnes:
Do you need anything else here? I see that the PR is still marked as a draft.

n8kim1:
Nope, just rebase and merge that I'm finally having a chance now -- thanks for the check!

PR #332, opened 2020-08-26
Refactor query.py

skearnes:
* Refactors the query backend to provide three mutually exclusive query types (see the docstring for `query.py`).
* Plumbs the changes through `serve.py` and the associated HTML template.
* Updates `query_longtests.py`; we'll need additional tests here and for `serve.py` later (but this PR is getting big).

### DIFF
--- a/ord_schema/interface/query.py
+++ b/ord_schema/interface/query.py
@@ -40,6 +40,7 @@ For example, a reaction query might have the following predicates:
 Note that a predicate is matched if it applies to _any_ input/output.
 """
 
+import abc
 import binascii
 import enum
 import json
@@ -68,13 +69,14 @@ def fetch_results(cursor):
     return reactions
 
 
-class ReactionQueryBase:
+class ReactionQueryBase(abc.ABC):
     """Base class for reaction-based queries."""
 
+    @abc.abstractmethod
     def json(self):
         """Returns a JSON representation of the query."""
-        raise NotImplementedError()
 
+    @abc.abstractmethod
     def run(self, cursor, limit=None):
         """Runs the query.
 
@@ -86,7 +88,6 @@ class ReactionQueryBase:
         Returns:
             Dict mapping reaction IDs to serialized Reaction protos.
         """
-        raise NotImplementedError()
 
 
 class ReactionIdQuery(ReactionQueryBase):
--- a/ord_schema/interface/query.py
+++ b/ord_schema/interface/query.py
@@ -42,6 +42,7 @@ Note that a predicate is matched if it applies to _any_ input/output.
 
 import binascii
 import enum
+import json
 
 import psycopg2
 from psycopg2 import sql
@@ -68,6 +69,10 @@ def fetch_results(cursor):
 class ReactionQueryBase:
     """Base class for reaction-based queries."""
 
+    def json(self):
+        """Returns a JSON representation of the query."""
+        raise NotImplementedError()
+
     def run(self, cursor, limit=None):
         """Runs the query.
 
@@ -92,6 +97,10 @@ class ReactionIdQuery(ReactionQueryBase):
             reaction_ids: List of reaction IDs.
         """
         self._reaction_ids = reaction_ids
+
+    def json(self):
+        """Returns a JSON representation of the query."""
+        return json.dumps({'reactionIds': self._reaction_ids})
 
     def run(self, cursor, limit=None):
         """Runs the query.
@@ -120,6 +129,10 @@ class ReactionSmartsQuery(ReactionQueryBase):
             reaction_smarts: Reaction SMARTS.
         """
         self._reaction_smarts = reaction_smarts
+
+    def json(self):
+        """Returns a JSON representation of the query."""
+        return json.dumps({'reactionSmarts': self._reaction_smarts})
 
     def run(self, cursor, limit=None):
         """Runs the query.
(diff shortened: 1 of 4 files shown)


skearnes:
@connorcoley This has support for searching by

* Reaction ID
* Reaction SMARTS
* Reaction components (input/output; exact/substructure/similarity/smarts)

Do we also want to enable similarity searching with reaction SMILES? I'd prefer to keep the options fewer if we don't anticipate this being very useful/popular.

skearnes:
@antonkast-google I've finished the first pass of plumbing this through `search.{py,html}`. I expect there will be more changes as I start writing tests and testing the updated template (which I probably butchered due to my lack of HTML/JS skills).

ord_schema/interface/query.py
```diff
@@ -20,15 +47,252 @@
...
+
+    Args:
+        cursor: psycopg.cursor instance.
+
+    Returns:
+        Dict mapping reaction IDs to serialized Reaction protos.
+    """
+    reactions = {}
+    for reaction_id, serialized in cursor:
+        reactions[reaction_id] = serialized
```
antonkast-google, line 67: Is result ranking never meaningful? We'll never care about ORDER BY?

ord_schema/interface/query.py
```diff
@@ -20,15 +47,252 @@
...
+    return reactions
+
+
+class ReactionQueryBase:
```
antonkast-google, line 71: I know nothing about python inheritance. What is the purpose of a base class with no implementation? Is this a case for abc?

ord_schema/interface/query.py
```diff
@@ -20,15 +47,252 @@
...
+        Returns:
+            Dict mapping reaction IDs to serialized Reaction protos.
+        """
+        del limit  # Unused.
+        query = sql.SQL("""
+            SELECT DISTINCT reaction_id, serialized 
+            FROM reactions 
+            WHERE reaction_id = ANY (%s)""")
+        args = [self._reaction_ids]
+        logging.info('Running SQL command:%s',
```
antonkast-google, line 123: whitespace at ":%s"?

ord_schema/interface/query.py
```diff
@@ -20,15 +47,252 @@
...
+                     cursor.mogrify(query, args).decode())
+        cursor.execute(query, args)
```
antonkast-google, line 125: Does this do the right thing in case of multiple reaction IDs? I don't know psycopg, but I would have expected something to say how to convert the list to a string, e.g. a delimiter.

ord_schema/interface/query.py
```diff
@@ -20,15 +47,252 @@
...
+        return json.dumps({
+            'useStereochemistry': self._do_chiral_sss,
+            'similarity': self._tanimoto_threshold,
+            'components': [predicate.json() for predicate in self._predicates],
+        })
 
-class Predicate:
-    """Structure and code generation for ORD query predicates."""
+    def _setup(self, cursor):
+        """Prepares the database for a query.
```
antonkast-google, line 199: This is fine, but I thought things were more readable when all the SQL was a simple string. Is there an advantage in going command/args/logging/execute for each statement? Maybe something about sql.SQL?

ord_schema/interface/query.py
```diff
@@ -20,15 +47,252 @@
...
+            return {}
+        self._setup(cursor)
+        components = [
+            sql.SQL("""
+            SELECT DISTINCT reaction_id, serialized
+            FROM reactions """)
+        ]
+        components.extend(self._get_tables())
+        components.append(sql.SQL("""
+            WHERE """))
```
antonkast-google, line 272: whitespace? Is the goal to come out later with a pretty-printed query string? Where can we see that?

ord_schema/interface/query.py
```diff
@@ -200,148 +390,27 @@ def __init__(self, dbname, user, password, host, port):
...
-                _, serialized = result
-                reaction = reaction_pb2.Reaction.FromString(
-                    binascii.unhexlify(serialized.tobytes()))
-                reactions.append(reaction)
-            self._connection.rollback()
-        return dataset_pb2.Dataset(reactions=reactions)
+        if return_ids:
+            return dataset_pb2.Dataset(reaction_ids=reactions.keys())
+        unserialized = []
+        for serialized in reactions.values():
```
antonkast-google, line 412: Made a dictionary then threw away the keys?

Could you skip a step and preserve the DB result order?

ord_schema/interface/query.py
skearnes: I don't think the order is ever meaningful.

ord_schema/interface/query.py
skearnes: Yes, `abc` would be more explicit; I'll update.

ord_schema/interface/query.py
skearnes: Not needed since the query starts with a newline in the heredoc.

ord_schema/interface/query.py
skearnes: Yes, note that `self._reaction_ids` is a list, so `args` is a list of lists. See also `test_reaction_id_query` in `query_longtest.py`.

ord_schema/interface/query.py
skearnes: The psycopg2 documentation is very clear about this:

> __Warning__ Never, __never__, __NEVER__ use Python string concatenation (+) or string parameters interpolation (%) to pass variables to a SQL query string. Not even at gunpoint.

https://www.psycopg.org/docs/usage.html#the-problem-with-the-query-parameters

The `sql` module is the preferred way to construct queries where you need to use variables for SQL identifiers like table names. Simple queries that have static identifiers should use the `%s` syntax and pass `args` to `execute`.

ord_schema/interface/query.py
skearnes: There is logging here that shows the complete (and variable-substituted) string; if you run the tests it shows up in the console.

ord_schema/interface/query.py
skearnes: The dictionary keys are used if `return_ids=True`. I don't think the result order is important.

skearnes:
> Do the URLs still work?

Yep! I've tested `/id/<reaction_id>` and the reaction ID lookup in the form on `/`.

skearnes:
I've done more testing and added a few fixes. There's some cleanup to do on the style sheet, I think, but it's basically functional.

PR #336, opened 2020-08-31
Refactor generate_text.py into a library and script

### LINKED ISSUE #284: Convert visualization.generate_text to a library rather than a script
> Not necessarily required for this PR, but I think it would be good to convert this module to a library rather than a script. These flags can be converted to function args, for instance, and then removed.

_Originally posted by @skearnes in https://github.com/Open-Reaction-Database/ord-schema/pull/283_

skearnes:
Fixes #284

### DIFF
--- a/ord_schema/visualization/generate_text.py
+++ b/ord_schema/visualization/generate_text.py
@@ -11,35 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Text generation script for Reaction messages.
-
-This script is meant to convert a Reaction message into a full-fledged text
-paragraph that would be appropriate for inclusion in a Supplemental Information
-document for a publication.
-
-Example usage:
-* For normal operation from a pbtxt
-  $ python generate_text.py --input_reaction=reaction.pb --input_format=pbtxt
-      --type text
-"""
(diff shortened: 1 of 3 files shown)


ord_schema/visualization/generate_text.py
```diff
@@ -69,38 +47,15 @@ def _generate(reaction, template_string, line_breaks):
 
 def generate_text(reaction):
     """Generates a textual reaction description."""
-    return _generate(reaction, _TEXT_TEMPLATE, False)
+    with open(os.path.join(os.path.dirname(__file__), 'template.txt'),
```
connorcoley, line 50: Is there a reason you prefer this over defining a module-level constant for `_HTML_TEMPLATE` and `TEXT_TEMPLATE`? Requiring each call to `generate_text` or `generate_html` to read a file from disk seems somewhat inefficient.

skearnes: I guess I'm just generally avoidant of complex global initialization...these are only mostly constants since they can vary with whatever is in the file and that seems to rub against https://google.github.io/styleguide/pyguide.html#25-global-variables

I agree it's less efficient but I doubt it will noticeably affect performance. That said, I'll change it back if you feel strongly about this.

connorcoley: I feel weakly about it, so we can keep it like you have

PR #337, opened 2020-08-31
Require provenance in datasets

### LINKED ISSUE #273: When to add new event records
In #271, @connorcoley wrote:

> when validating submissions, if we detect that a user has modified a Reaction but they haven't added an event_modified RecordEvent, we should raise an error.

I'm not sure I agree with this. I think that puts a pretty large burden on users and doesn't give much benefit---we already have the `record_created` details. In fact, I'm wondering if we should not be adding `record_modified` events during submission at all for the same reason---they don't specify *what* was changed so they are not adding anything that we don't already know from the user-provided `record_created` event.

I propose that `record_modified` only be added when an existing record (already submitted) is modified by a different PR. Each PR should add just one event, either `record_created` (submission) or `record_modified` (modification).

Thoughts?

skearnes:
Fixes #273 

Changes:
* Replaces `validate_ids` with `strict`
* Strict checks include ID validation as well as provenance checks

I struggled with the different levels of strictness that are required by the editor and the submission workflow. The former doesn't care about reaction IDs but both care about provenance. My compromise is to allow empty reaction IDs when strictly validating Reactions but not Datasets. This allows the editor to use `strict=True` without emitting warnings about bad IDs.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -251,8 +251,9 @@ including required fields and checks for consistency across messages.
   message.
 * If `reaction_id` is set for a `Reaction` in `reactions`, it must be unique.
 * Each entry in `reaction_ids` must match `^ord-[0-9a-f]{32}$`.
-* When run with `validate_ids=True` (default is False), `dataset_id` must
-  match `^ord_dataset-[0-9a-f]{32}$`.
+* When run with `strict=True` (default is False):
+  * `dataset_id` must match `^ord_dataset-[0-9a-f]{32}$`.
+  * Each entry in `reactions` must have a defined `reaction_id`.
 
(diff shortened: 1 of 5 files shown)


skearnes:
@connorcoley I added ValidationOptions per our offline discussion; PTAL

skearnes:
> For defaults, should `require_provenance` be true? Or can you also change the editor's `serve.py` to use the non-default options in this PR before merging?
> 
> Otherwise LGTM - thanks.

I set `require_provenance=True` for the editor in https://github.com/Open-Reaction-Database/ord-schema/pull/337/commits/f220d5cca3ddac699ed12e68d42f444c681e98eb.

PR #339, opened 2020-09-05
Add JS docstrings

### LINKED ISSUE #293: Fix JS warnings
The current JS linter uses `--jscomp_warning` and doesn't fail on (many) warnings. I'd like to change this to `--jscomp_error`.

connorcoley:
Fixes #293


connorcoley:
I would appreciate any input from @skearnes @antonkast-google or @n8kim1 about these docstrings. If you like them, I'd suggest that the load/unload descriptions can probably be used as a template for the other files. Currently, I've only updated `amounts.js`, `amountCrudes.js`, and `compounds.js` so they pass the linter.

P.S. I'm following https://google.github.io/styleguide/jsguide.html#jsdoc

skearnes:
Thanks Connor, this looks great! I'll start from the bottom of the directory and work my way up. We can meet somewhere near the middle (and Nathan, if you join in maybe start at inputs.js or nodes.js?)

skearnes:
FYI just pushed workups.js to make sure everything works with my setup; I was slightly less verbose than Connor :)

antonkast-google:
It's great to document all this. I should have been doing that myself.

A couple suggestions based on personal experience:

1) Comments should add information, e.g. avoid "setValue() /* Sets the value. */".

2) Type annotations are great so long as there is enforcement. Otherwise, comments are almost always read from source, so leaving out @param etc. is actually more comprehensible.

skearnes:
Note that `@param` is still useful even without type annotations as structured documentation.

Do you know how to adjust the compilation to enforce types at compile- and run-time? We should do this if we can.

skearnes:
@connorcoley note that `@return` does not need a variable name; https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments

connorcoley:
Right, that makes sense. Thanks

skearnes:
Good to go! @antonkast-google @connorcoley @n8kim1 PTAL since I had to guess at what was going on in some places.

editor/js/outcomes.js
```diff
@@ -268,6 +313,12 @@ function unloadAnalysis(analysisNode, analyses) {
   }
 }
 
+/**
+ * Fetches a processed_data record defined in the form and adds it to
+ * `processes`.
+ * @param {!Node} node Root node for the Data record.
+ * @param {!jspb.Map<string, !proto.ord.Data>} processes
```
connorcoley, line 320: I don't love the variable name `processes` here (and previous function names like `loadProcess`). Is it worth changing this to processedData, loadProcessedData, etc.?

editor/js/reaction.js
```diff
@@ -239,6 +286,11 @@ function downloadReaction() {
   xhr.send(binary);
 }
 
+/**
+ * Downloads a dataset as a serialized Dataset proto.
+ * @param {string} fileName The name of the dataset to fetch.
+ * @return {!Uint8Array}
```
connorcoley, line 292: It's more correct to say that this function returns a `Promise`. It seems like there is a lot of discussion around how to document type when it is fulfilled or rejected, but no clear consensus. Google's guide seems to suggest you would want to use `!Promise<!Uint8Array>` here

editor/js/reaction.js
```diff
@@ -262,7 +319,12 @@ function putDataset(fileName, dataset) {
   xhr.send(binary);
 }
 
-// For testing, compare a local Dataset to a Dataset on the server.
+/**
+ * Compares a local Dataset to a Dataset on the server (used for testing).
+ * @param {string} fileName The name of a dataset on the server.
+ * @param {!proto.ord.Dataset} dataset A local Dataset.
+ * @return {!Uint8Array}
```
connorcoley, line 326: Same comment re: `!Promise<!Uint8Array>`

editor/js/reaction.js
```diff
@@ -401,7 +474,12 @@ function isEmptyMessage(obj) {
   return false;
 }
 
-// The template is a jQuery selector. The root may be a jQuery object.
+/**
+ * Adds an instance of `template` to the root node.
+ * @param {string} template A jQuery selector.
+ * @param {!Node} root A jQuery object.
+ * @return {!Node}
```
connorcoley, line 481: Could add "The new copy of the template" to the return documentation

editor/js/reaction.js
```diff
@@ -679,7 +825,12 @@ function initValidateHandlers() {
   });
 }
 
-// Convert a Message_Field name from a data-proto attribute into a proto class.
+/**
+ * Converts a Message_Field name from a data-proto attribute into a proto class.
+ * @param {string} protoName Underscore-delimited protocol buffer field name,
+ *     such as Reaction_provenance.
+ * @return {!typeof jspb.Message|null}
```
connorcoley, line 832: Is this the correct way to record the type, or should it be `{?typeof jspb.Message}`?

editor/js/conditions.js
```diff
@@ -107,10 +115,15 @@ function unload() {
   return conditions;
 }
 
+/**
+ * Validates the reaction conditions defined in the form.
+ * @param {!Node} node Root node for the reaction conditions.
+ * @param {!Node} validateNode Target node for validation results.
```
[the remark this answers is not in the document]
skearnes: Done for all instances.

editor/js/dataset.js
```diff
@@ -80,6 +100,13 @@ function download() {
   xhr.send();
 }
 
+/**
+ * Fetches a dataset from the server.
+ * @param {string} fileName The filename of the dataset to fetch.
+ * @param {!Function} listener Function used to load the dataset into the
```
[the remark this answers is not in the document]
skearnes: Done.

editor/js/dataset.js
```diff
@@ -80,6 +100,13 @@ function download() {
...
+ *     editor.
+ * @return {!Uint8Array} The serialized Dataset proto.
```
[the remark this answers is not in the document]
skearnes: Removed; I didn't need to add this and the linter doesn't require it.

editor/js/outcomes.js
skearnes: I'll do this as part of #349 so we only change JSDoc in this PR; sound good?

editor/js/reaction.js
skearnes: Done.

editor/js/reaction.js
skearnes: Done.

editor/js/reaction.js
skearnes: Done.

skearnes:
Thanks for the comments; PTAL.

editor/js/outcomes.js
connorcoley: Yep, sounds good

connorcoley:
LGTM. I can't approve the PR since I'm also an author. Would be good to get another set of eyes on it

editor/js/codes.js
```diff
@@ -80,6 +93,12 @@ function unload(codes) {
   });
 }
 
+/**
+ * Fetches a single automation_code section from the form and adds it to
```
n8kim1, line 97: Is it worth noting which single code this is? (we do describe this in the node param description, so it's not necessary)

antonkast-google:
I will make a pass later today and see if I can add anything.

editor/js/compounds.js
```diff
@@ -306,7 +403,13 @@ function addNameIdentifier(node) {
   xhr.send(name);
 }
 
-// Shortcut to add an identifier by drawing.
+/**
+ * Displays the Ketcher drawing tool for defining compound identifiers. A
+ * callback is defined to create SMILES and MOLBLOCK-type identifiers; this
+ * callback is triggered upon submission from the Ketcher window.
```
n8kim1, line 409: "submission of a molecular drawing" perhaps? Was a little confusing to me at first glance...

editor/js/inputs.js
```diff
@@ -39,12 +43,25 @@ function load(inputs) {
   });
 }
 
+/**
+ * Populates a single reaction input section in the form.
```
n8kim1, line 47: "Creates and populates"? since this method involves a call to `add`, as opposed to loadInputUnnamed which doesn't actually add a new div

n8kim1 (approved):

n8kim1:
Looks good to me (especially annotations); I just have a few small suggestions about descriptions (that I don't feel too strongly about anyways). Would still like Anton's approval too.

Thanks Steve and Connor!

editor/js/codes.js
skearnes: Yeah, I think we could add "specified by `node`" but that feels redundant. Do you agree?

editor/js/compounds.js
skearnes: Done.

editor/js/inputs.js
skearnes: That's fair; I can update this in many places.

skearnes: Done in 746da3e5377414e24279e939c62e2d70df7fc0ef

editor/js/codes.js
n8kim1: Yep that's fair -- as-is is good with me.

antonkast-google (approved):

antonkast-google:
Nothing to add. LGTM

PR #352, opened 2020-09-10
Use current head in editor test

skearnes:
#351 was too simple; it always uses `main` in `Open-Reaction-Database/ord-schema`.


connorcoley (approved):
& this will get the right version even though `GITHUB_REF` is set to `refs/heads/main`?

skearnes:
> & this will get the right version even though `GITHUB_REF` is set to `refs/heads/main`?

Yes, the `--build-arg` flags in the GitHub Action YAML override the default values in the Dockerfile.

PR #355, opened 2020-09-11
Use local state for editor Dockerfile

### LINKED ISSUE #354: Update the editor Dockerfile to use the local context
Currently the default behavior of the Dockerfile is to checkout from `Open-Reaction-Database/ord-schema:main` and there are some lines that can be uncommented to pull in any changes in the `editor/` context. This causes a lot of silent failures where users are expecting the Dockerfile to use whatever branch they are on at the time of the build.

skearnes:
Fixes #354 

This should also make it possible for us to automate docker builds on pushes to `main`.

### DIFF
--- a/editor/Dockerfile
+++ b/editor/Dockerfile
@@ -12,14 +12,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# To build the editor using the current state of ord-schema:
+# $ cd path/to/ord-schema
+# $ docker build --file=editor/Dockerfile -t openreactiondatabase/ord-editor .
+#
+# To test the editor:
+# $ docker run --rm -d -p 5000:5000 --name editor openreactiondatabase/ord-editor
+# $ node editor/js/test.js
+# $ docker stop editor
+#
+# To push the built image to Docker Hub:
+# docker push openreactiondatabase/ord-editor
+
 FROM continuumio/miniconda3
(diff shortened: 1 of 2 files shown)


connorcoley (approved):

editor/Dockerfile
```diff
@@ -12,14 +12,27 @@
...
-WORKDIR /usr/src/app
 
 RUN apt-get update && apt-get install --yes build-essential procps unzip
 RUN conda install -c rdkit rdkit
 RUN conda install flask nodejs
 
 # Fetch/build editor dependencies.
+# NOTE(kearnes): Do this before COPYing the local state so it can be cached.
+WORKDIR /usr/src/app/ord-schema/editor
 RUN git clone https://github.com/Open-Reaction-Database/ketcher.git
```
antonkast-google, line 36: Does this mean we are cloning and building ketcher every time the test runs? Or is caching possible?

I'm hoping this test can run quickly.

antonkast-google (approved):

editor/Dockerfile
skearnes: The test currently runs in ~5 min. The only way to cache on GitHub AFAIK would be to create another image that includes all the steps up to getting the local state and having the Dockerfile inherit from that instead of the current miniconda3 image. We could certainly do this.

PR #361, opened 2020-09-13
Constrain editable fields for numeric inputs to avoid surprising behavior

### LINKED ISSUE #348: Invalid floats disappear from the pbtxt (and resolve as NaN)
...we should consider factoring out a separate function that checks for NaN and raises and exception on invalid input.

### LINKED ISSUE #360: Handling commas in numbers (esp. when used as decimal points)
Reminder: we had some prior discussion about how to handle commas/periods gracefully. This doesn't have to be changed for this PR, but consider the following:

```
>>> parseFloat('123,456')
123
>>> parseFloat('123.456.789')
123.456
```

Should we silently reject these values? Or should we define a new function `on('change')` for each numerical entry field that will strip commas, raise alerts if invalid, etc.?

_Originally posted by @connorcoley in https://github.com/Open-Reaction-Database/ord-schema/pull/358#discussion_r487422999_

connorcoley:
If we can tell that a string will fail to be parsed correctly into a float when the form is packaged into a proto, we should provide an indication to the user. This effectively adds a handler that is triggered every time a numeric field is changed to only allow characters 0-9, a single decimal point, and a leading hyphen. Other characters (e.g., commas) are simply stripped from the field. Because the user will get real-time feedback about what they've entered, I think this behavior is acceptable. It might seem odd at first though:

- "1.234" -> "1.234"
- "1.234.567" -> "1.234567"
- "123,456.78" -> "123456.78"
- "-0.5apple67" -> "-0.567"
- "5-3" -> "53"
- "1.5" -> 15 (`integertext` fields only, which is just the addition order)

Will close #360
Fixes #348

### DIFF
--- a/editor/js/reaction.js
+++ b/editor/js/reaction.js
@@ -149,31 +149,31 @@ function selectText(node) {
 
 /**
  * Ensures that the text entered in a float input is valid by forbidding any
- * charactersides besides 0-9, a single period to signify a decimal, and a
+ * characters besides 0-9, a single period to signify a decimal, and a
  * leading hyphen.
  * @param {!Node} node
  */
 function sanitizeFloat(node) {
   var stringValue = $(node).text();
   var negativeSign = (stringValue[0] === '-' ? '-' : '');
-  stringValue = stringValue.replace(/[^0-9\.]/g,'');
+  stringValue = stringValue.replace(/[^0-9\.]/g, '');
   const matches = stringValue.match(/\./g);
   if (matches && matches.length > 1) {
     var parts = stringValue.split('.');
-    stringValue = parts.shift() + (parts.length ? '.': '') + parts.join('');
+    stringValue = parts.shift() + (parts.length ? '.' : '') + parts.join('');
   }
   $(node).text(negativeSign + stringValue);
 }
 
 /**
  * Ensures that the text entered in an integer input is valid by forbidding any
- * charactersides besides 0-9 and a leading hyphen.
+ * characters besides 0-9 and a leading hyphen.
  * @param {!Node} node
  */
 function sanitizeInteger(node) {
   var stringValue = $(node).text();
   var negativeSign = (stringValue[0] === '-' ? '-' : '');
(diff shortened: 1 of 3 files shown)


editor/js/reaction.js
```diff
@@ -115,7 +115,9 @@ function ready() {
  */
 function listen(node) {
   addChangeHandler($(node), dirty);
-  $('.edittext').on('focus', event => selectText(event.target));
+  $('.edittext', node).on('focus', event => selectText(event.target));
```
connorcoley, line 118: Note: this was improperly scoped before so that every time a new element was added to the form, a new onfocus handler was created. This led to >10 of selectText calls to be triggered every time a user clicked on a text field

skearnes:
How would you feel about making all commas errors? I'm worried about our European friends entering things like `1,23` and having that converted to `123` when they really want `1.23`.

I think I'd suggest the same behavior for all non-allowed characters---make them errors rather than stripping them.

connorcoley:
I struggled to figure out how we would display errors there. What does "error" mean to you? Changing the outline of the input field to be red? A javascript pop-up alert? It can't show up as a validation error since those all come from the Python backend after we package up the message.

skearnes:
> I struggled to figure out how we would display errors there. What does "error" mean to you? Changing the outline of the input field to be red? A javascript pop-up alert? It can't show up as a validation error since those all come from the Python backend after we package up the message.

If possible, I think a red outline or warning icon next to the field would be ideal. Something client-side that can be fast. There's a general pattern of "errors creating the proto before we can even validate it" that this fits into...

skearnes:
(Looks like this should also fix #348)

editor/js/reaction.js
skearnes: Good catch!

connorcoley:
@skearnes PTAL -- now just adds a 2 px red border to the div that contains the invalid value

skearnes (approved):

PR #362, opened 2020-09-14
Sort out docker-compose for the web interface

skearnes:
The following commands assume that the `ord-postgres` image is either built locally or pulled from Docker Hub:

```
$ cd .../ord-schema
$ docker build --file=ord_schema/interface/Dockerfile -t openreactiondatabase/ord-interface .
$ cd ord_schema/interface
$ docker-compose up
```

Now the interface is available at `localhost:5000`

PR #363, opened 2020-09-14
Show a table of results in the query interface

skearnes:
Note that this PR is based on #362

Uses [DataTables](https://datatables.net/) for showing query results. The table is created on demand, so we don't have to wait for the reaction tables/images to render for all of the results before we show the page---only the current page of results is rendered. It's still pretty slow, but maybe we can speed it up by using smaller images?

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -84,8 +84,8 @@ jobs:
           cd "${GITHUB_WORKSPACE}/editor"
           pip install flask pylint yapf
           npm install google-protobuf puppeteer
-          wget https://storage.googleapis.com/ord-editor-test/editor_test_protobuf_1dae8fdd.tar
-          tar -xzf editor_test_protobuf_1dae8fdd.tar
+          wget https://github.com/protocolbuffers/protobuf/releases/download/v3.13.0/protobuf-js-3.13.0.tar.gz
+          tar -xzf protobuf-js-3.13.0.tar.gz
           wget https://github.com/google/closure-library/archive/v20200517.tar.gz
           tar -xzf v20200517.tar.gz
           git clone https://github.com/Open-Reaction-Database/ketcher.git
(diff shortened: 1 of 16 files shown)


skearnes:
PS it's _way_ faster to generate the HTML table using SVGs instead of PNGs...

connorcoley:
Looks great -- much better than just having the pbtxt. A few nitpicky things (which can be improved in time):

- "Search" is really searching/downselected by reaction ID, right?
- The "Reagents" and "Reactions" selector at the top could be clearer if prepended with "Search by". I also think "Reagents" should be switched to "Components" or "Compound" since reagent holds a more particular meaning in synthesis
- Searching by reaction ID (e.g., ord-008793d82f684c548c06ceba947fa47a) took a _very_ long time and I'm not sure why. Is there no index on this field? It could just be my computer.
- I think if you try to search by "Reagents" but have part of the "Reactions" field populated, it affects the query

skearnes:
> Looks great -- much better than just having the pbtxt. A few nitpicky things (which can be improved in time):
> 
> * "Search" is really searching/downselected by reaction ID, right?

Do you mean the search bar on the table? Yes, that's just searching in the table of returned results.

> * The "Reagents" and "Reactions" selector at the top could be clearer if prepended with "Search by". I also think "Reagents" should be switched to "Components" or "Compound" since reagent holds a more particular meaning in synthesis

Agreed.

> * Searching by reaction ID (e.g., ord-008793d82f684c548c06ceba947fa47a) took a _very_ long time and I'm not sure why. Is there no index on this field? It could just be my computer.

Interesting, this always goes pretty quickly for me locally. We can run more tests as we polish the interface in later PRs.

> * I think if you try to search by "Reagents" but have part of the "Reactions" field populated, it affects the query

Yes, "Reactions" has precedence. We should have the query give precedence to whichever tab was active on "Go".

skearnes:
PTAL; updated with #362 base post-merge.

connorcoley (approved):

PR #364, opened 2020-09-15
Use SVG instead of PNG in HTML reaction table

skearnes:
It's _way_ faster and looks just as good.

Notable changes:
* Adds support for SVG cropping. This requires two passes; the first pass is used to identify the extent of the drawn image, and the second pass repeats the drawing on an appropriately-sized canvas.
* The primary use case is the drawings in the query interface, but I updated the uses in the editor as well. I'm not quite sure how to get the component images to center-align (in the Inputs section), but that's minor.

### DIFF
--- a/editor/html/reaction.html
+++ b/editor/html/reaction.html
@@ -195,7 +195,7 @@ limitations under the License.
                     </span>
                   </legend>
                   <!-- Compound rendering -->
-                  <div class="component_rendering" style="margin-left: auto; margin-right: auto;"></div>
+                  <div class="component_rendering" style="text-align: center;"></div>
                   <div class="identifiers"></div>
                   <!-- Shortcuts for name resolving & drawing -->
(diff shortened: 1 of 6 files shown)


connorcoley:
image

Would it be possible to better handle cases where `max_x - min_x == 0` or `max_y - min_y == 0`? These structures (e.g., methanol and water in the figure above) look a little funny. I'm not sure what the best general solution to this would be -- a larger minimum size?

connorcoley:
Second comment from editor reviewing:

Would it be possible to crop the CVG used in the editor rendering? It's not 100% necessary. If we can't figure out how to reliably crop without affecting bond length / sizes for very small compounds, we can leave the viewbox

Centering would be nice. If you add a `max-width:300px` to the outer div, it works well for the uncropped images

skearnes:
Thanks; PTAL I added a `min_size` parameter. There's probably a cleaner way to do this but it works for now.

connorcoley (approved):
Looks great

skearnes:
Fixed component image centering; PTAL.

Issue #335, opened 2020-08-27
Enhancements for the search/browse interface


skearnes:
Here's a list of suggested improvements/milestones for the search & browse interface:

* Polish the form
  * Show a results table rather than just a list of reaction IDs
  * Link to separate landing pages for individual reactions rather than showing the raw pbtxt
* Reaction landing pages
  * Reuse the editor code in read-only mode to display the raw data
  * Visualization of the reaction using the templates [here](https://github.com/Open-Reaction-Database/ord-schema/tree/main/ord_schema/visualization)
  * Links to connected records for multi-step reactions
* GCE deployment (probably with `docker-compose`, since we have separate images for the postgres backend and the interface)

skearnes:
@antonkast-google we talked about a read-only version of the editor; is that still feasible?

skearnes:
What should be in the proposed "results table"? Suggestions:

* yield/conversion
* type (e.g. flow vs. batch)
* limiting reactant(s) structure
* primary product structure

connorcoley:
Is the full HTML table too much?

Other bits of useful information:
- Creation date
- Experimenter information if defined
- Link to source publication / DOI if defined

antonkast-google:
Sure, a reaction reader based on the editor client is feasible. The main chore is to conditionally hide bunch of stuff in reaction.html. Also the client needs a new backend for its XHRs.

skearnes:
> Is the full HTML table too much? Ideally, we would make this also capture batch/flow, use of electrochem/photochem
> 
> Other bits of useful information:
> 
> * Creation date
> * Experimenter information if defined
> * Link to source publication / DOI if defined

I think that these things are certainly useful on the reaction-landing-page level. I don't imagine they will be very useful in a top-level view of the query results (and it might be hard to represent them compactly). I'm thinking of what I've seen in e.g. ChEMBL: https://www.ebi.ac.uk/chembl/g/#search_results/compounds/query=ibuprofen (make sure to click on "Table" view).

skearnes:
The specific things you mentioned would actually be great, but I the full table could be too much...just to clarify.

connorcoley:
I think showing the full reaction drawing (w/o amounts) would be good, but I agree the table might be too much. Chemists are used to Reaxys- or SciFinder-like interfaces for tabulated reactions

skearnes:
[...]

PR #365, opened 2020-09-15
ord.reaction.freeze() makes the editor read-only.

antonkast-google:
Addressing #335, this introduces a global freeze() function in the editor that irreversibly mutates the UI into a read-only state.

This may be useful for example in rendering individual results from queries to the ORD.

### DIFF
--- a/editor/js/reaction.js
+++ b/editor/js/reaction.js
@@ -37,7 +37,8 @@ exports = {
   getSelector,
   getSelectorText,
   setOptionalBool,
-  getOptionalBool
+  getOptionalBool,
+  freeze
 };
 
 goog.require('ord.conditions');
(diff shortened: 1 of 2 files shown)


editor/js/reaction.js
```diff
@@ -906,3 +907,23 @@ function enumToStrings(protoEnum) {
 function stringToEnum(name, protoEnum) {
   return protoEnum[name];
 }
+
+/**
+ * Switch the UI into a read-only mode. This is irreversible.
+ */
+function freeze() {
```
skearnes, line 914: How do you expect this will be called? Will the editor be loaded in a mutable state and then frozen? Ideally the user would never see the unfrozen state.

antonkast-google: I figured whoever is sorting out the search interface would just call freeze(). Does that make sense?

Before freeze: https://photos.app.goo.gl/aBsTius8ptqxK79k9

After freeze: https://photos.app.goo.gl/fjZZwV3fPrvnXwNi9

skearnes: Right, I'm wondering though if the page will load in a mutable state first? How quickly does the transition happen? Is it possible to load the form the first time in frozen mode?

skearnes (approved):

editor/js/reaction.js
antonkast-google: It used to be that the editor initialized the UI in a single JS task via the XHR response handler for the reaction request. Under that design, a call to freeze() at the end of that handler would set up the frozen DOM without any intermediate state getting rendered on the screen.

On another GitHub issue however, I recently read the troubling news that we have an initialization race. Even worse, we are solving it with a sleep.

If I understood that issue correctly, then I think all bets are off with regard to displaying intermediate states, and I would solve the problem you asked about by first understanding our initialization.

skearnes: Thanks for bringing up that issue; I'll merge this while we work on that.

PR #366, opened 2020-09-16
Draw compounds only after fully loaded

### LINKED ISSUE #344: Editor does not reload compound drawings
When I click a previously defined reaction on the Dataset page, the reaction loads without rendering the images for the compounds I drew with Ketcher (or that have valid structural identifiers, to be more precise).

n8kim1:
Seems to fix #344; would like testing! This fixes by _creating_ the image node entirely after the compound itself is loaded and we get the data of the image's png.

If this seems good, I'll clean it up.

Having an empty `src` in the nodes was problematic -- I think that sometimes, we would set `src` midway during the image loading. It wouldn't load with the new `src`, nor would it refresh.

### DIFF
--- a/editor/html/reaction.html
+++ b/editor/html/reaction.html
@@ -195,9 +195,7 @@ limitations under the License.
                     </span>
                   </legend>
                   <!-- Compound rendering -->
-                  <div style="text-align: center">
-                    <img class="component_rendering" src="" style="margin-left: auto; margin-right: auto;"/>
-                  </div>
+                  <div class="image_container" style="text-align: center"></div>
                   <div class="identifiers"></div>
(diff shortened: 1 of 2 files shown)


skearnes:
Note that #364 gets rid of the empty `src`.

n8kim1:
Neat -- does that fix the issue too?

I'll keep this PR open for now, if that's ok? Can take another look once #364 is done.

skearnes:
> Neat -- does that fix the issue too?
> 
> I'll keep this PR open for now, if that's ok? Can take another look once #364 is done.

Sounds good; #364 is in so you're good to go...

n8kim1:
Closed by #344 being closed; switching to SVG's also incorporated these changes.

Issue #377, opened 2020-09-17
Ensure consistent behavior of floattext/integertext validation


connorcoley:
Alpha tester noticed an abundance of red boxes around input fields, indicating invalid floats/integers, for what appeared to be reasonable inputs. We need to do more testing of these validation functions to ensure that they behave correctly.

PR #380, opened 2020-09-19
Allow whitespace at beginning/end of numeric fields

connorcoley:
Might help address #377 - achieving consistent behavior with numeric checks.


skearnes:
If we move to `<input type="number"` these problems should go away...but I don't feel strongly about starting with this more localized solution.

PR #392, opened 2020-09-21
Round floating point values for display

### LINKED ISSUE #229: Floating point errors when loading/saving in editor
**Describe the bug**
In the editor, floating-point errors occur when loading or saving a float value.

**To Reproduce**
Loading a float value:
1. Load a reaction in the editor
2. Set a value to 0.1
3. Save the reaction (note that it's correct in the pbtxt file)
4. Load the reaction again
5. See the value textbox be filled with a float that's not exactly equal to 0.1

Saving a float value:
1. Load a reaction in the editor
2. Set a value to 0.1212121212
3. Save the reaction (note that it's truncated in the pbtxt file)
4. Load the reaction again
5. See the value textbox be filled with a float that's not exactly the original value nor truncated value

**Expected behavior**
While the loss of precision isn't an issue, the extra digits in the editor's value may be jarring.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: Chromium 83, JS 8.3, Python 3.7.6 (Conda 4.8.3)

**Some thoughts**
Various conversions occur in a few places. Strings of digits are converted to floats when a protobuf message is created, both during loading (in the backend) and saving (in the frontend); I assume errors can't be avoided here. On the other hand, when the backend writes to the text file, only 8 digits are saved -- a setting we could change. (see [backend code](github.com/Open-Reaction-Database/ord-schema/blob/main/editor/py/serve.py#L326), [`float_format` arg in protobuf API docs](googleapis.dev/python/protobuf/latest/google/protobuf/text_format.html#google.protobuf.text_format.MessageToString) )

If this issue is worth fixing, what would be best? 
I'm thinking that when the frontend loads, it could round every float (for example to 8 digits, to match with the backend). Or, when the backend saves a reaction, it could write more digits than just 8.

@antonkast-google @skearnes @connorcoley

skearnes:
Fixes #229 

* Rounds floating point values loaded into the form from protocol buffers to 7 decimal digits (matching 32-bit float precision).
* Rounds float values similarity in python when formatting "units" messages (affects the HTML reaction table).

### DIFF
--- a/editor/js/reaction.js
+++ b/editor/js/reaction.js
@@ -192,14 +192,15 @@ function checkInteger(node) {
  * @return {number}
  */
 function prepareFloat(value) {
-  // Round to 15 decimal places; this avoid floating point precision issues
+  // Round to 15 significant digits; this avoid floating point precision issues
   // that can be quite jarring to users.
   //
   // See:
   //   * https://stackoverflow.com/a/3644302
   //   * https://medium.com/swlh/ed74c471c1b8
-  const factor = Math.pow(10, 15);
-  return Math.round(value.toPrecision(15) * factor) / factor;
+  const precision = 15;
+  const factor = Math.pow(10, precision);
+  return Math.round(value.toPrecision(precision) * factor) / factor;
 }
 
 /**
--- a/editor/js/reaction.js
+++ b/editor/js/reaction.js
@@ -450,7 +450,7 @@ function loadReaction(reaction) {
   $('.floattext').each(function(index) {
     const node = $(this);
     if (node.text() !== '') {
-      node.text(prepareFloat(node.text()));
+      node.text(prepareFloat(parseFloat(node.text())));
     }
   });
 }
--- a/editor/js/reaction.js
+++ b/editor/js/reaction.js
@@ -187,6 +187,22 @@ function checkInteger(node) {
 }
 
 /**
+ * Prepares a floating point value for display in the form.
+ * @param {number} value
+ * @return {number}
+ */
+function prepareFloat(value) {
+  // Round to 7 decimal places; this avoid floating point precision issues
+  // that can be quite jarring to users.
+  //
+  // See:
+  //   * https://stackoverflow.com/a/3644302
+  //   * https://medium.com/swlh/ed74c471c1b8
+  const factor = Math.pow(10, 7);
+  return Math.round(value * factor) / factor;
+}
+
+/**
  * Adds a jQuery handler to a node such that the handler is run once whenever
  * data entry within that node is changed, *except through remove* -- this must
  * be handled manually. (This prevents inconsistent timing in the ordering of
@@ -428,6 +444,14 @@ function loadReaction(reaction) {
     ord.provenance.load(provenance);
   }
   $('#reaction_id').text(reaction.getReactionId());
+
+  // Clean up floating point entries.
+  $('.floattext').each(function(index) {
+    const node = $(this);
+    if (node.text() !== '') {
+      node.text(prepareFloat(node.text()));
+    }
+  });
 }
 
 /**
--- a/editor/js/reaction.js
+++ b/editor/js/reaction.js
@@ -192,15 +192,15 @@ function checkInteger(node) {
  * @return {number}
  */
 function prepareFloat(value) {
-  // Round to 15 significant digits; this avoid floating point precision issues
+  // Round to N significant digits; this avoid floating point precision issues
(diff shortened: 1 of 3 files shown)


connorcoley:
This isn't quite right IMO. It could be a fine patch for this issue, but we should really have a more sophisticated way of doing this based on the number of digits in the significand, regardless of the exponent.

For instance, you can put in "0.00000000000001" and it will be stored as "1e-14". It will then be read into the form as "9.9999998245167e-15" if you roundtrip it to a pbtxt. The pbtxt doesn't always store things to 7 decimal places.

This also raises a new issue for my hacky float/integer check (#377), since we can now have exponents in the field denoted by the letter e.

skearnes:
> This isn't quite right IMO. It could be a fine patch for this issue, but we should really have a more sophisticated way of doing this based on the number of digits in the significand, regardless of the exponent.
> 
> For instance, you can put in "0.00000000000001" and it will be stored as "1e-14". It will then be read into the form as "9.9999998245167e-15" if you roundtrip it to a pbtxt. The pbtxt doesn't always store things to 7 decimal places.
> 
> This also raises a new issue for my hacky float/integer check (#377), since we can now have exponents in the field denoted by the letter e.

Hmm...the proto definition uses `float`, which only supports 32-bit floats, so "0.00000000000001" is technically invalid. However, as you mention, `text_format` does the right thing on serialization (`value: 1e-14`) but loads the borked version when you deserialize. This could be fixed by treating these as `double`s when rounding (more precision) but that will break the fix for things with fewer decimal places.

So I see a few options here:
1. Disallow values with more precision than the underlying protocol buffer definition (7 decimal places for floats).
1. Treat all floating point values as `double`s and go out to 15 places for rounding.
1. Something more sophisticated like trying both roundings and seeing which one gives the cleanest value...

FWIW I think option 1 is probably best.

skearnes:
PS `<input type="number"...` allows 'e' but not other non-numeric characters.

skearnes:
I can't think of a way to do this dynamically based on the number of significant digits, since we'll lose that information when we try to round trip (it comes into the loaded protocol buffer incorrect). Am I missing something?

skearnes:
On second thought, I think I agree with you about significant digits. Let me dig deeper.

skearnes:
It looks like javascript's `toPrecision` method should be part of the answer. Possibly coupled with rounding...

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Number/toPrecision

skearnes:
PTAL at the latest commits that use up to 15 significant digits.

skearnes:
So I got things working for 1e-14, but broke 0.04 since it doesn't round to 0.04 with 15 digits of precision.

skearnes:
Oh I think I figured it out; just use `parseFloat` instead of `Math.round`, in combination with `toPrecision`.

connorcoley:
> Am I missing something?

I don't think so. I've had a surprisingly hard time trying to reason my way through possible solutions to this problem. It doesn't seem like it should be that hard.

skearnes:
image

and after reload:

image

connorcoley:
That behavior looks good to me. There's no red outline in your example, but if you change the field and trigger the checkFloat function, it will complain and then not let you successfully validate the reaction. How should we change the test? Just look for round-trip consistency w/ prepareFloat(parseFloat(node.text())) ?

skearnes:
> That behavior looks good to me. There's no red outline in your example, but if you change the field and trigger the checkFloat function, it will complain and then not let you successfully validate the reaction. How should we change the test? Just look for round-trip consistency w/ prepareFloat(parseFloat(node.text())) ?

The test still passes as-is, since we never call `checkFloat`, but I agree that is brittle. I suggest we fix `checkFloat` to allow for scientific notation (my thought is to make it recursive and split on `'e'`) or move to `<input type="number"...` for these fields.

If it's ok with you I'd prefer to check this in separately from fixing this issue since it's currently passing.

connorcoley:
I was going to suggest the opposite and say we should introduce a test that triggers a change/blur event for all fields, re-validates a known-to-be-valid reaction, and ensures there are no validation errors. 

I don't think we should merge this one in w/o fixing the checkFloat issue, since we would have a situation where a user would be unable to record a reaction exactly as the form is trying to display it. For the purposes of this PR, I might suggest you just disable checkFloat and checkInteger temporarily. We can re-enable them when we fix the exponent issue.

connorcoley (approved):

PR #395, opened 2020-09-22
Allow scientific notation

skearnes:
Supercedes #380; also related to #377

### DIFF
--- a/editor/js/reaction.js
+++ b/editor/js/reaction.js
@@ -62,6 +62,9 @@ const session = {
 // Export session, because it's used by test.js.
 exports.session = session;
 
+const FLOAT_PATTERN = /^-?\d+(?:\.\d+)?(?:[eE]-?\d+)?$/;
+const INTEGER_PATTERN = /^-?\d+$/;
+
 /**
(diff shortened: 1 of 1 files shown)


editor/js/reaction.js
```diff
@@ -62,6 +62,9 @@ const session = {
 // Export session, because it's used by test.js.
 exports.session = session;
 
+const FLOAT_PATTERN = /^-?\d+(?:\.\d+)?(?:[eE]-?\d+)?$/;
```
connorcoley, line 65: What about `^-?\d*(?:\.\d+)?(?:[eE]-?\d+)?$` to allow decimals w/o a leading 0? e.g., ".5" or "-.2"

connorcoley (approved):

editor/js/reaction.js
skearnes: Hmm...that would also allow things like `e-14`. Expanded a little to make those cases mutually exclusive; PTAL.

connorcoley: LGTM

PR #397, opened 2020-09-23
Create a base image for faster docker building/testing

skearnes:
Drastically improves GitHub build times for the editor tests and standardizes on recommended Dockerfile practices for managing image size, etc.


skearnes:
After sleeping on this, I don't actually think it's a good idea. It adds more complexity to the pipeline for minimal benefit. I'll create a different PR with the changes to reduce image size.

PR #398, opened 2020-09-23
Add nanoliter volume unit

### LINKED ISSUE #387: Add "nanoliter" option for volume fields

connorcoley:
Will close #387

### DIFF
--- a/ord_schema/proto/reaction.proto
+++ b/ord_schema/proto/reaction.proto
@@ -977,10 +977,10 @@ message Time {
 message Mass {
   enum MassUnit {
     UNSPECIFIED = 0;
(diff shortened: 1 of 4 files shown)


ord_schema/proto/reaction.proto
```diff
@@ -1006,6 +1006,7 @@ message Volume {
     MILLILITER = 1;
     MICROLITER = 2;
     LITER = 3;
+    NANOLITER = 4;
```
skearnes, line 1009: Feel free to move this up in the list so it shows up in a logical order, just keep the current tag number.

skearnes (approved):

PR #400, opened 2020-09-23
Catch ValueErrors raised when trying to render molecules

### LINKED ISSUE #383: Reaction summary is not drawn if a component is missing a structural identifier
The lack of a valid structural identifier should not prevent the entire reaction scheme from being drawn.

connorcoley:
Fixes #383

Bonus typo in the script version due to renamed argument

Let me know if you have strong feelings about a placeholder other than "[Component]"

### DIFF
--- a/ord_schema/scripts/generate_text.py
+++ b/ord_schema/scripts/generate_text.py
@@ -54,5 +54,5 @@ def main(argv):
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


skearnes (approved):
Maybe '[Compound]' instead, since the ketcher button says "draw compound" and this matches the underlying proto?

PR #402, opened 2020-09-23
Editor test expansion - simulate changing text fields

connorcoley:
This PR adds a new set of test (right now, a single test using a valid reaction) where we compare the number of validation errors before/after "changing" fields by triggering `$('.edittext').trigger('blur');`. We don't expect the number of validation errors to change in this case.

### DIFF
--- a/editor/js/test.js
+++ b/editor/js/test.js
@@ -78,10 +78,12 @@ const puppeteer = require('puppeteer');
     await page.waitFor('body[ready=true]');
     const test = page.evaluate(async function(url) {
       ord.reaction.validateReaction();
-      const prevErrors = parseInt($('.validate_status', '#reaction_validate').html());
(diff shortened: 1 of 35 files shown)


editor/js/test.js
```diff
@@ -30,22 +30,22 @@ const puppeteer = require('puppeteer');
   page.on('console', msg => console.log('console>', msg.text()));
 
   // Round-trip these reactions through the DOM and compare at the server.
-  const tests = [
+  var tests = [
```
skearnes, line 33: Why not leave this as const and then create a separate validationTests (also const) array below?

skearnes (approved):

editor/js/test.js
connorcoley: Fixed

Issue #404, opened 2020-09-23
UX improvements for the editor


skearnes:
Comments and suggestions from Devin:
* There's no obvious type hierarchy; adding one would make navigation clearer.
* The box outlines create a lot of visual noise; there's so much of it that the only thing you can do is ignore all of it.
* Navigation via the sidebar is awkward:
  * Oddly positioned
  * Doesn't show where you are
  * Only shows the top level (it should include e.g. named inputs as sublevels)
  * Hides content rather than navigating within the overall document structure (and it's not intuitive to un-hide content)
* Style
  * Consider a header that includes document-level information and actions a la Google Docs
  * Consider adding a "summary" tab in the sidebar for getting to the reaction image
  * Consider making the lines clickable to collapse a la Reddit
  * Group actions together where possible (e.g. don't have the input "remove" button way off to the right when it could be next to "validate")
  * Make button language consistent: "draw compound", "look up name", and "add identifier" are all different ways to do the same thing
  * Consider labeling the identifiers (e.g. 1, 2, 3) so there is a natural place to put the remove button

Suggested priorities:
1. Sidebar
1. Type hierarchy
1. Boxes + action grouping
1. Header
1. Other styling e.g. with bootstrap

skearnes:
![image](https://user-images.githubusercontent.com/952729/94045372-9aa56200-fd8c-11ea-8721-b786a1d24a60.png)

PR #406, opened 2020-09-25
Updates to the editor navigation bar

skearnes:
Related to #404 

* Changes sidebar behavior to scroll to the selected section rather than show/hide.
* Adds support for updating the sidebar selection as the user scrolls.

NOTE: The dynamic sidebar selection is quite complicated---how do you decide which section you're "in"? As a way of sidestepping this issue I chose to highlight the names of any sections that are currently visible:

image

### DIFF
--- a/editor/html/reaction.html
+++ b/editor/html/reaction.html
@@ -1083,21 +1083,15 @@ limitations under the License.
         navSelectors[section] = selector;
       });
 
-      // Show only the given section of the reaction schema.
+      // Navigate to the given section of the reaction schema.
       function setSelected(section) {
         $('.navSection').css('background-color', '');
-        $('.section').hide('slow');
-        if (section != null) {
-          $('#section_' + section).show('slow');
-          const selector = navSelectors[section];
-          selector.css('background-color', 'lightblue');
-        } else {
-          $('.section').show('slow')
-        }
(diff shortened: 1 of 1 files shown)


skearnes:
Update: I played with this some more and have a mostly functional single-selection sidebar working. It jumps around a bit when scrolling by the smaller sections but generally behaves as you'd expect.

connorcoley:
I like the scrolling to the position but I do find the behavior of the sidebar a little unintuitive now. It seems to select whichever section has most recently come into frame. If you try to navigate to Notes after being further down, for example, you end up with "Notes" selected. If you navigate to Notes after being further up, you can end up with Outcomes selected depending on the size of your window and display resolution. Notes will be at the top of the page though, which is good

skearnes:
Yeah it's going to be tricky to get that right, I think. What do you think of the multi-selection option from the original description?

connorcoley:
I don't think the multi-selection is a bad idea necessary...it would highlight anything that's visible on the page. It will be pretty useful when we have individual inputs listed under Inputs

skearnes:
OK, I reverted to the original multi-selection proposal. We can fix single-selection later if we want. PTAL.

connorcoley (approved):

PR #418, opened 2020-09-30
Improve read-only editor access

### LINKED ISSUE #417: Multi-line or long text entries in boxes are not displayed properly when loaded in the editor
If the length of a string value exceeds the default width of the field, the text spills over. Fields will automatically resize themselves when actively typing a long string, but the behavior upon loading is wonky:

image

skearnes:
* Adds a "freeze" function to the database view.
* Adjusts CSS to make read-only mode more readable; fixes #417 
* Automatically enters read-only mode for reviewing submissions

### DIFF
--- a/editor/css/reaction.css
+++ b/editor/css/reaction.css
@@ -36,7 +36,7 @@ body {
   min-height: 20px;
   border: 1px solid #c0c0c0;
   white-space: pre-wrap;
-  background-color: white;
+  background-color: #EBEBE4;
 }
 .shorttext {
   width: 75px;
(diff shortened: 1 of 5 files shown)


skearnes:
FYI updated background color; I went with #ebebe4, which has a bit of yellow in it...

connorcoley:
I don't think you meant to add this styling in `reaction.css`. The background color for the edittext CSS styling should be white (or undefined). I think this is a good color choice for a `$('.edittext').css('background-coior', '#ebebe4');` in `ord.reaction.freeze`

skearnes:
haha fixed :)

connorcoley (approved):

Issue #421, opened 2020-09-30
Pass validation warnings through to the editor


connorcoley:
Currently, the editor only reports validation errors. I think it would be useful to see validation _warnings_ as well (that's why we have them in the first place!)

PR #428, opened 2020-10-01
Relax percentage values

### LINKED ISSUE #422: Relax float range for Percentage fields
Validation does not allow Percentages to be less than 0 or greater than 105. However, any value is possible if the precision is low enough. For example, approximately quantified yields from HTE can easily exceed 105% and a negative conversion is also possible given experimental error. The restriction on values should either be removed or changed to a warning.

skearnes:
Fixes #422 by issuing a warning for values outside the expected [0, 100] range.

The tests are unsatisfying since they don't capture warnings; I'll fix this when addressing #421


skearnes:
Gentle ping in case you missed this one...

connorcoley (approved):

PR #429, opened 2020-10-02
Display validation errors in the editor

### LINKED ISSUE #421: Pass validation warnings through to the editor

skearnes:
Fixes #421 

image


connorcoley (approved):
Looks like a nice solution, thanks!

PR #434, opened 2020-10-03
Allow selectors to be out of order (fixes #433)

### LINKED ISSUE #433: Web editor assumes that ENUMs will be listed in reaction.proto in order
After trying to expand the workup steps to include ALIQUOT, I noticed a validation error relating to missing information for a TEMPERATURE step. The web editor's selectors will list options in the order they appear in the proto, but does not pay attention to the actual field ID assigned in the proto definition.

i.e., the workup type list
```
UNSPECIFIED = 0;
CUSTOM = 1;
ADDITION = 2;
ALIQUOT = 18;
TEMPERATURE = 3;
```
will result in a selector showing UNSPECIFIED, CUSTOM, ADDITION, ALIQUOT, TEMPERATURE, but selecting ALIQUOT is treated as a value of 3 and resolved to TEMPERATURE in the back end.

This is likely already causing bugs because the volume enum is not in order.

connorcoley:


skearnes (approved):
Any chance this affects existing data?

connorcoley:
I don't believe so, no. It's not an issue for programmatic definition, and I when reviewing templated enumerations, I looked at the pbtxt in a text editor, not as loaded onto the form

PR #435, opened 2020-10-03
Aliquot workup step

### LINKED ISSUE #424: Add new ALIQUOT workup step
Add workup step of `ALIQUOT` for the common case where a sample of the crude product is taken for dilution and injection onto a column. Allow an "amount" of Mass or Volume. Details like "into HPLC vial" will be kept as unstructured text in the details field. Because workup steps are linear sequences of actions, we assume that the aliquoted volume becomes the material that we're following in subsequent steps.

connorcoley:
Resolves #424 by adding a mass/volume amount field to the workup message to be used for ALIQUOT-type steps. Adds corresponding fields to the editor, which requires lots of additional js to get the selectors correct. Addressed by basing implementation on the crude component amount fields, but long-term might want to consolidate all of the amount selector logic to avoid repeated code.

### DIFF
--- a/editor/db/test/full.pbtxt
+++ b/editor/db/test/full.pbtxt
@@ -371,6 +371,11 @@ reactions {
     }
     target_ph: 89.0
     is_automated: true
+    volume {
+      value: 123.0
+      precision: 0.5
+      units: MILLILITER
+    }
   }
   outcomes {
     reaction_time {
--- a/editor/html/reaction.html
+++ b/editor/html/reaction.html
@@ -284,12 +284,14 @@ limitations under the License.
                     <span class="collapse"></span>
                     Amount
                   </legend>
-                  <input type="radio" class="crude_amount_mass" value="mass" checked> mass
-                  <input type="radio" class="crude_amount_volume" value="volume"> volume
-                  <span style="padding-left:25px">value</span> <div class="crude_amount_value edittext shorttext floattext"></div>
-                  +/- <div class="crude_amount_precision edittext shorttext floattext"></div>
-                  <div class="crude_amount_units_mass selector" data-proto="Mass_MassUnit"></div>
-                  <div class="crude_amount_units_volume selector" data-proto="Volume_VolumeUnit" style="display: none;"></div>
+                  <div>
+                    <input type="radio" class="crude_amount_mass" value="mass" checked> mass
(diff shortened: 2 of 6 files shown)


editor/html/reaction.html
```diff
@@ -284,12 +284,14 @@
                     <span class="collapse"></span>
                     Amount
                   </legend>
-                  <input type="radio" class="crude_amount_mass" value="mass" checked> mass
-                  <input type="radio" class="crude_amount_volume" value="volume"> volume
-                  <span style="padding-left:25px">value</span> <div class="crude_amount_value edittext shorttext floattext"></div>
-                  +/- <div class="crude_amount_precision edittext shorttext floattext"></div>
-                  <div class="crude_amount_units_mass selector" data-proto="Mass_MassUnit"></div>
-                  <div class="crude_amount_units_volume selector" data-proto="Volume_VolumeUnit" style="display: none;"></div>
+                  <div>
```
connorcoley, line 287: Note: wrapping the contents of a collapsible fieldset is necessary to prevent the contents from getting weird and out of order when you collapse/uncollapse. This maybe should have been a separate PR bugfix

editor/db/test/full.pbtxt
```diff
@@ -371,6 +371,11 @@ reactions {
     }
     target_ph: 89.0
     is_automated: true
+    volume {
```
skearnes, line 374: The proto doc says these amounts are only for aliquots---should this trigger a validation error?

editor/html/reaction.html
skearnes: SGTM

skearnes (commented):
Can you create an issue for the consolidation you suggested in the PR description?

editor/db/test/full.pbtxt
connorcoley: I wasn't sure how strict we wanted to be in workup validations. Right now, we only ensure that if the step is type X, fields {Y} are defined. We don't ensure that irrelevant fields are _not_ defined for any other types

skearnes: OK, maybe something to think about for future PRs, especially if a field is irrelevant for certain types.

skearnes (approved):

PR #440, opened 2020-10-05
Show inputs in the sidebar

skearnes:
Related to #404 

Note that getting the in-view/selected input to highlight will involve some extra work resetting the `InteractionObserver` etc.

image

### DIFF
--- a/editor/css/reaction.css
+++ b/editor/css/reaction.css
@@ -19,7 +19,7 @@ body {
   font-family: Roboto;
 }
 #sections {
-  margin-left: 128px;
+  margin-left: 150px;
 }
 #top_buttons {
   text-align: center;
@@ -131,14 +131,20 @@ fieldset {
   margin-top: 10px;
   margin-right: 26px;
   position: fixed;
-  width: 110px;
+  width: 132px;
 }
 .navSection {
   padding: 8px;
   margin-bottom: 4px;
   cursor: pointer;
 }
-.navSection:hover {
+.inputNavSection {
+  padding: 8px 8px 8px 16px;
+  margin-bottom: 4px;
+  cursor: pointer;
+  font-size: 90%;
+}
+.navSection:hover, .inputNavSection:hover {
   background-color: lightblue;
 }
 .data_uploader .adder {
--- a/editor/css/reaction.css
+++ b/editor/css/reaction.css
@@ -139,7 +139,7 @@ fieldset {
   cursor: pointer;
 }
 .inputNavSection {
-  padding: 8px 8px 8px 16px;
+  padding: 8px;
   margin-bottom: 4px;
   cursor: pointer;
   font-size: 90%;
--- a/editor/html/reaction.html
+++ b/editor/html/reaction.html
@@ -104,7 +104,7 @@ limitations under the License.
                   <span class="validate" onclick="ord.inputs.validateInput($(this).closest('.input'))"></span>
                </legend>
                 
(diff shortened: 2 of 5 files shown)


skearnes:
I can probably just create a separate `InteractionObserver` for each input; that should make things much easier...I'll still do this in a different PR.

connorcoley:
Few things:
- On load (e.g., of the full reaction pbtxt example), the sidebar contains a single "Input 0" rather than what the loaded reaction contains (e.g., the two inputs with names "4" and "")
- Removing an input does not trigger the sidebar to be updated
- Should we only call `updateSidebar` when an input is added/removed or has its name changed? Rather than with any change within the whole input node? 

I do think resetting the InteractionObserver will be a nice feature. Note that the section highlighting doesn't take into account the portion of the form occluded by the header, so when I click on "Inputs", "Identifiers" is still also selected. You might consider including that in the same issue for adding input-level highlighting

skearnes:
> Few things:
> 
> * On load (e.g., of the full reaction pbtxt example), the sidebar contains a single "Input 0" rather than what the loaded reaction contains (e.g., the two inputs with names "4" and "")

Good catch; this wasn't happening on the Nielsen test data. The problem seems to be that `updateSidebar` was called only once sometime during the form initialization. Seems to be fixed now.

> * Removing an input does not trigger the sidebar to be updated

Fixed.

> * Should we only call `updateSidebar` when an input is added/removed or has its name changed? Rather than with any change within the whole input node?

Good idea. Done.
 
> I do think resetting the InteractionObserver will be a nice feature. Note that the section highlighting doesn't take into account the portion of the form occluded by the header, so when I click on "Inputs", "Identifiers" is still also selected. You might consider including that in the same issue for adding input-level highlighting

Will do.

connorcoley:
More comments:
- There is now an error being printed to the console when loading a reaction (e.g., the full example). This might be unrelated to the current PR, but it's happening in the sidebar when calling `clearSelected()` on undefined.
- I can no longer remove an input: `Uncaught TypeError: Cannot read property 'remove' of undefined`. I think you forgot to add `ord.inputs.remove` to the `exports`? 
- Given that you're calling `ord.inputs.updateSidebar();` from workups.js, I think you should keep it in `ord.reaction` since that is the catch-all for any form-wide js functions. As a related question, why do you need to update the sidebar after adding a workup input node?

skearnes:
> More comments:
> 
> * There is now an error being printed to the console when loading a reaction (e.g., the full example). This might be unrelated to the current PR, but it's happening in the sidebar when calling `clearSelected()` on undefined.

I noticed that too. Some quick attempts to fix it failed miserably; I think it's unrelated to this PR.

> * I can no longer remove an input: `Uncaught TypeError: Cannot read property 'remove' of undefined`. I think you forgot to add `ord.inputs.remove` to the `exports`?

Fixed.

> * Given that you're calling `ord.inputs.updateSidebar();` from workups.js, I think you should keep it in `ord.reaction` since that is the catch-all for any form-wide js functions. As a related question, why do you need to update the sidebar after adding a workup input node?

This is because `updateSidebar` grabs everything with the `input` class, including a newly added workup input. The extra call in workups.js is to remove workup inputs from the list (since `updateSidebar` excludes nodes with the `workup_input` class). I think I want to keep this function in inputs.js, so I've updated `ord.inputs.add` to allow the `workup_input` class to be added on creation of the new node and avoids having to call `updateSidebar` twice. (Note that there are other calls to `ord.inputs.*` functions in workups.js already.)

connorcoley (approved):

PR #451, opened 2020-10-09
Updates for latest closure linter warnings

skearnes:


antonkast-google:
Have you looked into the IE compatibility? Since we're not transpiling, my understanding is this will not work on IE 10 or earlier. I don't think that matters.

antonkast-google (approved):

skearnes:
> Have you looked into the IE compatibility? Since we're not transpiling, my understanding is this will not work on IE 10 or earlier. I don't think that matters.

Agreed; I don't think we need to worry about this right now.

PR #452, opened 2020-10-09
Add docker-compose configuration for the editor

skearnes:
Running `docker-compose up` starts two images: one containing a postgres database and the other containing the ORD web editor. The postgres database is exposed to the editor and can be accessed by connecting to `ORD_EDITOR_DB` with `ORD_EDITOR_POSTGRES_PASSWORD`. The postgres container writes to a mounted volume; this ensures that the database persists when restarting the image. On GCP, the mounted volume should be a persistent disk.

Also changes the current `ORD_EDITOR_DB` variable, which refers to the location of the local or mounted filesystem, to `ORD_EDITOR_LOCAL`.

### DIFF
--- a/editor/README.md
+++ b/editor/README.md
@@ -108,7 +108,7 @@ so remember to download your Dataset when you are done.
 
 ## How it Works
 
-Datasets to be edited are stored in the ORD_EDITOR_DB directory as `.pbtxt`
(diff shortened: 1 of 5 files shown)


skearnes:
Note that docker-compose will not be used for deployment on GCP. Instead, we will have a separate VM hosting the database that the frontend VM(s) can connect to. So this PR is really for unblocking local development.

antonkast-google (approved):

PR #455, opened 2020-10-09
Canonicalize SMILES from Ketcher and name resolvers

### LINKED ISSUE #445: Ketcher SMILES format
From Brian Cole: Ick, looks like Ketcher returned a Kekule SMILES :frowning:

skearnes:
Fixes #445


connorcoley (approved):

connorcoley:
We should use the non-canonicalized SMILES from Ketcher when RDKit fails to parse it

PR #456, opened 2020-10-10
Quick input definition by string

### LINKED ISSUE #372: Add a helper for simple solutions
An alpha tester reported that encoding multiple simple solutions (e.g. a single solute) was a bit tedious. We might consider adding a "wizard" that allows a user to specify a solute, solvent, and concentration to fill in the relevant underlying parts of the form.

### LINKED ISSUE #444: Stock solutions from vendors
From Kjell Jorner: Regarding stock solutions, they may be bought from manufacturers and original recipe unknown

(re: not allowing solutions to be defined by concentration in the schema)

connorcoley:
Fixes #444 and fixes #372 

This adds a very simple parsing tool to "quick add" an input. Only two patterns are supported and they are quite rigid:

(1) [AMOUNT] of [NAME]
(2) [AMOUNT] of [CONCENTRATION] [SOLUTE] in [SOLVENT]

Any thoughts on making it more flexible? I'll leave it to Steve if he wants to get fancy with regex. This is at least functional and does seem to add a bit of convenience.

### DIFF
--- a/ord_schema/resolvers.py
+++ b/ord_schema/resolvers.py
@@ -1,0 +1,172 @@
+# Copyright 2020 Open Reaction Database Project Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Automated updates for Reaction messages."""
+
+import datetime
+import uuid
+import re
+import urllib.parse
+import urllib.request
+import urllib.error
+from absl import logging
+
+from ord_schema import message_helpers
+from ord_schema import units
+from ord_schema.proto import reaction_pb2
+
+_COMPOUND_STRUCTURAL_IDENTIFIERS = [
+    reaction_pb2.CompoundIdentifier.SMILES,
+    reaction_pb2.CompoundIdentifier.INCHI,
+    reaction_pb2.CompoundIdentifier.MOLBLOCK,
+    reaction_pb2.CompoundIdentifier.CXSMILES,
+    reaction_pb2.CompoundIdentifier.XYZ,
+]
+
+_USERNAME = 'github-actions'
+_EMAIL = 'github-actions@github.com'
+
+
+def name_resolve(value_type, value):
+    """Resolves compound identifiers to SMILES via multiple APIs."""
+    smiles = None
+    for resolver, resolver_func in _NAME_RESOLVERS.items():
+        try:
+            smiles = resolver_func(value_type, value)
+            if smiles is not None:
+                return smiles, resolver
+        except urllib.error.HTTPError as error:
+            logging.info('%s could not resolve %s %s: %s', resolver, value_type,
+                         value, error)
+    raise ValueError(f'Could not resolve {value_type} {value} to SMILES')
+
+
+def resolve_names(message):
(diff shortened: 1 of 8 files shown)


ord_schema/resolvers.py
```diff
@@ -0,0 +1,188 @@
...
+    amount_string, description = input_string.split(' of ')
+    if ' in ' not in description:
+        component_name = description
+        component = reaction_input.components.add()
+        component.CopyFrom(
+            message_helpers.build_compound(name=component_name.strip(),
+                                           amount=amount_string))
+        resolve_names(reaction_input)
+        return reaction_input
+    pattern = re.compile(r'(\d+.?\d*)\s*(\w+) (.*) in (.*)')
```
skearnes, line 162: This looks ok but I'd probably use `(.+)` instead of `(.*)`. I might also replace `\s*` with `\s?` so you only allow a maximum of one space before the units.

ord_schema/resolvers.py
```diff
@@ -0,0 +1,188 @@
...
+    match = pattern.fullmatch(description.strip())
+    if not match:
+        raise ValueError('String did not match template!')
+    conc_value, conc_units, solute_name, solvent_name = match.groups()
+    solute = reaction_input.components.add()
+    solvent = reaction_input.components.add()
+    solute.CopyFrom(message_helpers.build_compound(\
```
skearnes, line 169: Please remove this line break (`\`)

ord_schema/resolvers_test.py
```diff
@@ -0,0 +1,84 @@
...
+class InputResolversTest(parameterized.TestCase, absltest.TestCase):
+
+    def test_input_resolve(self):
+        string = '10 g of THF'
+        reaction_input = resolvers.resolve_input(string)
+        self.assertEqual(reaction_input.components[0].mass,
+                         reaction_pb2.Mass(value=10, units='GRAM'))
+        self.assertEqual(
+            reaction_input.components[0].identifiers[0],
+            reaction_pb2.CompoundIdentifier(type='NAME', value='THF'))
```
skearnes, line 52: Also check the SMILES?

ord_schema/resolvers_test.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assertEqual(len(reaction_input.components), 1)
+
+        string = '100 mL of 5.0uM sodium hydroxide  in water'
```
skearnes, line 55: Hmm...this technically breaks the template since there are two spaces before "in". Do you want to allow this? If not, consider using a character class like `[a-zA-Z0-9\.]` instead of `.` in the regex.

ord_schema/resolvers_test.py
```diff
@@ -0,0 +1,84 @@
...
+        reaction_input = resolvers.resolve_input(string)
+        self.assertEqual(len(reaction_input.components), 2)
+        self.assertEqual(reaction_input.components[0].moles,
+                         reaction_pb2.Moles(value=500, units='NANOMOLE'))
+        self.assertEqual(
+            reaction_input.components[0].identifiers[0],
+            reaction_pb2.CompoundIdentifier(type='NAME',
```
skearnes, line 62: Check the SMILES?

skearnes (approved):
Looks handy!

ord_schema/resolvers_test.py
connorcoley: Done, checking type and value of the identifier

ord_schema/resolvers.py
connorcoley: Done

ord_schema/resolvers.py
connorcoley: Done

ord_schema/resolvers_test.py
connorcoley: I don't necessarily mind extra spaces. I've removed it from the test case though.

ord_schema/resolvers_test.py
connorcoley: Done

PR #462, opened 2020-10-11
Migrate editor from files to Postgres.

antonkast-google:
- Rewrite serve.py for a Postgres backend. Python and JS tests pass unchanged.

- Add scripts to initialize the schema, define default users, and migrate legacy data.

- Integrate the editor processes with Docker Compose, now required by tests.

- Move access tokens from URLs to cookies in preparation for authentication. Add a /login page to switch among users.

- Simplify review mode, replacing git subprocesses with a "sync" operation.

- Fix a bug in dataset upload where JS made incorrect assumptions about pbtxt file encodings.


editor/py/serve.py
```diff
@@ -490,72 +503,56 @@ def get_molfile():
...
-    repo = client.get_repo('Open-Reaction-Database/ord-data')
-    pull_requests = {}
-    for pull_request in repo.get_pulls():
-        datasets = []
-        for file in pull_request.get_files():
-            if file.filename.endswith('.pbtxt'):
-                datasets.append(file.filename[:-6])
-        pull_requests[pull_request.number] = (pull_request.title, datasets)
+    """For the review user only, render datasets with GitHub metadata."""
+    if flask.g.user_id != REVIEWER:
```
skearnes, line 509: Maybe navigating to /review should automatically log in as the review user? Requiring the extra login step when the intention is clear seems inappropriate. However, this does give users a chance to record their user ID before logging out so they don't lose access to their work.

I think what I want here is for users to be able to enter the review portal without having to log out. Thinking ahead, when we switch to GitHub authentication we can't require users to log into a reviewer account; instead, we just want to show them the pending submissions in read-only mode and let them easily get back to their personal datasets.

I'm happy to punt on this and allow it to be addressed in a follow-up PR.

editor/py/serve.py
```diff
@@ -568,29 +565,39 @@ def prevent_caching(response):
...
+    dataset = dataset_pb2.Dataset()
+    text_format.Parse(pbtxt, dataset)
+    return dataset
+
+
+def get_pbtxt(name):
+    """Reads a pbtxt proto from the datasets."""
+    with flask.g.db.cursor() as cursor:
+        query = psycopg2.sql.SQL(
+            'SELECT pbtxt FROM datasets WHERE user_id=%s AND dataset_name=%s')
```
skearnes, line 580: Are we storing this as bytes now? That will avoid some of the encoding/decoding round trips.

editor/py/serve.py
```diff
@@ -662,85 +668,118 @@ def get_file(path):
...
+        timestamp = int(time.time())
+        cursor.execute(query, [access_token, user_id, timestamp])
+        flask.g.db.commit()
+        response = flask.redirect('/')
+        # Expires in a year.
+        response.set_cookie('Access-Token', access_token, max_age=31536000)
+        return response
+
+
+def make_user(name='auto'):
```
skearnes, line 739: Please add a more complete docstring with Args

https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

skearnes (approved):
Thanks Anton, this looks great and it will be a nice improvement to the backend.

There's enough changing here that I think we had better merge this into a new branch of `main` until we have more time to battle test it. I'll create the branch and change the base of the PR.

antonkast-google:
Agreed that this should get more manual testing. Anyone can hit it now at http://34.123.154.187/.

editor/py/serve.py
antonkast-google: Agreed there are UX and permission design questions here. Let's follow up.

editor/py/serve.py
antonkast-google: Nope, we are storing strings. I agree we should store protos in binary encoding. I expect user-perceptible latency improvements.

editor/py/serve.py
antonkast-google: Done.

PR #463, opened 2020-10-11
Log dataset revisions and render them for download.

antonkast-google:
Make it possible to recover from unintended deletions.

- A new table, "revisions" logs datasets with timestamps.

- A new JS behavior logs a checkpoint before taking action on any "remove" button.

- A new endpoint, "/checkpoint" appends to the revisions log.

- A new web page, "/revisions" shows this log for a given dataset, each with a timestamp linking to the historical pbtxt.

It's not exactly an undo feature, but it makes recovery possible from some kinds of mistake. The procedure after a loss is:  go to /revisions; pick a log entry; download it; and upload it again.

This could be integrated into the editor with little effort by adding an editor affordance showing the revision list and linking from there to a Flask function that reverts the dataset and reloads the page.


skearnes:
Forked as https://github.com/Open-Reaction-Database/ord-editor/pull/8

PR #468, opened 2020-10-13
Add support for missing values during templating

### LINKED ISSUE #375: Support reaction templating when some reaction instances are missing a component
Reaction templating allows simple string replacements for component identifiers, which is great when enumerating many reactions with, e.g., different ligands. However, a "no ligand" condition cannot be included in the enumeration. We should consider how to support this somewhat common case. Currently, the best workflow involves defining two separate templates, enumerating two separate datasets, and merging them during submission.

skearnes:
Fixes #375 

It's very focused on missing data for placeholders in `input` messages as opposed to other places in templates.


ord_schema/templating.py
```diff
@@ -63,20 +63,65 @@ def _escape(string):
...
+        ValueError: If the substituted reaction template cannot be parsed.
     """
-    pattern = re.compile('|'.join(map(re.escape, substitutions.keys())))
-    return pattern.sub(lambda match: _escape(substitutions[match.group(0)]),
-                       string)
+    check_null = False
+    for pattern, value in substitutions.items():
+        if pd.isnull(value):
+            check_null = True
+        string = string.replace(pattern, _escape(str(value)), 1)
```
connorcoley, line 103: Why only replace a single instance? I could imagine a situation in which multiple values need to be replaced. For example,
- During a solvent screen, multiple inputs are prepared as stock solutions in the same `$solvent_smiles$` that varies between experiments.
- A `$compound_name$` appears in an identifier and also in an unstructured field, like the procedure details

ord_schema/templating.py
```diff
@@ -63,20 +63,65 @@ def _escape(string):
     return string.replace('\\', '\\\\')
 
 
+def _is_null(value):
+    """Returns whether a value is null."""
+    return pd.isnull(value) or value == 'nan'
```
connorcoley, line 68: Are empty strings always parsed by Pandas as NaNs? Would you be opposed to adding `or (isinstance(value, str) and not value.strip())`?

ord_schema/templating.py
```diff
@@ -63,20 +63,65 @@ def _escape(string):
...
+
+
 def _replace(string, substitutions):
```
connorcoley, line 71: Could you rename this function? It now performs string substitutions, parses the reaction, performs the empty field checks, and returns a Reaction message rather than a string. I feel like "_replace" doesn't do it justice.

ord_schema/templating_test.py
```diff
@@ -106,6 +104,44 @@ def test_bad_placeholders(self):
...
+        expected_dataset = dataset_pb2.Dataset()
+        reaction1 = expected_dataset.reactions.add()
+        reaction1.CopyFrom(reaction)
+        reaction1.inputs['one'].components[0].identifiers[0].value = 'CN'
+        del reaction1.inputs['two']  # No components after amount removal.
+        reaction2 = expected_dataset.reactions.add()
+        reaction2.CopyFrom(reaction)
+        del reaction2.inputs['one'].components[0]  # No indentifiers.
+        reaction2.inputs['two'].components[0].mass.value = 1.5
+        self.assertEqual(dataset, expected_dataset)
```
connorcoley, line 143: It took me a very long time to understand this test case, admittedly. Would you be able to change the construction of your `df` template to construct from dictionaries? This will make it clearer that there are two partially-defined reactions: `{'smiles': ['CN', ''], 'mass': ['', 1.5]}` where maybe the empty string mass should be a NaN? Even better would be to construct a multiline .csv and read from a tempfile so it exactly mimics how the file might be parsed by Pandas.

connorcoley (changes requested):

ord_schema/templating.py
connorcoley:
```
>>> pd.isnull('')
False
```

ord_schema/templating.py
skearnes: That's fair; updated.

ord_schema/templating.py
skearnes: Happy to add that. Pandas *does* convert missing values to NaN but if there's an explicit empty string it wouldn't be caught.

ord_schema/templating.py
skearnes: Changed to `_fill_template`.

ord_schema/templating_test.py
skearnes: I'll create a CSV to mimic reality; good suggestion.

skearnes:
Thanks for the review; PTAL @connorcoley

connorcoley (approved):
Thanks!

PR #477, opened 2020-10-16
Add Santanilla example

michaelmaser:
Fully worked example for Santanilla experiment 2 submission. Adds dative SMILES to metal complexes, removes yields and reports only peak areas, adds aliquot step to workup, addresses changes to ord-schema


skearnes (changes requested):
It looks like your `git add` picked up some unrelated files:

* examples/.DS_Store
* examples/10_Perera_Science_Suzuki/.DS_Store
* examples/2_Nielsen_Deoxyfluorination_Screen/example_nielsen.ipynb

Could you please remove these? And you may want to add `.DS_Store` to the .gitignore.

skearnes (commented):
Also, I think we should leave examples/1_Santanilla_Nanomole_HTE/Santanilla_experiment2_dataset.pbtxt out of the PR and instead create a PR on ord-data.

@connorcoley do you agree?

connorcoley:
> Also, I think we should leave examples/1_Santanilla_Nanomole_HTE/Santanilla_experiment2_dataset.pbtxt out of this PR and use it to create a PR on ord-data.
> 
> @connorcoley do you agree?

Yes, I think it isn't necessary to include the dataset output here. Making it an official submission to ord-data would be terrific.

skearnes (commented):
Oops, I think you deleted example_nielsen.ipynb entirely instead of reverting your commit. I'd suggest copying it from HEAD and recommitting as-is.

michaelmaser:
> It looks like your `git add` picked up some unrelated files:
> 
> * examples/.DS_Store
> * examples/10_Perera_Science_Suzuki/.DS_Store
> * examples/2_Nielsen_Deoxyfluorination_Screen/example_nielsen.ipynb
> 
> Could you please remove these? And you may want to add `.DS_Store` to the .gitignore.
> 
> Update: Also examples/1_Santanilla_Nanomole_HTE/.DS_Store

Done.

> Also, I think we should leave examples/1_Santanilla_Nanomole_HTE/Santanilla_experiment2_dataset.pbtxt out of this PR and use it to create a PR on ord-data.
> 
> @connorcoley do you agree?

Done. Saved for ord-data PR.


> Oops, I think you deleted example_nielsen.ipynb entirely instead of reverting your commit. I'd suggest copying it from HEAD and recommitting as-is.

Fixed!

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
```diff
@@ -0,0 +1,8234 @@
...
+    "from rdkit.Chem.Draw import IPythonConsole, rdMolDraw2D\n",
+    "from IPython.display import SVG"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Load the dataset\n",
+    "The original dataset published by the authors is found in ./1259203_Datafiles.xlsx, and the \"Data S2- Experiment 2\" page was converted to .csv in ./experiment_2.csv"
```
skearnes, line 96: Consider using pd.read_excel

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_excel.html

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
```diff
@@ -0,0 +1,8234 @@
...
+   ],
+   "source": [
+    "# Extract unique values for each reaction component\n",
+    "reagent_dict = {}\n",
+    "component_types = ['Electrophile', 'Nucleophile', 'Catalyst', 'Base']\n",
+    "\n",
+    "# Create dictionary for each component set with names and SMILES\n",
+    "for component in component_types:\n",
+    "    component_dict = {}\n",
+    "    for i, unique in enumerate(list(data[component].unique())):\n",
```
skearnes, line 789: I don't think I understand the reason for including the indices here. Why not just use `unique` as the key?

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
```diff
@@ -0,0 +1,8234 @@
...
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# Add product identifier (dependent on nucleophile)\n",
+    "# Create nucleophile to product mapping\n",
+    "product_dict = {}\n",
+    "reagent_dict['Products'] = {}\n",
+    "for i, nucleophile in enumerate(reagent_dict['Nucleophiles'].keys()):\n",
```
skearnes, line 1126: Same here.

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
```diff
@@ -0,0 +1,8234 @@
...
+       "<IPython.core.display.SVG object>"
+      ]
+     },
+     "execution_count": 9,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# Testing PubChem SMILES for ferrocene complexes\n",
```
skearnes, line 1401: Maybe drop these cells since you explain what's going on in the comments above and AFAICT these cells don't contribute to the final Dataset?

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
```diff
@@ -0,0 +1,8234 @@
...
+    "    ARGUMENTS:\n",
+    "        (rdkit.Chem.rdchem.Mol)mol: mol object for the molecule in question\n",
+    "        \n",
+    "    RETURNS:\n",
+    "        (bool): boolean describing whether molecule has an atom with \n",
+    "                atomic number in the transition-metal ranges\n",
+    "    \"\"\"\n",
+    "    for atom in mol.GetAtoms():\n",
+    "        n = atom.GetAtomicNum()\n",
+    "        if (n>=22 and n<=29) or (n>=40 and n<=47) or (n>=72 and n<=79):\n",
```
skearnes, line 2171: Call `is_transition_metal` here instead of duplicating code.

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
```diff
@@ -0,0 +1,8234 @@
...
+    "            return True\n",
+    "    return False\n",
+    "\n",
+    "def set_dative_bonds(mol, fromAtoms=(7,8,15)):\n",
```
skearnes, line 2175: This would be good to add to e.g. `message_helpers.py` in another PR.

skearnes (commented):
Still reviewing this, but a few initial comments.

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
```diff
@@ -0,0 +1,8234 @@
...
+       "<text dominant-baseline=\"central\" style=\"font-size:7px;font-style:normal;font-weight:normal;fill-opacity:1;stroke:none;font-family:sans-serif;fill:#00CC00\" text-anchor=\"start\" x=\"1422.57\" y=\"239.737\"><tspan>Cl</tspan></text>\n",
+       "<text dominant-baseline=\"central\" style=\"font-size:7px;font-style:normal;font-weight:normal;fill-opacity:1;stroke:none;font-family:sans-serif;fill:#0000FF\" text-anchor=\"end\" x=\"1356.6\" y=\"248.826\"><tspan>H</tspan><tspan style=\"baseline-shift:sub;font-size:5.25px;\">2</tspan><tspan>N</tspan></text>\n",
+       "<path class=\"bond-0\" d=\"M 1781.41,143.941 L 1796.63,148.565\" style=\"fill:none;fill-rule:evenodd;stroke:#000000;stroke-width:2px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1\"/>\n",
+       "<path class=\"bond-1\" d=\"M 1796.63,148.565 L 1810.48,159.608\" style=\"fill:none;fill-rule:evenodd;stroke:#000000;stroke-width:2px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1\"/>\n",
+       "<path class=\"bo
```
skearnes, line 5406: It looks like you're not doing this; please remove the comment and the commented-out code.

skearnes, line 6171: The schema supports NANOLITER now.

skearnes, line 6405: I think this is fine, but it might be fun to try out the auto-offloading of `bytes_value` data here instead.

skearnes, line 6466: I think we want to set `volume_includes_solutes` for both solvents here.

skearnes, line 6550: Please remove unused code.

skearnes, line 7621: This would be better if the input names were things like 'electrophile' instead of 'reactant_1'.

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
michaelmaser: Done, thanks!

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
michaelmaser: Cleaned this up. Was using the index to name the molecules by their reaction role, but this isn't necessary. Replaced with `unique` (i.e., the dataset name) as suggested.

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
michaelmaser: Done.

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
michaelmaser: Done.

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
michaelmaser: Done. Good catch, thanks!

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
michaelmaser: All comments below this are referencing the same lines of raw notebook code that I can't really tell what they refer to. Could you direct me to the code? I cleaned up all the commented out code I could find. Meant to do this before submitting PR, though some issues Re: yield quantification were still up in the air at the time.

michaelmaser: Done. Good idea, thanks!

michaelmaser: Done, I think.

michaelmaser: Which solvents does this refer to? Is this from the workup?

michaelmaser: Great! Which component volume does this refer to?

michaelmaser: Does this refer to the images? I originally had the bytes_value as mentioned, but it was printing the full byte string every time it ran. Connor and I discussed this a little bit. Wanted a single image for the entire dataset; since this isn't available in the `Dataset` message, I opted to set a URL for each reaction instead of loading the full image data for 1536 examples. Let me know if this should be changed!

michaelmaser:
Thanks for the review, @skearnes!

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
skearnes: Yes, for the images. After #493 this should do the right thing.

skearnes: Sorry, this is specifically about:

```python
# Handle any cases where 'IS' = 0 and 0 division failed
# Reset any yields greater than 105% (ORD limit)
for column in ['Product_yield', 'ArBr_yield', 'Nu_yield']:
    data[column] = data[column].fillna(0)
    #data.loc[data[column] > 105, column] = 105
```

skearnes: All of the volumes are given as e.g. `0.25 uL`; this could be `250 nL`. Up to you.

skearnes: Yes, for the workup solvents.

skearnes (approved):
LGTM after the outstanding comments are addressed. Thanks!

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
michaelmaser: Done, thanks. Missed that one.

michaelmaser: Done.

michaelmaser: Done, reverted to bytes image data.

michaelmaser: Done, thanks. Good catch.

PR #478, opened 2020-10-17
Allow arbitrary compound data

### LINKED ISSUE #466: Add a repeated/map field for arbitrary Data to ReactionProducts
> 2) This library is monitored using an unconventional colorimetric analysis with indicator dye to trap the dehalogened byproduct HBr. Also the targeted measurement is rate constant, not yield or selectivity. I used selectivity cell to add the rate constant result and use detail to clarify it. Since the indicator is the limiting reactant used to capture the 1st-order decay rate along with 8 different concentrations span from 0-1.4 mM across 8 rows per photocatalyst column on the 96-welll, based on the input requirement, I improvised by using the average mM with +/- range and detailed in the add prep section. 

Using `selectivity` is probably the best place to put the rate constants in the current schema; @connorcoley should we add an arbitrary `Data` message to `ReactionProduct`?

_Originally posted by @skearnes in https://github.com/Open-Reaction-Database/ord-data/pull/10#issuecomment-707279628_

connorcoley:
Resolves #466

In addition to allowing arbitrary data to be defined for products (as suggested in issue #466), I thought it would be useful to also allow arbitrary data for compounds. I could imagine attaching the results of some QC to a starting material. For me, the most natural place to add a `map<string, Data>` field is to the Compound message so it can do double duty for starting materials and product compounds.

### DIFF
--- a/ord_schema/proto/reaction.proto
+++ b/ord_schema/proto/reaction.proto
@@ -260,6 +260,13 @@ message Compound {
     string how_computed = 4;
   }
   repeated Feature features = 13;
+  // Compounds may have arbitrary data associated with them. For example,
+  // product compounds could contain product-specific data that does not make
+  // sense to store at the analysis level, e.g., LC peak areas when multiple
+  // peaks are quantified. Reactants or other input species could contain
+  // data from quality control assays, e.g., an NMR spectrum from a purity
+  // check or the residual water ppm from a Karl Fischer titration.
+  map<string, Data> data = 15;
 }
 
 /**


skearnes (commented):
I like the idea of adding this map to `Compound`, but I think it would make sense to also add one to `ReactionProduct`. We already store compound-specific data at that level and it could be confusing to have product-level data in two different places.

WDYT?

connorcoley:
I did think about that. My inclination is that it would be worse to have a Data map for Compounds _and_ a Data map for ReactionProducts. I'm not strictly opposed to having it in both places. If we do that, then we're basically using `Compound product` solely for its `repeated CompoundIdentifier identifiers` and `amount` fields?

Re: compounds, a separate idea is that we could potentially remove Features from Compounds now since those could be considered a subset of Data.

michaelmaser:
Is product-level data not already in multiple places? The `ReactionProduct` has `Percentage yield`, its `Compound` has `amount` and `identifiers`, and the `outcomes.analyses` has `processed_data`. 

Think I added this on #466, but I think it would be nice to have `Data` as part of `ReactionProduct`. Not sure what that should do to the other messages.

skearnes:
This raises some interesting points. What's your reaction to the following?

* Add arbitrary data maps to both `Compound` and `ReactionProduct`
* Remove `features` from `Compound`
* _Remove_ `Compound` from `ReactionProduct` and replace it with `repeated CompoundIdentifier`  and `amount`

Essentially this makes `Compound` and `ReactionProduct` counterparts for inputs and outputs, respectively.

connorcoley:
> Remove Compound from ReactionProduct and replace it with repeated CompoundIdentifier and amount

Removing the Compound from ReactionProduct will break backwards compatibility and require us to revise all of the current examples and datasets. It will also make require some substantial changes to the editor in terms of how the Compound fieldset is reused in multiple places. I'm hesitant to make that change.

skearnes:
> Removing the Compound from ReactionProduct will break backwards compatibility and require us to revise all of the current examples and datasets. It will also make require some substantial changes to the editor in terms of how the Compound fieldset is reused in multiple places. I'm hesitant to make that change.

Understood. It would be a pain in the short-term. But now is the time to make this change if we think it would dramatically improve usability of the schema. If the long-term benefit seems modest, we can skip it.

skearnes:
As I think more about this specific question, I'm leaning toward not trying to rework the usage of `Compound`. I'm still in favor of adding a `Data` map to `ReactionProduct` so that product-level info is grouped at the same level.

Still, let's keep in mind that alpha testing is likely our last chance to make larger-scale, disruptive changes to the schema if we find compelling reasons to do so based on user feedback.

connorcoley:
> As I think more about this specific question, I'm leaning toward not trying to rework the usage of `Compound`. I'm still in favor of adding a `Data` map to `ReactionProduct` so that product-level info is grouped at the same level.
> 
> Still, let's keep in mind that alpha testing is likely our last chance to make larger-scale, disruptive changes to the schema if we find compelling reasons to do so based on user feedback.

I'll add that field to this PR.

And yes, point taken. @michaelmaser any opinion about the idea of replacing `Compound compound` in the ReactionProduct message with separate constituent fields `identifiers` and `amount`?

michaelmaser:
> And yes, point taken. @michaelmaser any opinion about the idea of replacing `Compound compound` in the ReactionProduct message with separate constituent fields `identifiers` and `amount`?

Yes, I like this. I agree would avoid re-working all of `Compound` usage, which works well for the rest of reaction components. 

I have been thinking about and am still wondering if/what sort of specific data fields should be included in `ReactionProduct` based on how they will be accessed downstream. Right now there is `Percentage compound_yield`, `Percentage purity`, and `Selectivity selectivity`. I like the addition of `identifiers` and `amount`.

If arbitrary `Data` maps are added, they would be directly linked to the product species, but the data would still have to be called by assigned name, right? I'm wondering if there is value in defining any structured/specific data fields within the product message to hopefully enforce consistency between submissions, like a `peak_area_count` message with enumerated methods, for example. Thinking mostly of HTE examples that commonly seem to report peak areas only...so if these are only recorded in arbitrary `analyses`, then there's no quantitative data linked to the product species. Thoughts? Could be a separate issue.

Aside; looking at the `ReactionProduct` message, I am realizing `Selectivity selectivity` might possibly need to be a repeated field...since a product could, e.g., be formed both enantio- and diastereoslectively. Thoughts?

skearnes:
@michaelmaser Just to clarify, I think that if we avoid reworking `Compound` we will keep `identifiers` and `amount` in `Compound` rather than adding them to `ReactionProduct`. Does that make sense?

I like the idea of adding some additional specific fields to `ReactionProduct` to encourage consistency across submissions.

PR #479, opened 2020-10-17
Name/SMILES getters/setters for compounds

### LINKED ISSUE #425: Add message helper for setting a compound's name or SMILES
During programmatic definition when manually templating, it can be annoying to keep track of the order of identifiers. For example, remembering whether `reaction.inputs['my_input'].components[0].identifiers[0]` is the name and `...identifiers[1]` is the SMILES or vice versa. 

A new message helper that is called as `message_helpers.set_name(compound, 'blah')` and iterates through the compound's identifiers to locate the one with the NAME type and alter the value would be convenient. Similarly, a function `message_helpers.set_smiles` that does the same for a SMILES-type compound identifier would be handy.

connorcoley:
Resolves #425

Note that the new function `get_compound_smiles` is NOT the same as `smiles_from_compound`. The newly-added former just gets the string value of the SMILES identifier (or None), whereas the latter tries to parse other structural identifiers as a back-up option.

I could envision combining these functions with an optional kwarg like `from_mol=False`. However, changing this would break the editor since it relies on `molblock_from_compound` at least. @skearnes what is our plan for synchronizing the two repos?


skearnes:
> I could envision combining these functions with an optional kwarg like from_mol=False. However, changing this would break the editor since it relies on molblock_from_compound at least. @skearnes what is our plan for synchronizing the two repos?

Please see https://github.com/Open-Reaction-Database/ord-editor/pull/11

ord_schema/message_helpers.py
```diff
@@ -431,6 +431,98 @@ def validate_reaction_smiles(reaction_smiles):
...
+
+    Args:
+        compound: Compound message.
+        identifier_type: The CompoundIdentifier type to retrieve the value of.
+        value: The value to set.
+
+    Returns:
+        The compound identifier that was modified or created.
+    """
+    for identifier in compound.identifiers:
```
skearnes, line 461: What happens if there are multiple identifiers of the same type? I don't think we check for that in validations.

ord_schema/message_helpers.py
```diff
@@ -431,6 +431,98 @@ def validate_reaction_smiles(reaction_smiles):
...
+        if identifier.type == identifier_type:
+            identifier.value = value
+            return identifier
+    identifier = compound.identifiers.add(type=identifier_type, value=value)
+    return identifier
+
+
+def get_compound_smiles(compound):
```
skearnes, line 469: re: merging code

I think it would be cool if smiles_from_compound called this function in its search for existing identifiers. But it's also not super complicated so I'm not saying that's a must.

After https://github.com/Open-Reaction-Database/ord-editor/pull/11 is merged you won't need to worry about breaking the editor.

skearnes (approved):

ord_schema/message_helpers.py
connorcoley: Added note to docstring

PR #485, opened 2020-10-20
Fixing bug in #479 made in haste

connorcoley:
In updating some message helpers to avoid redundant code, I hastily left in the wrong identifier. This patches a new bug from PR #479


skearnes (approved):

skearnes:
You might consider adding the test that would have caught this :)

Issue #48, opened 2020-03-19
Create Messages for common analytical data


connorcoley:
Rather than having completely unstructured analytical data (like strings with peaks for NMR), define messages that capture the real structure. Also need helper functions that parse strings into the more structured format.

bdeadman:
Messages were defined for analytical data in PR #496 but there are no helper functions defined. Need to determine if helper functions for analytical data are still relevant to ORD.

PR #496, opened 2020-10-27
Product measurement/characterization revision to the schema

### LINKED ISSUE #386: Clarify expected user behavior for reporting uncalibrated LC yields
Many HTE datasets use non-conventional means of quantifying yields, like relative peak areas in LC. We should be explicit in how we expect this data to be reported. Is it sufficient to say that a yield was quantified by LC without any internal standard or authentic standard? Or should we explicitly have a field that can capture this (relative common) scenario where the yield value is only relative.

### LINKED ISSUE #420: Improve authentic/internal standard definitions in analysis messages
- [ ] Warn users when "uses authentic standard" and "uses internal standard" are not defined since these are very important fields.
- [ ] Add a Compound-type field of "authentic_standard" to define what the authentic standard is. This will allow specification of vendor or preparation SYNTHESIZED and link to another reaction message  Many analyses are reported for new compounds (scaled up) along with HTE reactions that only use UPLC and/or MS; these should have  "uses authentic standard" set to True

### LINKED ISSUE #426: How to handle reactions where multiple LC peaks (incl. starting materials) are quantified?
During reaction screening, it is sometimes the case that we'll try to quantify the peak areas of all components we identify (leftover starting materials and the product(s)). Some questions are:
- How do we record the amount of a starting material left over? If we have one limiting reactant, we can use the conversion field. But what if we quantify species other than the limiting reactant? Are these species reported as products?
- If we do have multiple products tied to a single analysis, how do we interpret `uses_authentic_standard` and `uses_internal_standard`? These are defined at the analysis level, not the product level, so ambiguity is introduced. What if a calibration curve was generated for the intended product but we have an impurity whose identity we infer from an MS peak and whose yield we crudely estimate from uncalibrated LC areas?

### LINKED ISSUE #432: Add field in schema and web editor for "relative yield only"
In some cases, yields should only be treated relative to other yields within the same dataset, e.g., [the Santanilla HTE example](https://github.com/Open-Reaction-Database/ord-data/pull/9). I believe an optional boolean field `yield_is_relative` or something similar is needed. This is related to issue #420 but distinct enough to warrant its own discussion.

(CC @michaelmaser @skearnes @beef-broccoli)

### LINKED ISSUE #466: Add a repeated/map field for arbitrary Data to ReactionProducts

### LINKED ISSUE #506: Amount message?
As I was reviewing https://github.com/Open-Reaction-Database/ord-editor/issues/3, I found myself wondering why we don't have an `Amount` message:

```proto
message Amount {
  oneof kind {
    Mass mass = 1;
    Moles moles = 2;
    Volume volume = 3;
  }
}
```

This would change accessors like `getattr(compound, compound.WhichOneof('amount')).value` to `getattr(compound.amount, compound.amount.WhichOneof('kind')).value` but might be more readable since "Amount" will always show up in the pbtxt.

This would slightly reduce code duplication with the other amount oneof fields (in `CrudeComponent` and `Workup`) but also require additional validation to e.g. make sure that `moles` is not used when it is not appropriate.

I don't feel strongly about this either way, but wanted to bring it up as an opportunity to be slightly more explicit/consistent as we are making changes to the schema.

@connorcoley WDYT?

connorcoley:
Resolves #432 (where most discussion has been), resolves #386, resolves #420, resolves #426, resolves #466, resolves #506 

For discussion here:
1. Is it redundant to have both "counts" and "intensity"? I feel like the two aren't completely interchangeable, but there is certainly ambiguity.
2. Are we being completely consistent in terms of holding product-specific analytical data in this new ProductMeasurement field? In particular, I'm thinking about IR and HRMS of an isolated compound. The entire analysis is of the presumably-isolated product. Should ProductMeasurement accommodate a structured definition of the full IR spectrum? We shied away from this idea (#48) since it got rather complicated. However, HRMS is a very easy measurement to capture in a structured format -- the dominant peak is just a float m/z. It seems silly not to include this in the ProductMeasurement field. The temperature of a MP experiment or the angle of a CD experiment are similar.
3. To broaden the previous point, some analyses are applied to crude product mixtures, while others are applied to thought-to-be-pure compounds. in the former case, we definitely need this ProductMeasurement message to create a closer association between an analysis and a single product. But in the latter case, there is already an implicit connection. Some analyses result in a scalar value while others result in more structured data. One option is to add a Data field to the ProductMeasurement and a SPECTRUM type, but then there is ambiguity around when to use that field rather than the ReactionAnalysis's raw_data and processed_data maps. If we do add a Data field to ProductMeasurement, it would certainly be very handy for non-HTE data where people do full product characterizations; we would have access to many examples of structures and their NMR/IR with which to build predictive models.
4. I've kept color and texture in the ReactionProduct itself. They are not supposed to be tied to a specific analysis (unless we add "VISUAL_INSPECTION" as an analysis which is silly).
5. I don't think that uses_internal_standard and uses_authentic_standard need to be defined for the analysis, as they need to be defined for each (product, analysis) to handle the situation where multiple products are quantified and not all necessarily use an authentic or internal standard.

Note that this schema change  is not backwards compatible. The old schema is included in this repo to make data transfer easier.


ord_schema/proto/reaction.proto
```diff
@@ -861,26 +852,104 @@ message ReactionProduct {
...
+    COUNTS = 8;
+    // Intensity, e.g., of a TLC spot, ELSD, or GC FID.
+    INTENSITY = 9;
+    // Direct m/z measurement, e.g., from HRMS. The value field should be used
+    // to report the m/z, not the "mass" field.
+    MZ = 10;
+    // Temperature, e.g., from a melting point experiment.
+    TEMPERATURE = 11;
+    // Angle, e.g., from a circular dichroism experiment. Units of degrees
+    // centimer squared per decimol are assumed.
```
skearnes, line 906: We should probably add an Angle message with units to be consistent with e.g. Temperature.

ord_schema/proto/reaction.proto
```diff
@@ -861,26 +852,104 @@ message ReactionProduct {
...
-  optional float value = 3;
-  // TODO(kearnes): What does precision mean in this context?
-  optional float precision = 4;
+  MeasurementType type = 2;
+  // Additional information about the measurement and, in particular, how it
+  // was calculated if that is ambiguous given the analysis type.
+  string details = 3;
+  // Whether an internal standard was used for this measurement to normalize
+  // it, e.g., for quantification of yield or relative peak area.
+  optional bool uses_internal_standard = 4;
```
skearnes, line 915: Do we want another Compound message for the internal standard? Or should that be part of the inputs and/or the ReactionAnalysis message?

ord_schema/proto/reaction.proto
```diff
@@ -861,26 +852,104 @@ message ReactionProduct {
...
+  // information about its precise salt form and source/vendor.
+  Compound authentic_standard = 6;
+  // The numeric value for this measurement. For example, the yield expressed
+  // as a percentage (i.e., 50% yield corresponds to 50 and not 0.5). If the
+  // value is associated with units, those unit-specific fields should be used
+  // instead.
+  optional float value = 7;
+  optional float precision = 8;
+  // The value for this measurement metric when it cannot be represented by a
+  // single number.
```
skearnes, line 932: "e.g. selectivity as '2:1'"?

ord_schema/proto/reaction.proto
```diff
@@ -861,26 +852,104 @@ message ReactionProduct {
...
+  // being used as a comparison for other analytical methods.
+  optional bool uses_authentic_standard = 5;
+  // Details about the authentic standard used, optionally including
+  // information about its precise salt form and source/vendor.
+  Compound authentic_standard = 6;
+  // The numeric value for this measurement. For example, the yield expressed
+  // as a percentage (i.e., 50% yield corresponds to 50 and not 0.5). If the
+  // value is associated with units, those unit-specific fields should be used
+  // instead.
+  optional float value = 7;
```
skearnes, line 929: This means that the Percentage message is only used in one place (ReactionOutcome.conversion) and should probably be removed entirely.

Alternatively, we could rename Percentage as "Value" or "Measurement" and use it here. This would allow us to use `oneof` to wrap the different value fields.

skearnes:
> Resolves #432 (where most discussion has been), resolves #386, resolves #420, resolves #426
> 
> For discussion here:
> 
> 1. Is it redundant to have both "counts" and "intensity"? I feel like the two aren't completely interchangeable, but there is certainly ambiguity.

I don't have strong feelings about this, but I think if they are not interchangeable then keeping them separate is fine.

> 2. Are we being completely consistent in terms of holding product-specific analytical data in this new ProductMeasurement field? In particular, I'm thinking about IR and HRMS of an isolated compound. The entire analysis is of the presumably-isolated product. Should ProductMeasurement accommodate a structured definition of the full IR spectrum? We shied away from this idea (#48) since it got rather complicated. However, HRMS is a very easy measurement to capture in a structured format -- the dominant peak is just a float m/z. It seems silly not to include this in the ProductMeasurement field. The temperature of a MP experiment or the angle of a CD experiment are similar.

I'm not opposed to adding structured analysis messages piecemeal, especially if we can start with easy ones. We don't have to try to cover all the possibilities at once (and adding new ones to ProductMeasurement later will be backwards compatible).

> 3. To broaden the previous point, some analyses are applied to crude product mixtures, while others are applied to thought-to-be-pure compounds. in the former case, we definitely need this ProductMeasurement message to create a closer association between an analysis and a single product. But in the latter case, there is already an implicit connection. Some analyses result in a scalar value while others result in more structured data. One option is to add a Data field to the ProductMeasurement and a SPECTRUM type, but then there is ambiguity around when to use that field rather than the ReactionAnalysis's raw_data and processed_data maps. If we do add a Data field to ProductMeasurement, it would certainly be very handy for non-HTE data where people do full product characterizations; we would have access to many examples of structures and their NMR/IR with which to build predictive models.

This is indeed tricky. Maybe we need to think more broadly about how the analysis/product pairs should be structured. My current mental model is that an analysis may contain data that covers one or more compounds, and ProductMeasurement is the place to store compound-specific data extracted from an analysis. So I think that structured analysis data might be best placed in the ReactionAnalysis and summary statistics (like a specific peak area or quantified yield) are best placed in ProductMeasurement.

That doesn't really resolve what to do about single-compound analyses, like melting point, where the data could be in either place?

> 4. I've kept color and texture in the ReactionProduct itself. They are not supposed to be tied to a specific analysis (unless we add "VISUAL_INSPECTION" as an analysis which is silly).

SGTM.

> 5. I don't think that uses_internal_standard and uses_authentic_standard need to be defined for the analysis, as they need to be defined for each (product, analysis) to handle the situation where multiple products are quantified and not all necessarily use an authentic or internal standard.

Agreed; the placement in ProductMeasurement seems appropriate.

> Note that this schema change breaks many things (hence the draft PR) and is not backwards compatible

Huzzah for alpha testing!

michaelmaser:
Hope to take a look this afternoon

ord_schema/proto/reaction.proto
connorcoley: I think the addition of an internal standard should be part of an input or workup step.

The only situation where this breaks down is if the crude product mixture is split into multiple aliquots to be worked up and analyzed separately, where only some of the analyses use a workup with an internal standard

ord_schema/proto/reaction.proto
connorcoley: Will add

ord_schema/proto/reaction.proto
connorcoley: Will add

ord_schema/proto/reaction.proto
connorcoley: I don't love the additional nesting that would require, but maybe that's a minor point since it only affects the programmatic definition. We'd need to go to reaction.outcomes[0].products[0].measurements[0].value.value in that case. The advantage is that it would package up `float value` and `float precision` into a separate message.

michaelmaser:
> 2. Are we being completely consistent in terms of holding product-specific analytical data in this new ProductMeasurement field? In particular, I'm thinking about IR and HRMS of an isolated compound. The entire analysis is of the presumably-isolated product. Should ProductMeasurement accommodate a structured definition of the full IR spectrum? We shied away from this idea (#48) since it got rather complicated. However, HRMS is a very easy measurement to capture in a structured format -- the dominant peak is just a float m/z. It seems silly not to include this in the ProductMeasurement field. The temperature of a MP experiment or the angle of a CD experiment are similar.
>> I'm not opposed to adding structured analysis messages piecemeal, especially if we can start with easy ones. We don't have to try to cover all the possibilities at once (and adding new ones to ProductMeasurement later will be backwards compatible).

For IR, a similar field to HRMS but just repeated could be sufficient. If needed, an IR spectrum in a message could be captured by a `repeated float value`

> 3. To broaden the previous point, some analyses are applied to crude product mixtures, while others are applied to thought-to-be-pure compounds. in the former case, we definitely need this ProductMeasurement message to create a closer association between an analysis and a single product. But in the latter case, there is already an implicit connection. Some analyses result in a scalar value while others result in more structured data. One option is to add a Data field to the ProductMeasurement and a SPECTRUM type, but then there is ambiguity around when to use that field rather than the ReactionAnalysis's raw_data and processed_data maps. If we do add a Data field to ProductMeasurement, it would certainly be very handy for non-HTE data where people do full product characterizations; we would have access to many examples of structures and their NMR/IR with which to build predictive models.

A `repeated float value` of peak chemical shifts could _almost_ capture an NMR spectrum too, except for 1H where `string` splitting information is also needed. We could try to build a single `Spectrum` message that could handle these, IR, and others with `repeated float value` and `details` fields, or could build specific messages for each expected spectrum type that isn't just floats

>> My current mental model is that an analysis may contain data that covers one or more compounds, and ProductMeasurement is the place to store compound-specific data extracted from an analysis.

I agree with this model, where `ReactionAnalysis` is used largely for mixtures or in-progress/crude analysis, and `ProductMeasurement` is used for single-component or purified/final compound analysis.

>>So I think that structured analysis data might be best placed in the ReactionAnalysis and summary statistics (like a specific peak area or quantified yield) are best placed in ProductMeasurement.

I think this is fine, but if "structured analysis data" is carried out on a purified compound or is being used to determine yield/identity of a product, I think I would prefer it be placed in `ProductMeasurement` (I guess like you're saying for "specific peak area or quantified yield").

>>That doesn't really resolve what to do about single-compound analyses, like melting point, where the data could be in either place?

As above, I personally think these single-compound analyses should be contained in `ReactionProduct`.

> 4. I've kept color and texture in the ReactionProduct itself. They are not supposed to be tied to a specific analysis (unless we add "VISUAL_INSPECTION" as an analysis which is silly).

Agreed.

> 5. I don't think that uses_internal_standard and uses_authentic_standard need to be defined for the analysis, as they need to be defined for each (product, analysis) to handle the situation where multiple products are quantified and not all necessarily use an authentic or internal standard.
>> Agreed; the placement in ProductMeasurement seems appropriate.

Right, sounds good.

ord_schema/proto/reaction.proto
michaelmaser: >I think the addition of an internal standard should be part of an input or workup step.

Agreed.

ord_schema/proto/reaction.proto
michaelmaser: This is where the values get ambiguous for things like NMR/IR spectra, yes? Is it worth just making this value repeated? Or do we think full spectra messages will be needed where this is enforced?

ord_schema/proto/reaction.proto
michaelmaser: Do we want to give guidance on how this should be used? E.g., where "2:1" is (desired / all others) _or_ if there are more than two then use, e.g., "2:1:1", which would be "1:1" desired / all others (in the `SelectivityType` message it's written as "desired / undesired")? Also, how do you envision this will be handled downstream if some people enter this string, some enter "67:33" and others use `float value` 2? Some might also say the _selectivity value_ of a reaction giving a 2:1 product mixture is really 67%, which would be handled by `float value` as 67 or 0.67... Is it worth creating a structure to standardize all this?

ord_schema/proto/reaction.proto
```diff
@@ -861,26 +852,104 @@ message ReactionProduct {
...
+  optional float precision = 8;
+  // The value for this measurement metric when it cannot be represented by a
+  // single number.
+  string string_value = 9;
+  Mass mass = 10;
+  Temperature temperature = 11;
+  Time time = 12;
```
michaelmaser, line 936: Do these refer to what temperature and time the measurement was taken at? Or, is it supposed to be used with analyses like MP and measurements like RETENTION_TIME that report a temperature or time?

ord_schema/proto/reaction.proto
```diff
@@ -861,26 +852,104 @@ message ReactionProduct {
...
+  // The numeric value for this measurement. For example, the yield expressed
+  // as a percentage (i.e., 50% yield corresponds to 50 and not 0.5). If the
+  // value is associated with units, those unit-specific fields should be used
+  // instead.
+  optional float value = 7;
+  optional float precision = 8;
+  // The value for this measurement metric when it cannot be represented by a
+  // single number.
+  string string_value = 9;
+  Mass mass = 10;
```
michaelmaser, line 934: Is this meant for a mass obtained if the Measurement is `YIELD` or should this only be included in `Compound amount`? Or, is it mass found referring to `MZ` in HRMS? Both the found and calculated masses are usually reported for HRMS, are we going to try to handle this or just only report the found? Suppose this could be a reason for structured messages for the analysis types.

ord_schema/proto/reaction.proto
```diff
@@ -861,26 +852,104 @@ message ReactionProduct {
...
+  Temperature temperature = 11;
+  Time time = 12;
+  // The more specific type of selectivity that is being reported, e.g., EE,
+  // ER, DR, EZ, ZE. EZ, ER, and DR assume that any product labeled as desired
+  // is first in the calculation, whereas EZ and ZE ratios are calculated as
+  // written.
```
michaelmaser, line 940: "only to be used when MeasurementType = SELECTIVITY"?

ord_schema/proto/reaction.proto
```diff
@@ -822,28 +815,26 @@ message ReactionOutcome {
   Percentage conversion = 2;
```
michaelmaser, line 815: I know this refers to a starting material and not a product, but it almost feels like this is well-suited as a `ProductMeasurement.MeasurementType` since it is useful to know how the conversion was measured (NMR, LC, etc.). If the limiting reactant is being quantified (which it has to be to report conversion), it could technically be included in `ReactionOutcome.products`, in which case the `ProductMeasurement.MeasurementType = CONVERSION` could be specified along with the `analysis_key` (I think we haven't really decided on how to handle measurements of remaining starting materials).

ord_schema/proto/reaction.proto
connorcoley: I think we should be clear about expected usage. In the case you brought up, the ambiguity between 67, 0.67, and 2 wouldn't be addressed by the fields themselves since all three are floats, but by the comments. 

If someone were to be tempted to enter "2:1:1", we would need to know what the other species were (unless it was somehow obvious/unambiguous, which I think could be true if there are only two species 2:1 but not for more than two). In that case, I'm not actually sure that we care about "2:1:1" -- we would rather have the area/counts of each product as a separate measurement, which could be combined to get the relative yields. "2:1:1" is after an additional stage of processing. If that information isn't available (e.g., entering an already published reaction), then I think associating "2:1:1" with the first/primary product and having the other numbers correspond to other listed products, in order, might make sense.

ord_schema/proto/reaction.proto
connorcoley: The latter -- I will add a comment to make that clearer

ord_schema/proto/reaction.proto
connorcoley: That mass was meant to be for isolated yield, not m/z. 

That's a good question about HRMS. I would consider the calculated mass a feature of the compound itself, while the found is associated with a specific analysis/product/measurement. However, defining those values in two separate messages would be rather confusing and unintuitive. 

This does make it more attractive to define a complete set of structured messages for common analyses...

ord_schema/proto/reaction.proto
connorcoley: I agree that it's useful to know how it was reported. There's also the complication that one can quantify the remaining amounts of multiple starting materials, which a single `conversion` field doesn't address

ord_schema/proto/reaction.proto
michaelmaser: I agree, though this does raise the question of whether `SELECTIVITY` is necessary at all. If we plan to derive it downstream from the _yields_ of multiple species anyway (which I think is sensible), we then expect that each species is defined as a product and quantified. Selectivity would then be inherently extractable. 

That said, I can see the value of a quick call to the selectivity metric, so long as usage is consistent. It is also a clear marker of, e.g., enantio- or diastereoselective reactions, if looking to learn there.

connorcoley:
Discussion notes:
- Try to keep all original analytical data in one place -- the analysis message?
- Add a boolean field to an analysis to say whether it's for a crude or (theoretically) purified product?
- Storing structured analytical data for purified compounds isn't an explicit goal of this effort - let's focus on the primary reason for this modification, which is to allow us to distinguish relative/real yields and how other summary statistics (e.g., purity, selectivity) were derived with or without an internal standard
- A boolean field in the analysis message would be helpful -- "is product-specific" or "is_of_isolated_species" to indicate that the analysis is supposed to correspond to a single species

connorcoley:
New question for discussion, relevant to the `raw_data` and `processed_data` maps of an analysis: should we add arbitrary unit-ed fields here? For example, in addition to having float_value, string_value, etc., should we have `Temperature temperature_value`, `Mass mass_value`, etc.?

ord_schema/proto/reaction.proto
connorcoley: Added a comment about including multiple starting materials as "products" if individually quantified

ord_schema/proto/reaction.proto
connorcoley: Added

ord_schema/proto/reaction.proto
connorcoley: No longer relevant since we've removed structured analytical data from the ProductMeasurement

ord_schema/proto/reaction.proto
connorcoley: Not completely relevant since we've removed structured analytical data from the ProductMeasurement. I've left in a retention_time field since I imagine that will be very commonly associated with an AREA or INTENSITY for LC or LCMS analyses. What do you think? This is still focusing on a product-specific measurement extracted from an analysis.

Would it make sense to add a repeated field for EIC so the m/z ratios used to get an INTENSITY are specified here? That seems like it might be a fairly common value to specify for HTE data. Knowing the EIC that was used to calculate. I've added one in, but let me know if you think this is extraneous.

ord_schema/proto/reaction.proto
connorcoley: @michaelmaser 's comment is largely irrelevant after moving structured analytical data back over to the analysis message.

I've taken @skearnes 's suggestion to use a oneof -- I think that should fit well here.

ord_schema/proto/reaction.proto
connorcoley: Elaborated in a comment.

ord_schema/proto/reaction.proto
connorcoley: With the move of analytical data even for simple measurements like CD, MP, or HRMS back to the analysis message, I don't believe this suggestion is relevant

ord_schema/proto/reaction.proto
```diff
@@ -861,26 +860,136 @@ message ReactionProduct {
...
-  // TODO(kearnes): What does precision mean in this context?
-  optional float precision = 4;
+  MeasurementType type = 2;
+  // Additional information about the measurement and, in particular, how it
+  // was calculated if that is ambiguous given the analysis type.
+  string details = 3;
+  // Whether an internal standard was used for this measurement to normalize
+  // it, e.g., for quantification of yield or relative peak area. If this
+  // field is true, the value field should contain the normalized quantity.
+  optional bool uses_internal_standard = 4;
```
skearnes, line 918: nit: "uses" vs. "used"?

Or just `bool internal_standard`?

ord_schema/proto/reaction.proto
```diff
@@ -861,26 +860,136 @@ message ReactionProduct {
...
+  // Whether an authentic sample of a material was used for callibrating
+  // an analytical technique, e.g., for product identification by LCMS.
+  // "Authentic" refers to a sample of a compound whose structure has been
+  // proven by the accepted means (1H NMR, 13C NMR, IR, HRMS), and is now
+  // being used as a comparison for other analytical methods.
+  optional bool uses_authentic_standard = 5;
+  // Details about the authentic standard used, optionally including
+  // information about its precise salt form and source/vendor.
+  Compound authentic_standard = 6;
```
skearnes, line 927: Since we have this `Compound` here, we could just check whether it is defined instead of having the additional `bool`. Or if the bool is handy we should make sure to check consistency in validations.

ord_schema/proto/reaction.proto
```diff
@@ -861,26 +860,136 @@ message ReactionProduct {
...
+  // More details about how a particular performance metric (e.g., counts)
+  // was calculated from an analysis using mass spec. This includes
+  // distinguishing total ion current (TIC) counts from extracted ion
+  // chromatogram (EIC) counts and their corresponding mass ranges or masses.
+  message MassSpecType {
+    enum MassSpecTypeEnum {
+      UNSPECIFIED = 0;
+      CUSTOM = 1;
+      TIC = 2;
+      TIC_POS = 3;
```
skearnes, line 958: nit: I think it's fine to use accepted acronyms, but it's good practice to avoid _abbreviations_ (see https://google.github.io/styleguide/pyguide.html#316-naming). So I'd suggest e.g. `TIC_POSITIVE`.

ord_schema/proto/reaction.proto
```diff
@@ -916,6 +1025,10 @@ message ReactionAnalysis {
     ED = 23;  // Electron diffraction, including MicroED.
   }
   AnalysisType type = 1;
+  // Whether this analysis is intended to be of an isolated species. If false
+  // or unspecified, the assumption is that it is of a crude product mixture
+  // (or a partially worked-up product mixture).
+  optional bool is_of_isolated_species = 8;
```
skearnes, line 1031: This is a bit awkward. Possible alternatives:

- `is_isolated_species`
- `isolated_species`

We've (mostly) taken an `is/has/uses/used<X>` naming style for boolean values in the schema so far. I'm wondering if we should take this opportunity to drop these prefixes to make things flow a bit better. IMO a checkbox in the editor that says "desired_product" is at least as (or more) readable than "is_desired_product".

See https://medium.com/javascript-in-plain-english/naming-boolean-variables-prefixes-df973d7d7ec3

ord_schema/proto/reaction.proto
```diff
@@ -1127,12 +1231,29 @@ message FlowRate {
...
+  }
+}
 // Used for things like conversion and yield.
 message Percentage {
   optional float value = 1;
   // Precision of the measurement (with the same units as `value`).
   optional float precision = 2;
 }
+// Wrapper for floats to include a precision.
+message FloatValue {
```
skearnes, line 1252: I like this.

ord_schema/proto/reaction.proto
```diff
@@ -1127,12 +1231,29 @@ message FlowRate {
   optional float precision = 2;
   FlowRateUnit units = 3;
 }
+message CircularDichroism {
```
skearnes, line 1234: I assume this isn't finished?

skearnes (commented):
This is looking very good. A few nits and style comments.

skearnes:
> New question for discussion, relevant to the `raw_data` and `processed_data` maps of an analysis: should we add arbitrary unit-ed fields here? For example, in addition to having float_value, string_value, etc., should we have `Temperature temperature_value`, `Mass mass_value`, etc.?

Just to clarify, are you thinking about expanding the `Data` message or adding these to `ReactionAnalysis`?

connorcoley:
> > New question for discussion, relevant to the `raw_data` and `processed_data` maps of an analysis: should we add arbitrary unit-ed fields here? For example, in addition to having float_value, string_value, etc., should we have `Temperature temperature_value`, `Mass mass_value`, etc.?
> 
> Just to clarify, are you thinking about expanding the `Data` message or adding these to `ReactionAnalysis`?

This would be an expansion of the `Data` message

ord_schema/proto/reaction.proto
connorcoley: I like the current naming style tbh. I think `analysis.is_of_isolated_species = True` is clearer than `analysis.isolated_species = True`. Obviously the form can use whatever label we want

ord_schema/proto/reaction.proto
connorcoley: Oops

michaelmaser:
> New question for discussion, relevant to the `raw_data` and `processed_data` maps of an analysis: should we add arbitrary unit-ed fields here? For example, in addition to having float_value, string_value, etc., should we have `Temperature temperature_value`, `Mass mass_value`, etc.?

I like the idea of including these structured unit-ed / `Data` types here. Could open opportunity to add spectra / listed data forms later on

skearnes:
> > New question for discussion, relevant to the `raw_data` and `processed_data` maps of an analysis: should we add arbitrary unit-ed fields here? For example, in addition to having float_value, string_value, etc., should we have `Temperature temperature_value`, `Mass mass_value`, etc.?
> 
> I like the idea of including these structured / unit-ed `Data` types here. Could open opportunity to add spectra / listed data forms later on

I think I'm opposed to extending the `Data` message. It's designed to be the place to put things that have no other place---an unstructured "catch all" message. I recognize that there could be structured data that has no home, but I think that should be fixed with changes to the schema and not feature creep of `Data`.

I'd be more interested in adding these fields to `ReactionAnalysis`.

michaelmaser:
> I'd be more interested in adding these fields to `ReactionAnalysis`.

SGTM

michaelmaser:
LGTM + readability

ord_schema/proto/reaction.proto
```diff
@@ -818,32 +810,42 @@ message ReactionWorkup {
...
+  // messages to indicate which analyses were used to derive which performance
+  // values/metrics. The string used for the key carries no meaning outside of
+  // this cross-referencing.
+  map<string, Analysis> analyses = 4;
 }
 
 message ReactionProduct {
-  Compound compound = 1;
+  // The identity of the product compound.
+  repeated CompoundIdentifier identifiers = 1;
```
skearnes, line 834: Let's preserve the tags for now.

ord_schema/proto/reaction.proto
skearnes: We forgot to discuss this...

skearnes (approved):
A couple unresolved comments, but otherwise LGTM. Thanks!

ord_schema/proto/reaction.proto
connorcoley: I think the bool is handy, and there are cases where an internal standard is used but there might be some ambiguity in the publication or lack of details. I think that we can add a validation error if the compound is defined but the bool isn't True and a validation warning if the bool is True but the compound is not defined

ord_schema/proto/reaction.proto
connorcoley: After discussion, decided to avoid structured analytical data messages for now and stick with the unstructured Data fields in the analysis (with no data other than derived product-specific statistics in ProductMeasurement)

ord_schema/proto/reaction.proto
connorcoley: No longer relevant

ord_schema/proto/reaction.proto
connorcoley: Does this mean (1) avoid using the old numbers, (2) add `reserved` statements for the tag numbers, and/or (3) add `reserved` statements for the field names?

skearnes: Definitely (1). And if you delete a field, please do use `reserved` to keep track of the tag numbers.

Note that this _may_ require new field names for renamed messages. For instance `map<string, Analysis>` may need to be a new field rather than overwriting `map<string, ReactionAnalysis>`.

connorcoley: I'm slightly confused as to why it is important to avoid those tag numbers / reserve them if the plan is to dump everything to a pbtxt format where the field names are explicit. But I will push an updated version momentarily

skearnes: It's mostly to be safe and keep the *.pb files working. But that's not a huge deal. Up to you.

skearnes:
Thinking about migration: I think it will make migration easier if we don't delete anything from the schema. This way all of the old protos will continue to load properly and we can move things over without needing two copies of the repo (IIUC the current *.pbtxt messages won't parse after their fields are removed). It will be really ugly, but will make the migration much easier and then we'll clean everything up.

I'm thinking of something like this:

```proto
...
// BEGIN DEPRECATED
message Feature {	  map<string, Data> features = 15;
  string name = 1;	  // Compounds may be assayed for quality control; analytical data should be
  oneof kind {	  // defined in the analyses map.
    string string_value = 2;	  map<string, Analysis> analyses = 16;
    float float_value = 3;	
  }	
  // Details of computation. Software, theory level, solvation, etc.	
  string how_computed = 4;	
}	
repeated Feature features = 13;
// END DEPRECATED
map<string, Data> features = 15;
// Compounds may be assayed for quality control; analytical data should be
// defined in the analyses map.
map<string, Analysis> analyses = 16;
...
```

Is that too much trouble?

ord_schema/proto/reaction.proto
```diff
@@ -862,32 +868,147 @@ message ReactionProduct {
...
+
+  }
+  // A time to be used only when specifying the retention time at which a peak
+  // area or intensity was calculated, e.g., as in LC, LCMS, GC, or GCMS.
+  Time retention_time = 11;
+  // More details about how a particular performance metric (e.g., counts)
+  // was calculated from an analysis using mass spec. This includes
+  // distinguishing total ion current (TIC) counts from extracted ion
+  // chromatogram (EIC) counts and their corresponding mass ranges or masses.
+  message MassSpecType {
```
skearnes, line 964: nit: `MassSpecMeasurementType`?

connorcoley:
> Thinking about migration: I think it will make migration easier if we don't delete anything from the schema. ... Is that too much trouble?

No, I think that's reasonable. It will be very ugly, to be sure, but will make migration _significantly_ easier. There will be some edge cases where it's not possible to do that because of field name conflicts (e.g., we're changing the type of `Compound.features` but keeping the name), but the majority of cases will be fine

ord_schema/proto/reaction.proto
connorcoley: Will fix

connorcoley:
> > Thinking about migration: I think it will make migration easier if we don't delete anything from the schema. ... Is that too much trouble?
> 
> No, I think that's reasonable. It will be very ugly, to be sure, but will make migration _significantly_ easier. There will be some edge cases where it's not possible to do that because of field name conflicts (e.g., we're changing the type of `Compound.features` but keeping the name), but the majority of cases will be fine

I might rescind that answer. We can't do it for `Compound.features` or `ReactionOutcome.analyses` unless we want to choose different names for the fields, which I wouldn't advise.

I've added in the old schema as `reaction_old.proto`. This might seem sloppy, but we can at least make the transition since both versions will be importable in the same environment

skearnes:
> > > Thinking about migration: I think it will make migration easier if we don't delete anything from the schema. ... Is that too much trouble?
> > 
> > 
> > No, I think that's reasonable. It will be very ugly, to be sure, but will make migration _significantly_ easier. There will be some edge cases where it's not possible to do that because of field name conflicts (e.g., we're changing the type of `Compound.features` but keeping the name), but the majority of cases will be fine
> 
> I might rescind that answer. We can't do it for `Compound.features` or `ReactionOutcome.analyses` unless we want to choose different names for the fields, which I wouldn't advise.
> 
> I've added in the old schema as `reaction_old.proto`. This might seem sloppy, but we can at least make the transition since both versions will be importable in the same environment

SGTM. Thanks.

connorcoley:
Here is a snippet from the updated Liu et al. OrgSyn example:

```
## Product and characterization
outcome = reaction.outcomes.add()
outcome.reaction_time.CopyFrom(unit_resolver.resolve('6 h'))
analysis = reaction_pb2.Analysis() # for handy enum reference

outcome.analyses['isolated_weight'].type = analysis.WEIGHT
outcome.analyses['isolated_weight'].is_of_isolated_species = True
outcome.analyses['isolated_weight'].details = '27.5-28.0 g recovered after workup'

outcome.analyses['1H NMR'].type = analysis.NMR_1H
outcome.analyses['1H NMR'].is_of_isolated_species = True
outcome.analyses['1H NMR'].details = '400 MHz, CDCl3'
outcome.analyses['1H NMR'].data['peaks'].string_value = r'0.92 (s, 9H), 4.06 (s, 4H), 7.23 - 7.34 (m, 6H), 7.40 (d, J = 7.1 Hz, 4H)'
outcome.analyses['1H NMR'].data['peaks'].description = 'List of peaks'

outcome.analyses['13C NMR'].type = analysis.NMR_13C
outcome.analyses['13C NMR'].is_of_isolated_species = True
outcome.analyses['13C NMR'].details = '101 MHz, CDCl3'
outcome.analyses['13C NMR'].data['peaks'].string_value = r'27.1, 38.4, 62.4, 127.7, 128.3, 129.6, 136.2, 176.3'
outcome.analyses['13C NMR'].data['peaks'].description = 'List of peaks'

outcome.analyses['thin film IR'].type = analysis.IR
outcome.analyses['thin film IR'].is_of_isolated_species = True
outcome.analyses['thin film IR'].details = 'neat film, NaCl'
outcome.analyses['thin film IR'].data['peaks'].string_value = r'3064, 3031, 2973, 2932, 2906, 2872, 1751, 1496, 1479, 1456, 1273, 1116, 1029, 738, 698'
outcome.analyses['thin film IR'].data['peaks'].description = 'List of peaks [cm-1]'

outcome.analyses['HRMS'].type = analysis.HRMS
outcome.analyses['HRMS'].is_of_isolated_species = True
outcome.analyses['HRMS'].details = 'ESI-TOF'
outcome.analyses['HRMS'].data['expected'].float_value = 298.1802
outcome.analyses['HRMS'].data['expected'].description = 'Expected m/z'
outcome.analyses['HRMS'].data['found'].float_value = 298.1794
outcome.analyses['HRMS'].data['found'].description = 'Observed m/z'

outcome.analyses['quantitative NMR'].type = analysis.NMR_1H
outcome.analyses['quantitative NMR'].is_of_isolated_species = True
outcome.analyses['quantitative NMR'].details = 'Quantitative NMR using 1,1,2,2-tetrachloroethane (>98%, purchased from Alfa Aesar) in CDCl3'

# A single product was desired and characterized
product = outcome.products.add()
product.identifiers.add(type='SMILES', value='O=C(C(C)(C)C)ON(CC1=CC=CC=C1)CC2=CC=CC=C2')
product.identifiers.add(type='NAME', value='N,N-Dibenzyl-O-pivaloylhydroxylamine')
product.mass.CopyFrom(unit_resolver.resolve('27.75 g'))

# Define which analyses were used for which aspects of characterization
product.measurements.add(type='IDENTITY', analysis_key='1H NMR')
product.measurements.add(type='IDENTITY', analysis_key='13C NMR')
product.measurements.add(type='IDENTITY', analysis_key='HRMS')
product.measurements.add(type='YIELD', analysis_key='isolated_weight', percentage=dict(value=93.5, precision=0.5))
product.measurements.add(type='PURITY', analysis_key='quantitative NMR', percentage=dict(value=99))
```

ord_schema/proto/reaction.proto
```diff
@@ -196,6 +196,7 @@ message CrudeComponent {
  * be provided about how it was prepared and from where it was purchased.
  */
 message Compound {
+  reserved 13;
```
skearnes, line 199: Do you want to drop the `reserved` keys and renumber everything? This should be fine if you're copying reaction.proto.

examples/3_Liu_Copper_OrgSyn/example_liu.ipynb
```diff
@@ -772,54 +791,54 @@
     "## Product and characterization\n",
     "outcome = reaction.outcomes.add()\n",
     "outcome.reaction_time.CopyFrom(unit_resolver.resolve('6 h'))\n",
-    "analysis = reaction_pb2.ReactionAnalysis() # for handy enum reference\n",
+    "analysis = reaction_pb2.Analysis() # for handy enum reference\n",
     "\n",
     "outcome.analyses['isolated_weight'].type = analysis.WEIGHT\n",
+    "outcome.analyses['isolated_weight'].is_of_isolated_species = True\n",
     "outcome.analyses['isolated_weight'].details = '27.5-28.0 g recovered after workup'\n",
```
skearnes, line 798: This is interesting ("outcome.analyses['isolated_weight'].details = '27.5-28.0 g recovered after workup"). Do we need better structure in `Analysis` to allow for e.g. a `FloatValue` field?

skearnes (approved):
Tour de force from Connor!

examples/3_Liu_Copper_OrgSyn/example_liu.ipynb
connorcoley: That relates to our earlier conversations about how much structure we want to give analytical results. If we add a FloatValue to analysis, we should also add a Temperature (for melting point), maybe Mass (since this example requires knowing the units, too), etc.

Another way to record this information would be inside the `outcome.analyses['isolated_weight'].data` map I suppose

skearnes: I think I want to avoid using the Data message for this, since we don't have `precision` there.

I'm not sure I want to go down the road of adding additional structured value messages here yet; I think we decided to punt on that for now. We do have a quantitative yield value in the linked ProductMeasurement.

WDYT?

connorcoley: Agreed. The quantitative yield seems sufficient here. There's also the product's `amount` field which has a mass of 27.75 g recorded at the moment; in principle, that should probably have an associated precision of 0.25 g or so

ord_schema/proto/reaction.proto
connorcoley: I'll defer to you and @antonkast-google on protobuf best practices here. I kind of like the reserved keys so there's no risk of reading an old/outdated message, not that the risk is terribly high to begin with. The change that makes me the most nervous are the ones to enums, since they're just passed around as integers in the compiled Python code and it would be very easy to incorrectly migrate those (or forget to).

antonkast-google: Has anyone forked this repo? Do we know of anyone holding on to protos somewhere?

If not, I see no harm either way.

skearnes: I think that since there's no way to make a clean transition from the old to the new (correct me if I'm wrong there; renaming fields breaks `pbtxt` and reusing existing tag numbers breaks `pb`), we should drop the reserved fields and renumber things now so that when things break, they are more likely to break noisily.

Said another way: Using `reserved` and assigning new tag numbers does keep the `pb` files working, but they are not the source of truth---the `pbtxt` files are.

People holding on to old protos can use an earlier version of ord-schema to read them. We'll include details about the break when we release a version with the new changes (and possibly increment the major version to 1?).

connorcoley: SGTM

michaelmaser:
Updates look wonderful. Thanks, Connor!

> ```
> # Define which analyses were used for which aspects of characterization
> product.measurements.add(type='IDENTITY', analysis_key='1H NMR')
> product.measurements.add(type='IDENTITY', analysis_key='13C NMR')
> product.measurements.add(type='IDENTITY', analysis_key='HRMS')
> ```

Would just consider adding the IR analysis as an IDENTITY measurement as well:
`product.measurements.add(type='IDENTITY', analysis_key='thin film IR')`

Can look at notebooks soon.

skearnes:
@michaelmaser any additional comments? It would be great to merge this today.

michaelmaser:
Will take a look today. Sorry for delay, been a busy week.

examples/3_Liu_Copper_OrgSyn/example_liu.ipynb
```diff
@@ -461,112 +461,131 @@
...
       "        }\n",
       "      }\n",
-      "      processed_data {\n",
+      "      data {\n",
       "        key: \"found\"\n",
       "        value {\n",
       "          float_value: 298.1794128417969\n",
       "          description: \"Observed m/z\"\n",
       "        }\n",
       "      }\n",
```
michaelmaser: This is nice too

examples/3_Liu_Copper_OrgSyn/example_liu.ipynb
```diff
@@ -1142,126 +1163,151 @@
...
+      "    }\n",
+      "    measurements {\n",
+      "      analysis_key: \"chiral HPLC\"\n",
+      "      type: SELECTIVITY\n",
+      "      percentage {\n",
+      "        value: 99.0\n",
+      "      }\n",
+      "      selectivity_type {\n",
+      "        type: EE\n",
+      "      }\n",
```
michaelmaser, line 1280: This is great

examples/3_Liu_Copper_OrgSyn/example_liu.ipynb
```diff
@@ -2099,56 +2172,58 @@
...
+    "product.measurements.add(type='IDENTITY', analysis_key='13C NMR')\n",
+    "product.measurements.add(type='IDENTITY', analysis_key='HRMS')\n",
+    "product.measurements.add(type='IDENTITY', analysis_key='thin film IR')\n",
+    "product.measurements.add(type='YIELD', analysis_key='isolated_weight', percentage=dict(value=86))\n",
+    "product.measurements.add(type='PURITY', analysis_key='quantitative NMR', percentage=dict(value=97))\n",
+    "product.measurements.add(type='SELECTIVITY', analysis_key='chiral SFC', percentage=dict(value=90),\n",
+    "                         selectivity_type=dict(type='EE'))\n",
     "\n",
     "product.isolated_color = 'colorless'\n",
     "product.texture.type = product.Texture.OIL\n",
```
michaelmaser, line 2329: LGTM

examples/9_Ahneman_Science_CN_Coupling/example_Ahenman.ipynb
```diff
@@ -369,26 +369,19 @@
...
-    "prod_2a.compound_yield.CopyFrom(\n",
-    "    reaction_pb2.Percentage(value=10.65781182) \n",
-    ")\n",
-    "\n",
     "# The UPLC analysis was used to confirm both identity and yield\n",
-    "prod_2a.analysis_identity.append('UPLC analysis')\n",
-    "prod_2a.analysis_yield.append('UPLC analysis')\n",
+    "prod_2a.measurements.add(type='IDENTITY', analysis_key='UPLC analysis')\n",
+    "prod_2a.measurements.add(type='YIELD', analysis_key='UPLC analysis', percentage=dict(value=10.65781182),\n",
+    "                             uses_internal_standard=True)\n",
```
michaelmaser, line 384: Again, are these calibrated?

examples/9_Ahneman_Science_CN_Coupling/example_Ahenman.ipynb
```diff
@@ -1969,7 +2002,7 @@
        "}"
       ]
      },
-     "execution_count": 20,
+     "execution_count": 21,
      "metadata": {},
      "output_type": "execute_result"
     }
```
michaelmaser, line 2056: LGTM

michaelmaser:
Notebooks look great. Left some comments, but think anything could be addressed later if desired. Otherwise, LGTM.

michaelmaser (approved):

examples/9_Ahneman_Science_CN_Coupling/example_Ahenman.ipynb
connorcoley: That's not completely clear to me. Each distinct product _was_ synthesized separately and used to identify retention times at least. I can't see whether yields were calculated based on a true calibration curve though. I would think they are given that all reported yields are properly between 0 and 100...

ord_schema/proto/reaction.proto
```diff
@@ -861,26 +860,136 @@ message ReactionProduct {
...
+    // Yields should only be reported as quantitative yields, e.g., measured
+    // using an isolated weight analysis, quantitative NMR with an internal
+    // standard, LC peak area with an internal standard and calibration curve
+    // obtained using an authentic standard. 
+    YIELD = 3;
+    // Selectivites include enantiomeric excess/ratios, diastereomic ratios,
+    // etc. The detailed type should be specified in selectivity_type.
+    SELECTIVITY = 4;
+    // The purity of the product mixture, normalized to all other species
+    // observed in the analysis.
```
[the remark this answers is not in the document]
connorcoley: Will fix shortly

ord_schema/proto/reaction.proto
```diff
@@ -861,26 +860,136 @@ message ReactionProduct {
...
-    ZE = 6;  // Alkene geometry ratio. (x:1) (Z / E)
+    // An identity type indicates that the corresponding analysis was used
+    // to confirm the identity of the reported product.
+    IDENTITY = 2;
+    // Yields should only be reported as quantitative yields, e.g., measured
+    // using an isolated weight analysis, quantitative NMR with an internal
+    // standard, LC peak area with an internal standard and calibration curve
+    // obtained using an authentic standard. 
+    YIELD = 3;
+    // Selectivites include enantiomeric excess/ratios, diastereomic ratios,
```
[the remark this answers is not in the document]
connorcoley: Will fix shortly

connorcoley:
@skearnes good to merge?

skearnes:
> @skearnes good to merge?

Please resolve Mike's outstanding comments, and then we're good to go! Thanks for pushing this forward. After the merge, I'll create a new release for ord-schema and start working on the migration for the other repos. I'll work on new branches from `main` so we can all contribute, like we did here.

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
```diff
@@ -5616,45 +5628,47 @@
...
-    "nucleophile.analysis_identity.append('UPLC')\n",
-    "nucleophile.analysis_yield.append('UPLC')"
+    "nucleophile = outcome.products.add(is_desired_product=False) \n",
+    "nucleophile.identifiers.add(value=r'placeholder', type='SMILES')\n",
+    "nucleophile.measurements.add(type='IDENTITY', analysis_key='UPLC', uses_authentic_standard=True)\n",
+    "nucleophile.measurements.add(type='AREA', analysis_key='UPLC', uses_authentic_standard=True)\n",
+    "nucleophile.measurements.add(type='AREA', analysis_key='UPLC', uses_authentic_standard=True,\n",
+    "                             uses_internal_standard=True, is_normalized=True)\n",
+    "\n",
+    "istd = outcome.products.add(is_desired_product=False)\n",
```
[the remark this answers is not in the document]
skearnes: Related question: is there a reason for the asymmetry in how internal and authentic standards are treated? e.g. should we move all the internal standard stuff to ProductMeasurement instead of having it be part of inputs or workups?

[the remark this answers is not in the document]
connorcoley: Missed this comment earlier somehow...

The reason for adding the internal standard as a product is so we can capture its raw LC peak area. It's not my favorite way to encode that information, but I think it's the most general/flexible. This does have the unfortunate side effect of having the internal standard show up in the list of products. We can strip it out by looking for workup components but there's a risk that users would just look at the list of reactants and list of products. Theoretically, we could have a separate list of `repeated ReactionProduct standards` to parallel `repeated ReactionProduct products` in each reaction outcome? 

@skearnes the asymmetry is just because authentic standards aren't (typically) present in the crude reaction mixture itself, whereas internal standards are often added in solution as part of the input or during a workup.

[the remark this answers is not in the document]
skearnes: Got it, thanks. Separating into `standards` and `products` seems like a reasonable workaround. @michaelmaser what do you think?

skearnes:
PS it's fine with me if you want to merge now and address e.g. separating `standards` in subsequent PRs. This PR is getting huge and we can start working on the migration in the meantime...

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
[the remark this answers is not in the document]
michaelmaser: Yes, I think that sounds reasonable. Should work as `get_reaction_smiles` only adds from `outcome.products`

examples/2_Nielsen_Deoxyfluorination_Screen/example_nielsen.ipynb
```diff
@@ -407,30 +407,23 @@
...
-    "prod_2a.compound_yield.CopyFrom(\n",
-    "    reaction_pb2.Percentage(value=40, precision=4.8) \n",
-    ")\n",
-    "\n",
+    "prod_2a.measurements.add(type='YIELD', analysis_key='19f nmr of crude',\n",
+    "                         percentage=dict(value=40, precision=4.8), uses_internal_standard=True)\n",
     "# The \"19f nmr of crude\" analysis was used to confirm both identity and yield\n",
-    "prod_2a.analysis_identity.append('19f nmr of crude')\n",
-    "prod_2a.analysis_yield.append('19f nmr of crude')\n",
+    "prod_2a.measurements.add(type='IDENTITY', analysis_key='19f nmr of crude')\n",
```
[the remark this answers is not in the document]
connorcoley: Because it wasn't specified and I wasn't 100% sure, I left it ambiguous. Do you think that's the right approach?

examples/1_Santanilla_Nanomole_HTE/example_Santanilla.ipynb
```diff
@@ -5616,45 +5628,47 @@
...
-    ")\n",
+    "product.identifiers.add(value=r'placeholder', type='SMILES')\n",
     "product.is_desired_product = True\n",
-    "product.analysis_identity.append('UPLC')\n",
-    "product.analysis_yield.append('UPLC')\n",
+    "\n",
+    "product.measurements.add(type='IDENTITY', analysis_key='UPLC', uses_authentic_standard=True)\n",
+    "product.measurements.add(type='AREA', analysis_key='UPLC', uses_authentic_standard=True)\n",
+    "product.measurements.add(type='AREA', analysis_key='UPLC', uses_authentic_standard=True,\n",
+    "                         uses_internal_standard=True, is_normalized=True)\n",
```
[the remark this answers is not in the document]
connorcoley: Yes, that's the hope. If a ProductMeasurement type is YIELD, then it should be treated as a yield. If it's anything else, then end users should approach it with more caution.

PR #498, opened 2020-11-01
Add ord_client{_test}.py

skearnes:
Initial PR for a python client for the Open Reaction Database.

Currently supports fetching reactions and datasets by ID. Most requests (except for datasets, which can be fetched directly from GitHub) are routed through the ORD search interface located at https://client.open-reaction-database.org. Further PRs will expand the functionality of the client to include the search features of the web interface.

__Question__: Should we interact directly with the Postgres database instead of going through the flask app? I'm leaning "no" since I'm not sure we want to expose the database to the public and allow arbitrary connections that aren't guarded by flask routing...


ord_schema/interface/ord_client.py
```diff
@@ -0,0 +1,132 @@
...
+        Args:
+            dataset_ids: List of dataset IDs.
+
+        Returns:
+            List of Dataset messages.
+        """
+        # NOTE(kearnes): Return datasets in the same order as requested.
+        datasets = {
+            dataset_id: self.fetch_dataset(dataset_id)
+            for dataset_id in set(dataset_ids)
```
connorcoley, line 81: Should this be an ordered set to keep things in order?

ord_schema/interface/ord_client.py
```diff
@@ -0,0 +1,132 @@
...
+
+        Raises:
+            ValueError: A reaction ID is invalid.
+        """
+        for reaction_id in reaction_ids:
+            if not validations.is_valid_reaction_id(reaction_id):
+                raise ValueError(f'Invalid reaction ID: {reaction_id}')
+        response = requests.post(urllib.parse.urljoin(self._target,
+                                                      '/api/fetch_reactions'),
+                                 json=list(set(reaction_ids)))
```
connorcoley, line 113: Likewise: ordered set here?

connorcoley (approved):

ord_schema/interface/ord_client.py
skearnes: Are you thinking of a specific data structure? I'm not aware of an OrderedSet class, and set()/np.unique() return things in sorted order.

connorcoley: Was the intention to sort the results? Based on the comment, I thought your intention was to take return things in the order they were listed in the query.

It doesn't seem like there is a good built-in library with an OrderedSet. There's an `ordered-set` on PyPI but let's not use that. I suppose a hacky solution would be to iterate over `{k:None for k in dataset_ids}.keys()` since dictionaries are ordered. 

I don't think we _have_ to return things in the same query order, but you should revise your comment to say that datasets are returned in a sorted order.

ord_schema/interface/ord_client.py
connorcoley: Never mind, the return state of this function re-sorts things

ord_schema/interface/ord_client.py
connorcoley: Just realized the return statement re-sorts things, never mind.

connorcoley (approved):

connorcoley:
> Question: Should we interact directly with the Postgres database instead of going through the flask app? I'm leaning "no" since I'm not sure we want to expose the database to the public and allow arbitrary connections that aren't guarded by flask routing...

I also vote no

PR #501, opened 2020-11-11
Add helper functions for setting dative bonds to transition metals

michaelmaser:
PTAL. There may be edge cases I'm not thinking of for carbon-metal or other X-M bonds that this wouldn't handle properly. E.g., if carbon is both metal-bound and additionally anionic, the C-M bond would be converted (which may not be desired) since the total valence + formal charge != 4. I don't know of any examples of this, but let me know if there are any other things like this you think of. I will send a notebook of tests for NHC ligands that demonstrate this handling.


skearnes:
I think I forgot about this; sorry. Could you take a look at the failing test?

michaelmaser:
Looks like may be a yapf error. 

From log: 

>2020-11-24T21:59:07.5045751Z Successfully installed yapf-0.30.0
>2020-11-24T21:59:10.8625742Z ##[error]Process completed with exit code 1.
>2020-11-24T21:59:10.8734066Z Post job cleanup.

From an old yapf GitHub issue: 

>A return code of `0` indicates that nothing was formatted and the program didn't have an internal failure.
>A return code of `1` indicates that the program had an internal failure. It's this way because Python apparently returns `1` in such cases.
>A return code of 2 indicates that some files were changed.

Any thoughts on what this may be?

skearnes:
> Looks like may be a yapf error.
> 
> From log:
> 
> > 2020-11-24T21:59:07.5045751Z Successfully installed yapf-0.30.0
> > 2020-11-24T21:59:10.8625742Z ##[error]Process completed with exit code 1.
> > 2020-11-24T21:59:10.8734066Z Post job cleanup.
> 
> From an old yapf GitHub issue:
> 
> > A return code of `0` indicates that nothing was formatted and the program didn't have an internal failure.
> > A return code of `1` indicates that the program had an internal failure. It's this way because Python apparently returns `1` in such cases.
> > A return code of 2 indicates that some files were changed.
> 
> Any thoughts on what this may be?

Did you run the `./format.sh` script on your code? You probably have something that yapf doesn't like.

skearnes:
I ran `./format.sh` locally and there are a bunch of changes in `message_helpers.py`; committing those should fix the problem.

ord_schema/message_helpers.py
```diff
@@ -520,6 +520,88 @@ def set_compound_smiles(compound, value):
...
+    """Determines if an atom is a transition metal.
+    
+    Args:
+        atom: The atom in question. Should be of type rdkit.Chem.rdchem.Atom
+        
+    Returns:
+        Boolean for whether the atom is a transition metal.
+    """
+    n = atom.GetAtomicNum()
+    return (n>=22 and n<=29) or (n>=40 and n<=47) or (n>=72 and n<=79)
```
skearnes, line 533: This can be expressed more succinctly:

`return 22 <= n <= 29 or 40 <= n <= 47 or 72 <= n <= 79`

ord_schema/message_helpers.py
```diff
@@ -520,6 +520,88 @@ def set_compound_smiles(compound, value):
...
+        Boolean for whether the molecule has a transition metal.
+    """
+    for atom in mol.GetAtoms():
+        if is_transition_metal(atom):
+            return True
+    return False
+
+
+def set_dative_bonds(mol, from_atoms=(7,15)):
+    """Convert metal-ligand bonds to dative. Replaces some single bonds between
```
skearnes, line 552: Please make the first line standalone and descriptive (rather than imperative); see https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

```python
"""Converts metal-ligand bonds to dative.

Replaces some ...
"""
```

ord_schema/message_helpers.py
```diff
@@ -520,6 +520,88 @@ def set_compound_smiles(compound, value):
...
+    metals and atoms with atomic numbers in fromAtoms with dative bonds.
+    For all atoms except carbon, the replacement is only done if the atom has
+    "too many" bonds. To handle metal-carbene complexes, metal-carbon bonds
+    are converted to dative if the sum of the explicit and implicit valence
+    of the carbon atom does not equal its default valence, 4.
+
+    Args:
+        mol: The molecule to be converted. 
+        fromAtoms: tuple of atomic numbers corresponding to atom types that
+        should have atom-metal bonds converted to dative. Default is N and P
```
skearnes, line 562: Default is `<the atomic numbers for N and P>`.

Maybe use the atom symbols here instead?

ord_schema/message_helpers.py
```diff
@@ -520,6 +520,88 @@ def set_compound_smiles(compound, value):
...
+    metals = [atom for atom in edit_mol.GetAtoms() 
+              if is_transition_metal(atom)]
+    for metal in metals:
+        for nbr in metal.GetNeighbors():
+            nbr_atom = nbr.GetAtomicNum()
+            # Handles carbon-bound (e.g., NHC-type) ligands
+            # Converts carbon-metal bond to dative if carbon's total valence +
+            # formal charge does not equal 4
+            if nbr_atom in from_atoms and nbr_atom == 6:
+                if nbr.GetTotalValence() + nbr.GetFormalCharge() \
```
skearnes, line 579: Please use parentheses instead of line breaks.

ord_schema/message_helpers.py
```diff
@@ -520,6 +520,88 @@ def set_compound_smiles(compound, value):
...
+                != pt.GetDefaultValence(nbr_atom) and \
+                edit_mol.GetBondBetweenAtoms(nbr.GetIdx(),
```
skearnes, line 581: Can you just update the type of the bond instead of deleting it and adding a new one?

skearnes (changes requested):
Please add a test for these methods in `message_helpers_test.py`.

michaelmaser:
> I ran `./format.sh` locally and there are a bunch of changes in `message_helpers.py`; committing those should fix the problem.

Fixed! Thanks.

ord_schema/message_helpers.py
michaelmaser: Changed!

ord_schema/message_helpers.py
michaelmaser: Done, thanks.

ord_schema/message_helpers.py
michaelmaser: Done. Good idea, thanks!

ord_schema/message_helpers.py
michaelmaser: Done

ord_schema/message_helpers.py
michaelmaser: I believe this can be done with `ReplaceBond`, but requires creating a new `Bond` object and getting the bond index, etc. Can be done, I only worry about the direction (`BondDir`) of the dative bond being correct. Can check with `GetBondDir` or `GetStartAtom`, but ends up being a bit lengthier than what's there now. Is there another reason to use `ReplaceBond` over this?

michaelmaser:
Maybe hold off on merging for now. Want to do some testing to make sure changes are working. 

pylint is unhappy about whitespace. Will fix and test functions tomorrow.

ord_schema/message_helpers.py
connorcoley: I think remove/add is a fine solution -- the code looks nice

skearnes: Ah I didn't appreciate the control over the direction. SGTM.

skearnes:
LGTM; I'll approve after some tests are added.

michaelmaser:
Tests added!

ord_schema/message_helpers_test.py
```diff
@@ -388,6 +388,113 @@ def test_identifier_getters(self):
                          _BENZENE_MOLBLOCK)
 
 
+class SetDativeBondsTest(parameterized.TestCase, absltest.TestCase):
+
+    def test_has_transition_metal(self):
+        xphos = reaction_pb2.Compound()
+        xphos.identifiers.add(type='SMILES',
```
skearnes, line 395: These look pretty good, but a couple of tips on writing tests:

* In general, try to make the test as simple as possible. It should be trivial for a reviewer to verify that the test is correct. For instance, instead of creating a `Compound` message with a complicated SMILES identifier, you could just do `self.assertFalse(message_helpers.has_transition_metal(Chem.MolFromSmiles('P')))`. AFAICT there's no reason to create the proto wrapper or use such a complicated structure for what you're testing here.
* Prefer parameterized tests over listing multiple test cases in the same method. See, for example, https://github.com/Open-Reaction-Database/ord-schema/blob/main/ord_schema/message_helpers_test.py#L95.

michaelmaser: >AFAICT there's no reason to create the proto wrapper or use such a complicated structure for what you're testing here.

I see, thanks. The only reason I added the wrapper was to make sure the method worked within the proto. I had already checked that the function worked in RDKit alone, but I see that's what we're going for in the test here too. Will update.

>Prefer parameterized tests over listing multiple test cases in the same method.

Sounds great. Will work on these now.

michaelmaser: Simplified/parameterized tests. Plan to add a couple more simple ones to `test_set_dative_bonds` tonight or tomorrow

ord_schema/message_helpers_test.py
```diff
@@ -388,6 +388,25 @@ def test_identifier_getters(self):
                          _BENZENE_MOLBLOCK)
 
 
+class SetDativeBondsTest(parameterized.TestCase, absltest.TestCase):
```
skearnes, line 391: Much clearer; thanks!

skearnes (approved):

PR #504, opened 2020-11-14
Add search functionality to ord_client

### LINKED ISSUE #450: Create a python client library
Imagine having something like this:

```python
import ord

client = ord.OrdClient()
# Returns a dataset_pb2.Dataset instance.
dataset = client.fetch(<dataset_id>)
# Returns a reaction_pb2.Reaction instance.
reaction = client.fetch(<reaction_id>)
```

skearnes:
* Replicates the search functionality from https://client.open-reaction-database.org in the Python client.
* Fixes a bug in multi-component queries (use `INTERSECT` instead of `AND`).
* Enables logging of all SQL queries in the `postgres` container.

Resolves #450


ord_schema/interface/ord_client.py
```diff
@@ -130,3 +130,70 @@ def fetch_reaction(self, reaction_id):
...
+            A Dataset containing the matched reactions.
+        """
+        if components is not None:
+            component_params = [
+                component.get_params() for component in components
+            ]
+        else:
+            component_params = None
+        params = {
+            'reaction_ids': ','.join(reaction_ids) if reaction_ids else None,
```
connorcoley, line 161: Should reaction_ids be checked for validity (just via regex matching)?

ord_schema/interface/ord_client_test.py
```diff
@@ -52,14 +53,53 @@ def test_fetch_reaction(self, reaction_id, created_by):
...
+        dataset = self.client.query(reaction_ids=[
+            'ord-f0621fa47ac74fd59f9da027f6d13fc4',
+            'ord-c6fbf2aab30841d198a27068a65a9a98'
+        ])
+        self.assertLen(dataset.reactions, 2)
+
+    def test_query_reaction_smarts(self):
+        dataset = self.client.query(
+            reaction_smarts='FC(F)(F)c1ccc([F,Cl,Br,I])cc1>CS(=O)C>')
+        self.assertLen(dataset.reactions, 862)
```
connorcoley, line 73: Are all of these tests making use of test data? Or the live database? If the latter, it seems unwise to pin our tests to specific numbers of results

skearnes:
Moving to ord-interface.

PR #507, opened 2020-11-20
Add Amount and Source messages

skearnes:
Related to #506 

* Adds an `Amount` message to avoid duplicated `oneof` fields.
  * I chose to include `volume_includes_solutes` in the `Amount` message. This is similar to the other use cases for `Amount` where not all of the fields will be suitable (e.g. moles for `CrudeComponent`).
  * The field is named `aliquot_amount` in `ReactionWorkup`.
* Adds a `Source` message to wrap various vendor metadata.
  * This is a ride-along; let me know if you think we should drop it.

I'll update the tests once we're aligned on the proto changes.


connorcoley (commented):
Schema changes look reasonable to me. I think grouping the compound source in a message is a logical addition as well.

skearnes:
PTAL

ord_schema/validations.py
```diff
@@ -443,19 +443,40 @@ def validate_addition_speed(message):
...
             ' mass or volume specified explicitly', ValidationError)
     if ((not message.HasField('has_derived_amount') or
-         not message.has_derived_amount) and not message.HasField('amount')):
+         not message.has_derived_amount) and
+            not message.amount.HasField('kind')):
         warnings.warn(
             'Crude components should either have a derived amount or'
             ' a specified mass or volume', ValidationError)
+    if (message.amount.WhichOneof('kind') and
+            message.amount.WhichOneof('kind') not in ['mass', 'volume']):
```
connorcoley, line 473: You shouldn't need the first bool `message.amount.WhichOneof('kind')` here

ord_schema/validations.py
```diff
@@ -443,19 +443,40 @@ def validate_addition_speed(message):
...
+        warnings.warn(
+            'Crude component amounts must be specified by mass or volume',
+            ValidationError)
+    if message.amount.HasField('volume_includes_solutes'):
+        warnings.warn(
+            'volume_includes_solutes should only be used for input Compounds')
```
connorcoley, line 479: Missing `ValidationError` argument to `warnings.warn`

connorcoley (approved):

ord_schema/validations.py
skearnes: The value could be `None`; added it to the list to simplify the test.

ord_schema/validations.py
skearnes: Done.

ord_schema/validations.py
connorcoley: Ah, I forgot about the case where we don't require an amount.

PR #512, opened 2020-11-23
Add more docs dependencies

skearnes:
Mostly trying to get all the lexers in place so the RTD build will pass.


connorcoley (approved):

skearnes:
Confirmed that RTD builds are passing now.

PR #513, opened 2020-11-24
Adjustments to message names in ProductMeasurement

skearnes:
Adjusts some message/enum naming to avoid unnecessary verbosity.

@connorcoley @michaelmaser It's not clear to me that `MassSpecMeasurementType`/`MassSpecDetails` should be in `ProductMeasurement` instead of `Analysis`?


connorcoley (approved):
The MS details do make sense in here, since they don't refer to the method of conducting the analysis itself, but rather how a quantitative performance metric was extracted. It's similar to specifying the wavelength at which a UV peak area was quantified in an LC trace: the full LC analysis will contain multiple wavelengths, but you'll generally use a single one when integrating to find peak areas. 

Edit: I suppose we could add a `Wavelength wavelength` to the measurement as well...

skearnes:
Ah ok; so just to be clear, the values in the MassSpecDetails message may be different for measurements based on the same analysis?

connorcoley:
> Ah ok; so just to be clear, the values in the MassSpecDetails message may be different for measurements based on the same analysis?

Yes, exactly

skearnes:
Added a Wavelength message; PTAL.

connorcoley:
LGTM

skearnes:
FYI Also switched `MassSpecDetails` to `MassSpecMeasurementDetails` but kept the field name as `mass_spec_details`.

PR #515, opened 2020-11-28
is_limiting only applies to REACTANT components

skearnes:


connorcoley (approved):
Should this be added to `validations.py`? If a compound has a non-REACTANT role and `is_limiting` is defined, raise an error? 

I don't think this is necessarily error-worthy if `is_limiting = False`. Was your intention mostly to clean up the example?

skearnes:
> Should this be added to `validations.py`? If a compound has a non-REACTANT role and `is_limiting` is defined, raise an error?
> 
> I don't think this is necessarily error-worthy if `is_limiting = False`. Was your intention mostly to clean up the example?

Yes, this is mostly to clean up the example since the editor only outputs `is_limiting` for `REACTANT`s and the round-trip tests were failing.

PR #520, opened 2020-12-03
Move ord_schema/interface to ord-interface

skearnes:
The ORD interface has a new home: https://github.com/Open-Reaction-Database/ord-interface


connorcoley (approved):
Hooray for fast tests

PR #533, opened 2021-01-08
Add helper functions for converting protos into a DataFrame

skearnes:
This should make it significantly easier to move from a `Dataset` or list of `Reaction`s into a more familiar python environment and extract the fields needed for downstream tasks.

@marcosfelt FYI


ord_schema/message_helpers.py
```diff
@@ -782,3 +787,96 @@ def create_message(message_name):
...
+        trace = tuple()
+    row = {}
+    for field, value in message.ListFields():
+        if field.label == field.LABEL_REPEATED:
+            if (field.type == field.TYPE_MESSAGE and
+                    field.message_type.GetOptions().map_entry):
+                value_field = field.message_type.fields_by_name['value']
+                for key, subvalue in value.items():
+                    this_trace = trace + (f'{field.name}["{key}"]',)
+                    row.update(
```
connorcoley, line 847: Should we explicitly check for whether a field in the dict will be overwritten? Based on your naming system, I don't see why it would be, but wanted to throw that out as an idea.

connorcoley (approved):
I like the typing!

ord_schema/message_helpers_test.py
```diff
@@ -482,5 +485,105 @@ def test_invalid_messages(self, message_name):
...
+                 'children["a"].value': 1.2,
+                 'children["b"].value': 3.4
+             }),
+    )
+    def test_message_to_row(self, message, expected):
+        row = message_helpers.message_to_row(message)
+        pd.testing.assert_frame_equal(pd.DataFrame([row]),
+                                      pd.DataFrame([expected]),
+                                      check_like=True)
+
```
michaelmaser, line 544: These look great

michaelmaser (approved):

michaelmaser:
Quick question/test: Is this mainly meant for, e.g., HTE datasets with a single reaction type & setup where the messages have identical fields just with changing values? I'm just wondering what the dataframe would look like if there was variability between rows. 

Example: one substrate in a screen requires addition of an extra input (this happens in Santanilla); does a full column get created for that input where all rows are `NaN` besides those with that substrate? Or, is there just one column with for repeated inputs (as lists or nested), where all rows with that substrate would just have an extra entry?

I'm guessing this is handled by the tracing, just haven't thought through what these scenarios would give.

ord_schema/message_helpers.py
skearnes: Yes, this is a good check. Done.

ord_schema/message_helpers_test.py
skearnes: Thanks!

skearnes:
> Quick question/test: Is this mainly meant for, e.g., HTE datasets with a single reaction type & setup where the messages have identical fields just with changing values? I'm just wondering what the dataframe would look like if there was variability between rows.
> 
> Example: one substrate in a screen requires addition of an extra input (this happens in Santanilla); does a full column get created for that input where all rows are `NaN` besides those with that substrate? Or, is there just one column with for repeated inputs (as lists or nested), where all rows with that substrate would just have an extra entry?
> 
> I'm guessing this is handled by the tracing, just haven't thought through what these scenarios would give.

Good question! For an extra input with a new key, a new column would be created that would be empty for the other reactions (since that input would have a unique trace that becomes the column name).

Note that this function is general: you can pass any type of proto to it; it doesn't even require that the protos are all of the same type.

skearnes:
> I like the typing!

Yes, I've started using this in most of my new code. It's great stuff.

PR #534, opened 2021-01-09
Add documentation for dataset templating

### LINKED ISSUE #531: Documentation for dataset templating
I don't think it's documented anywhere (except in the tutorial video) how to edit the pbtxt for templating. Hopefully we will make this obsolete with https://github.com/open-reaction-database/ord-editor/issues/63

skearnes:
Resolves #531

### DIFF
--- a/docs/guides/templates.rst
+++ b/docs/guides/templates.rst
@@ -38,14 +38,21 @@ column in the data spreadsheet and start and end with ``$``. For instance:
 .. code-block::
 
   ...
-  inputs {
-  key: "alcohol in THF"
-  value {
-    components {
-      identifiers {
-        type: SMILES
-        value: "C"  <-- Change "C" to "$alcohol_smiles$".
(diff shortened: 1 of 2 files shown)


connorcoley (approved):
Thanks for putting this together. 

As one suggestion, it would be useful if Step 2 mentioned that values that are strings in quotation marks should _keep_ those quotation marks (which your example does), but that other values (like the float yield in the example) don't need quotation marks.

Do you want to add a note about how empty components or 0 amounts are handled?

skearnes:
> Thanks for putting this together.
> 
> As one suggestion, it would be useful if Step 2 mentioned that values that are strings in quotation marks should _keep_ those quotation marks (which your example does), but that other values (like the float yield in the example) don't need quotation marks.

Done.
> 
> Do you want to add a note about how empty components or 0 amounts are handled?

Done.

PR #541, opened 2021-01-21
Add type annotations

skearnes:
Adds type annotations using the `typing` module to all non-test code. See https://google.github.io/styleguide/pyguide.html#319-type-annotations.

Also:
* Removes some scripts in `ord_schema/proto` that are no longer used (so I don't have to annotate them).
* Moves `process_datasets{_test}.py` to `ord_schema/scripts` (since it is a script).
* Factors out commonly used types into `ord_schema/__init__.py`.
* Adds a test to the CI for pytype.


connorcoley (approved):
Did a semi-thorough set of spot checks and it looks good to me. Thanks a ton for doing this

PR #547, opened 2021-01-26
Script for parsing USPTO reaction data

skearnes:
The script, intentionally, does not capture everything. The CML is complicated and if we were starting from scratch I would just rewrite the original parsing code to output ORD records rather than trying to parse the CML.


skearnes:
Hi @connorcoley, I made some changes after our discussion yesterday:

* Added `REACTION_CXSMILES` identifier type.
* Made `analysis_key` optional (warning instead of error).

I chose not to try to parse pH or stirring into `ReactionConditions`: (1) some pH values are ranges and I don't really want to wrangle those right now; (2) `ReactionWorkup` better captures the CML fields like `duration` that we don't have in `ReactionConditions` (and it would be a pain to try to forward that information to the `ReactionOutcome`).

ord_schema/units.py
```diff
@@ -146,27 +160,49 @@ def __init__(self,
...
+            if '±' in string:
+                value = float(value)
+                precision = float(range_value)
+            elif not allow_range:
+                raise ValueError('string appears to contain a range of values '
+                                 f'but allow_range is False: {string}')
+            else:
+                values = np.asarray([value, range_value], dtype=float)
+                value = values.mean()
+                precision = values.std()
```
connorcoley, line 203: Should we use the standard devision here or `max - min`? I think the only difference is a factor of $\sqrt{2}$, but my inclination is that a value of "70-80% yield" should be transcribed as $70 \pm 5$.

connorcoley (commented):
Thanks for those additions. Like we talked about, I think this is terrific and it'll be really nice to get this data in the repo.

I'm very conflicted about having stirring/temperature/duration be listed as workup steps instead of reaction conditions. On one hand, I understand that it can be ambiguous in this dataset. On the other hand, what value is this information providing in the workup list? How would that ever be used downstream? 

I know you say it would be a pain to forward duration to `ReactionOutcome`, but what would you think of the following post-processing steps:
- Process the reaction as you have it
- After the reaction is defined, iterate through the workup steps as long as the action is of type HEAT, STIR, or WAIT and it is the first time we are encountering those actions.
- Assume that in this case, the HEAT can be recorded as the temperature in `ReactionConditions`; STIR can be recorded as the stirring conditions; WAIT can be recorded as the `time` in `ReactionOutcome`. I guess in the cml schema, one STIR action could contain the temperature and duration information as attributes? So this might not work exactly.
- Add `dynamic_conditions=True` to all records since we can't guarantee we have accurately captured everything (note: should probably do this anyway)

ord_schema/units.py
skearnes: Just to clarify:

* Do you mean `75 \pm 5`?
* `np.std([70, 80]) == 5.0`; so it might do what you want already?

connorcoley: It would look that way. I guess `ddof=0` by default so never mind!

skearnes:
PTAL; it's not perfect and the CML doesn't always put the right values first.

* Takes `reaction_time` from `STIRRING` or `WAIT` workups.
* Takes `temperature` from `STIRRING` or `TEMPERATURE` workups.
* Takes `stirring` from `STIRRING` workups.

skearnes:
@connorcoley here's the gist I promised: https://gist.github.com/skearnes/c8426fb607eec55892e2508153241961

skearnes:
> @connorcoley here's the gist I promised: https://gist.github.com/skearnes/c8426fb607eec55892e2508153241961

Scratch that; here's a zipfile.
[test.zip](https://github.com/open-reaction-database/ord-schema/files/5921863/test.zip)

connorcoley:
Looking at more of these examples, the CML parsing is quite imperfect. There are also more "dynamic conditions" than I expected. Nevertheless, this seems like a very good first pass at adapting the CML schema to the ORD schema. We're not trying to improve parsing of the original text paragraph, so I think this looks good for now. We should approach this data with two stages of disclaimers: the text -> CML is imperfect, and the CML -> ORD is also imperfect

connorcoley (approved):

connorcoley:
I wonder if there would be any value in hand-annotating some of these text paragraphs. It looks like it would be very fast to go from the paragraph to an ORD entry...probably ≤10 minutes per example. Do you think that it's worth trying to train an information extraction pipeline that goes from patent paragraph to ORD schema entry?

skearnes:
> I wonder if there would be any value in hand-annotating some of these text paragraphs. It looks like it would be very fast to go from the paragraph to an ORD entry...probably an average of <5 minutes per example. Do you think that it's worth trying to train an information extraction pipeline that goes from patent paragraph to ORD schema entry?

I think that could be a fun side project, but I don't think this is blocking for any short-term downstream uses so I wouldn't want to go too deep down that hole personally...

PR #551, opened 2021-02-09
Add support for reading/writing gzipped messages

skearnes:


michaelmaser (approved):
Re comment on #552, I thought we decided not to gzip for simplicity? Did I misinterpret? I suppose supporting doesn't hurt either way.

skearnes:
> Re comment on #552, I thought we decided not to gzip for simplicity? Did I misinterpret? I suppose supporting doesn't hurt either way.

Yes, that's what we decided before I started looking at the Git LFS bandwidth requirements---we get "charged" for every download/clone of the repo, so reducing the size on Git LFS is actually pretty important. Hope that's ok to change?

michaelmaser:
Gotcha, SGTM

PR #552, opened 2021-02-10
Adjust process_dataset to avoid OOM errors

skearnes:
GitHub Actions is running out of memory when it tries to load all of the USPTO grant data in at the same time.


connorcoley (approved):

skearnes:
Thanks; waiting on #551 so I can make process_dataset output .pb.gz instead of .pb.

skearnes:
FYI added a new flag `--output_format` that defaults to `'.pb.gz'`.

PR #559, opened 2021-02-16
Allow reactions to pass validations if they only have reaction SMILES

skearnes:
* Adds a new option to ValidationOptions: `allow_reaction_smiles_only`
* When this is true (the default), reactions that have a `REACTION_{CX}SMILES` identifier, zero inputs, and zero outcomes will not be required to have inputs or outcomes. The other checks for e.g. provenance still apply.

### DIFF
--- a/ord_schema/validations.py
+++ b/ord_schema/validations.py
@@ -447,9 +447,7 @@ def validate_reaction(message: reaction_pb2.Reaction,
 
 def validate_reaction_identifier(message: reaction_pb2.ReactionIdentifier):
     check_type_and_details(message)
-    if message.type in [
-            message.REACTION_SMILES, message.REACTION_CXSMILES
-    ]:
+    if message.type in [message.REACTION_SMILES, message.REACTION_CXSMILES]:
         if message.type == message.REACTION_CXSMILES:
             smiles = message.value.split()[0]
         else:
--- a/ord_schema/validations.py
+++ b/ord_schema/validations.py
@@ -24,6 +24,7 @@ from absl import logging
 from dateutil import parser
 from rdkit import Chem
 from rdkit import __version__ as RDKIT_VERSION
+from rdkit.Chem import rdChemReactions
 
 import ord_schema
 from ord_schema import message_helpers
@@ -41,7 +42,7 @@ class ValidationOptions:
     # Require ReactionProvenance for Reactions.
     require_provenance: bool = False
     # Allow reactions with valid reaction SMILES and nothing else.
-    allow_reaction_smiles_only = True
+    allow_reaction_smiles_only: bool = True
 
 
 @dataclasses.dataclass
@@ -446,6 +447,22 @@ def validate_reaction(message: reaction_pb2.Reaction,
 
 def validate_reaction_identifier(message: reaction_pb2.ReactionIdentifier):
     check_type_and_details(message)
+    if Chem and message.type in [
+            message.REACTION_SMILES, message.REACTION_CXSMILES
+    ]:
+        if message.type == message.REACTION_CXSMILES:
+            smiles = message.value.split()[0]
+        else:
+            smiles = message.value
+        try:
+            rxn = rdChemReactions.ReactionFromSmarts(smiles, useSmiles=True)
+            if rxn is None:
+                warnings.warn(
+                    f'RDKit {RDKIT_VERSION} could not validate'
+                    f' REACTION_{{CX}}SMILES identifier {message.value}',
+                    ValidationError)
+        except ValueError as error:
+            warnings.warn(str(error), ValidationError)
     if not message.value:
         warnings.warn('value must be set', ValidationError)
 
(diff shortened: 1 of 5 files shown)


ord_schema/validations.py
```diff
@@ -40,6 +40,8 @@ class ValidationOptions:
     validate_ids: bool = False
     # Require ReactionProvenance for Reactions.
     require_provenance: bool = False
+    # Allow reactions with valid reaction SMILES and nothing else.
+    allow_reaction_smiles_only = True
```
connorcoley, line 44: For consistency, use typing here as well?

ord_schema/validations.py
```diff
@@ -410,12 +412,17 @@ def validate_reaction(message: reaction_pb2.Reaction,
...
     if options is None:
         options = ValidationOptions()
-    if len(message.inputs) == 0:
-        warnings.warn('Reactions should have at least 1 reaction input',
-                      ValidationError)
-    if len(message.outcomes) == 0:
-        warnings.warn('Reactions should have at least 1 reaction outcome',
-                      ValidationError)
+    if (options.allow_reaction_smiles_only and
+            message_helpers.get_reaction_smiles(message) and
```
connorcoley, line 416: I think we should try to parse the reactants, any agents, and product(s) into RDKit and issue ValidationWarnings if it doesn't parse. Like we do for SMILES in check_compound_identifiers?

ord_schema/validations_test.py
```diff
@@ -175,6 +175,19 @@ def test_reaction(self):
                                     'reaction input'):
             self._run_validation(message)
 
+    def test_reaction_smiles(self):
+        message = reaction_pb2.Reaction()
+        message.identifiers.add(value='test', type='REACTION_SMILES')
+        output = self._run_validation(message)
+        self.assertEmpty(output.errors)
+        self.assertEmpty(output.warnings)
```
connorcoley, line 183: If adding a Reaction SMILES value check, this warning list will not be empty

ord_schema/validations.py
skearnes: ah thanks; done

ord_schema/validations.py
skearnes: Right now this branch only executes if there are no reactants/agents or products at all. I'm thinking that I should add a check for valid reaction SMILES and leave this as-is? If there are reactants or products they will be checked in the normal way.

ord_schema/validations_test.py
skearnes: yes, i'll add this check

ord_schema/validations.py
connorcoley: Yes, good idea

skearnes:
Thanks; PTAL.

ord_schema/validations.py
```diff
@@ -446,6 +447,22 @@ def validate_reaction(message: reaction_pb2.Reaction,
 
 def validate_reaction_identifier(message: reaction_pb2.ReactionIdentifier):
     check_type_and_details(message)
+    if Chem and message.type in [
```
connorcoley, line 450: Under what circumstances will `Chem` be `None`?

ord_schema/validations.py
```diff
@@ -446,6 +447,22 @@ def validate_reaction(message: reaction_pb2.Reaction,
...
+            smiles = message.value.split()[0]
+        else:
+            smiles = message.value
+        try:
+            rxn = rdChemReactions.ReactionFromSmarts(smiles, useSmiles=True)
+            if rxn is None:
+                warnings.warn(
+                    f'RDKit {RDKIT_VERSION} could not validate'
+                    f' REACTION_{{CX}}SMILES identifier {message.value}',
+                    ValidationError)
```
connorcoley, line 463: For consistency with `validate_compound`, which uses `message_helpers.check_compound_identifiers`, let's make this a ValidationWarning not a ValidationError

connorcoley (approved):

ord_schema/validations.py
skearnes: sorry; just copied from validate_compound_identifier. We made rdkit a formal dep, so I'll remove that.

ord_schema/validations.py
skearnes: done

skearnes: Actually, we use warning for NAME and error for SMILES (in validate_compound_identifier). The check_compound_identifiers check is less strict, but that's more designed for checking consistency than parse-ability...WDYT?

skearnes: (Just to be clear, the current code will issue a ValidationError if the SMILES can't be parsed for a compound.)

connorcoley: Oh, sorry you're right -- I was going off of name and forgot that was only for a consistency check. In that case, let's stick to ValidationError as you have it

PR #562, opened 2021-02-18
Updates to reaction rendering

skearnes:
* Only shows `PRODUCT` and `UNSPECIFIED` roles on the right side of the arrow in "compact" mode.
* Only shows `REACTANT` and `UNSPECIFIED` roles on the left side of the arrow in "compact" mode.
* Removes the reaction ID from the rendering (this should be handled separately).

Normal:
image

Compact:
image

### DIFF
--- a/ord_schema/visualization/filters.py
+++ b/ord_schema/visualization/filters.py
@@ -647,7 +647,10 @@ def _get_compact_components(
         component: Compound message.
         is_last: Whether this is the last compound that will be displayed.
     """
-    roles_to_keep = [reaction_pb2.ReactionRole.REACTANT]
+    roles_to_keep = [
+        reaction_pb2.ReactionRole.REACTANT,
+        reaction_pb2.ReactionRole.UNSPECIFIED,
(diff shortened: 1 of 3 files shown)


ord_schema/visualization/filters.py
```diff
@@ -660,12 +660,24 @@ def _get_compact_components(
             yield compound, False
 
 
+def _get_compact_products(
+    products: Iterable[reaction_pb2.ProductCompound]
+) -> List[reaction_pb2.ProductCompound]:
+    """Returns a list of product compounds for 'compact' visualization."""
+    roles_to_keep = [reaction_pb2.ReactionRole.PRODUCT]
```
connorcoley, line 667: Can you add UNSPECIFIED as a role to keep?

connorcoley (approved):

ord_schema/visualization/filters.py
skearnes: Done for inputs and outputs.

PR #566, opened 2021-02-19
Do not resolve NAME identifiers during submission

skearnes:
We're hitting the resolution servers millions of times in https://github.com/open-reaction-database/ord-data/pull/59 (if I hadn't stopped the job early).

### DIFF
--- a/ord_schema/updates.py
+++ b/ord_schema/updates.py
@@ -18,7 +18,6 @@ import re
 from typing import Mapping
 import uuid
 
-from ord_schema import resolvers
 from ord_schema.proto import dataset_pb2
 from ord_schema.proto import reaction_pb2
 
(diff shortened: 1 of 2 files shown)


connorcoley:
Aside from the flood of requests for this giant PR, are you opposed to using the name resolver upon submission in general?

skearnes:
> Aside from the flood of requests for this giant PR, are you opposed to using the name resolver upon submission in general?

I guess I don't feel strongly about that. It might be better to change the validations so that we warn when there are only NAME identifiers (that might be there already?) rather than actually adding the resolved SMILES during submission? I guess I'm thinking that we shouldn't be doing work that we should be pushing onto the submitters...

skearnes:
PS part of my thinking here, besides the USPTO submission, is around whether contributors can verify changes. We do this resolution in the second phase of the submission where the contributor is not usually involved, and I think I'd prefer that these resolutions be done by the contributors themselves in the first phase instead.

connorcoley:
> PS part of my thinking here, besides the USPTO submission, is around whether contributors can verify changes. We do this resolution in the second phase of the submission where the contributor is not usually involved, and I think I'd prefer that these resolutions be done by the contributors themselves in the first phase instead.

This is convincing for me

connorcoley (approved):

PR #570, opened 2021-03-05
Update list_dois.py to handle multiple suffixes

skearnes:
The current script doesn't correctly handle `.pb.gz` files.

### DIFF
--- a/ord_schema/scripts/list_dois.py
+++ b/ord_schema/scripts/list_dois.py
@@ -43,11 +43,15 @@ def main(argv):
     filenames = glob.glob(FLAGS.input, recursive=True)
     logging.info('Found %d datasets', len(filenames))
     dois = collections.defaultdict(list)
+    output_filenames = {}
     for filename in filenames:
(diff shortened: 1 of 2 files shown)


ord_schema/scripts/list_dois.py
```diff
@@ -43,11 +43,15 @@ def main(argv):
     filenames = glob.glob(FLAGS.input, recursive=True)
     logging.info('Found %d datasets', len(filenames))
     dois = collections.defaultdict(list)
+    output_filenames = {}
     for filename in filenames:
         logging.info('Checking %s', filename)
         dataset = message_helpers.load_message(filename, dataset_pb2.Dataset)
-        dataset_id = os.path.splitext(os.path.basename(filename))[0]
-        assert dataset.dataset_id == dataset_id
+        dataset_id = os.path.basename(filename).split('.')[0]
```
brilee, line 50: This snippet is possibly worth splitting out as a helper with the explicit test cases you want to cover

brilee (approved):

PR #577, opened 2021-03-19
Add Binder support for example notebooks

skearnes:
Here's a version of the Binder button that will work for this PR (the one in README.md points to `main`): [Binder](https://mybinder.org/v2/gh/skearnes/ord-schema/binder?filepath=examples)


connorcoley (approved):
Nice, this looks like a much cleaner way to do things. It seems like Binder has enough support that it's an okay medium-term solution (i.e., likely to stay free)?

skearnes:
> Nice, this looks like a much cleaner way to do things. It seems like Binder has enough support that it's an okay medium-term solution (i.e., likely to stay free)?

Yes, I hope it will stay that way...

PR #583, opened 2021-03-30
Add PROTEIN_SEQUENCE compound identifier type

### LINKED ISSUE #554: Add CompoundIdentifier for amino acid sequence

skearnes:
Resolves #554 

I went back and forth on things like `SEQUENCE`, `FASTA`, etc. but I think this probably hits the mark pretty well. I'd also be fine with `AMINO_ACID_SEQUENCE`. WDYT?

### DIFF
--- a/ord_schema/proto/reaction.proto
+++ b/ord_schema/proto/reaction.proto
@@ -333,7 +333,9 @@ message CompoundIdentifier {
     // Protein data bank ID (for enzymes)
     PDB_ID = 14;
     // Amino acid sequence (for enzymes).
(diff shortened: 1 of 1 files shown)


connorcoley:
+1 for `AMINO_ACID_SEQUENCE` since `PROTEIN_SEQUENCE` implicitly excludes shorter polypeptides that aren't proteins per se

While we're making this change, should we add `HELM` for HELM notation as well? 

Thanks for doing this

connorcoley (approved):

PR #591, opened 2021-04-20
Updates to the reaction landing page

skearnes:
* Raw protos are now shown in tooltips
* Use bootstrap everywhere

### DIFF
--- a/ord_schema/visualization/reaction.html
+++ b/ord_schema/visualization/reaction.html
@@ -23,7 +23,6 @@
         integrity="sha384-eOJMYsd53ii+scO/bJGFsiCZc+5NDVN2yr8+0RDqr0Ql0h+rP48ckxlpbzKgwra6" crossorigin="anonymous">
   <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.4.1/font/bootstrap-icons.css">
   <style>
-
       body {
           padding: 10px 20px;
       }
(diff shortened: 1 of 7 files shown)


ord_schema/visualization/reaction.html
```diff
@@ -797,10 +850,10 @@ <h4>Measurements</h4>
                         <td>{{ measurement|oneof('value')|product_measurement_value }}</td>
```
connorcoley, line 850: This line (L850) is giving me an error when I try rendering first reaction from my Biginelli dataset:

[biginelli.pb.zip](https://github.com/open-reaction-database/ord-schema/files/6369636/biginelli.pb.zip)

`TypeError: getattr(): attribute name must be string`

This happens because `message.WhichOneof(name)` is `None` for `name == 'value'` in the `_oneof` filter function. i.e., there are `IDENTITY` measurements without a value defined

skearnes: Thanks; fixed.

skearnes:
@connorcoley PTAL

connorcoley (dismissed):
While I am still having trouble copying any text from the tooltips (e.g., SMILES strings from the raw proto text), everything else looks great

connorcoley:
- [ ] When looking at a dataset or reaction page, the user information in the upper-right corner is lost depending on the window width
- [ ] "programmatic interface" link on the homepage under the "Enumerate" tab is missing a space after the link
- [ ] When a reaction is loaded with a volume field, the "volume includes solutes" text is not shown. However, it is correctly shown onchange
- [ ] The boolean selector for "includes solutes" isn't aligned pixel-perfect, but that's okay probably
- [ ] The cancel button on Ketcher doesn't work (i.e., doesn't leave the drawing)
- [ ] The "Done drawing" button on Ketcher adds the identifier but doesn't close/hide the window
- [ ] Specification of the automation platform does not seem to be saved
- [ ] Nitpicky, but can we change the "add pressure" button to "add pressure conditions"? Likewise for electrochemistry and flow. Having the slightly longer field will help them stand out from "add measurement". Making these longer might cause them to run onto multiple lines when loading a blank reaction, however.

connorcoley (approved):

PR #601, opened 2021-06-07
Add install_requires to setup.py

### LINKED ISSUE #600: Setup.py is missing install_requires
setup.py is missing `install_requires`, so dependencies are not installed when installing from pypi.  Happy to submit a PR for this if it is indeed a bug.

marcosfelt:
This fixes #600 by making sure dependencies are installed when the package is installed from pypi or source via pip. This avoids having to clone the whole repository in order to get the package installed.


marcosfelt:
Any updates on this @skearnes

skearnes (approved):

skearnes:
Thanks @marcosfelt; sorry for the delay!

skearnes:
@marcosfelt I think you need to run the format.sh script before this can be merged, so the check_python test will pass.

dcambie:
> @marcosfelt I think you need to run the format.sh script before this can be merged, so the check_python test will pass.

In this case this should be enough ;)
```console
pip install yapf
yapf -i .\setup.py
```

PR #608, opened 2021-07-16
Update list_dois.py

### LINKED ISSUE #606: In auto-generated DOI list, show dataset description
This will clarify where contributions came from in terms of _specific_ figures/tables as subsets of papers, which should help us detect duplicate submissions.

skearnes:
Resolves #606; see https://github.com/open-reaction-database/ord-data/pull/111 for example output.

NOTE: I ended up in a crazy dependency knot and had to move to `rdkit-pypi` to get the numpy version to stay above `1.20.0`.


skearnes:
@connorcoley PTAL

PR #611, opened 2021-09-25
Show dataset ID in reaction landing page breadcrumbs

skearnes:


ord_schema/resolvers.py
```diff
@@ -122,8 +123,9 @@ def _cactus_resolve(value_type: str, value: str) -> str:
 def _emolecules_resolve(value_type: str, value: str) -> str:
     """Resolves compound identifiers to SMILES via the eMolecules API."""
     del value_type  # Unused.
-    with urllib.request.urlopen('https://www.emolecules.com/lookup?q={}'.format(
-            urllib.parse.quote(value))) as response:
+    with urllib.request.urlopen(
+            os.path.join('https://www.emolecules.com'
```
connorcoley, line 127: Do you mind explaining the use of `os.path.join` here?

connorcoley (approved):

ord_schema/resolvers.py
skearnes: I was just using it to break up the URL across multiple lines; replaced with urllib.parse.urljoin.

PR #628, opened 2022-05-30
Replace absl/absltest with docopt/pytest

skearnes:
* Replace absl with docopt
* Replace absltest with pytest
* Delete ord_schema/visualizations (now part of ord-interface)
* Remove ord_schema/scripts/migrate_schema.py and *old.proto


brilee (approved):
LGTM, skimmed through the python files and I don't see any erroneous reformats or test replacements

PR #630, opened 2022-06-12
Push to PyPI after every merge

### LINKED ISSUE #629: pip install ord-schema

skearnes:
* Moves `requirements.txt` into setup.py for automatic handling with pip.
* Replaces conda rdkit with rdkit-pypi to allow for pip installation. Note that the rdkit installation is optional to allow for users to use conda rdkit instead.
* Updates the patch version on every merge to main and pushes to PyPI.

Resolves #629


skearnes:
@connorcoley: should we create a conda package instead of a pypi package to preserve the conda rdkit installation path? I'm a bit concerned about possible conflicts between conda rdkit and rdkit-pypi if we expect most people are still following the recommended installation path; see https://github.com/rdkit/rdkit/issues/1812#issuecomment-1153095189.

skearnes:
> @connorcoley @ipendlet after some digging I find myself thinking that we should provide both conda and pypi packages. That said, if we have to pick one to start with, conda might be a better path since that will avoid issues like this one: [kuelumbus/rdkit-pypi#48](https://github.com/kuelumbus/rdkit-pypi/issues/48). WDYT?

@connorcoley @ipendlet I think this question might become moot if `rdkit-pypi` is renamed to `rdkit` though.

skearnes:
@connorcoley @ipendlet I thought of a workaround for the conda/pip issues prior to any renaming. I'll use the extras_require logic to only install rdkit-pypi if explicitly requested and update the readme to indicate that you should only use that if you haven't installed rdkit already via conda.

ipendlet (approved):

PR #631, opened 2022-06-14
Include compiled proto wrappers

skearnes:
* Removes the need for protoc at install time.
* Checks to prevent the proto definitions and the compiled files from getting out of sync.
* Tested on M1 mac, Intel mac, and Linux.


ipendlet (approved):
Discussed changes in person with @skearnes .  To avoid the need to compile on install we are now including protocol bugger generated code directly.  To regenerate you now use the script and not the install.

PR #641, opened 2022-06-24
Use python 3.9

skearnes:


brilee (approved):
LGTM. There's some guidance at Google about how you ought to use various incantations of Iterable or AbstractX to annotate args/return values, but I think the list stuff is probably the right choice for simplicity

ord-service:
> LGTM. There's some guidance at Google about how you ought to use various incantations of Iterable or AbstractX to annotate args/return values, but I think the list stuff is probably the right choice for simplicity

Thanks; that guidance is in the [typing docs](https://docs.python.org/3/library/typing.html#typing.List) as well. Updated most of the args to match.

PR #653, opened 2022-11-01
Add new protobuf format for the new database

FanwangM:
Also updated CI configurations.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -44,7 +44,7 @@ jobs:
           coverage erase
           pytest --cov=ord_schema
           coverage xml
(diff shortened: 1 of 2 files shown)


ipendlet:
I'm not sure what reaction_new.proto in this commit is doing. Can you explain?

FanwangM:
> I'm not sure what reaction_new.proto in this commit is doing. Can you explain?

Sure. The intention is to have a second layer of the simplified and cleaned database, which we can train the machine learning models with. That's why we need to have a simplified version of the original ProtoBuf format. @ipendlet

connorcoley:
Moving to separate repo

PR #655, opened 2022-11-26
ORM query helpers

skearnes:
* Move base classes to a new `base.py` module
* Add `query{_test}.py` for defining ORM queries with fewer explicit joins


skearnes:
I'm reworking the mappers to use programmatic class definitions and single-table inheritance; closing this for now since it will mostly be obsolete.

PR #660, opened 2022-12-02
update protobuf version

skearnes:


skearnes:
Looks like the reason we pin to <3.20 is for compatibility with TensorFlow: https://github.com/tensorflow/tensorflow/blob/b59935d3b349e752eed4729606e5d64a17e48015/tensorflow/tools/pip_package/setup.py#L105

skearnes:
> Looks like the reason we pin to <3.20 is for compatibility with TensorFlow: https://github.com/tensorflow/tensorflow/blob/b59935d3b349e752eed4729606e5d64a17e48015/tensorflow/tools/pip_package/setup.py#L105

FYI @connorcoley @ipendlet @FanwangW we use TF in some of the examples

connorcoley:
Is this a reason to separate out the examples into their own repo? Seems like a complex dependency to have when it's not essential for many use cases

skearnes:
> Is this a reason (or an excuse) to separate out the examples into their own repo? Seems like a complex dependency to have when it's not essential for many use cases

That wouldn't fix the issue; if until TF supports protobuf>=3.20 we won't be able to use TF and ord-schema in the same interpreter since both require the proto libraries. I think our choices are (1) pin to whatever version of protobuf TF supports so that people can use TF, or (2) upgrade protobuf and declare that ord-schema cannot be used with TF.

FanwangM:
I think it's not worth it to lose the support of TF because of version updating. We can just keep the current version and wait for TF to support `protobug>=3.20`.

PR #666, opened 2023-01-25
proto updates

skearnes:
Updates to reaction.proto:
* `ReactionIdentifier.NAME` has been renamed to `REACTION_TYPE`
* `Vessel.plate_id` has been renamed to `vessel_id`
* `Vessel.plate_position` has been renamed to `position`
* Added `Vessel.row` and `Vessel.column` string fields
* Added `ReactionProvenance.reaction_metadata` map from strings to `Data` messages


ipendlet (approved):

skearnes:
re: https://github.com/open-reaction-database/ord-schema/pull/669#issuecomment-1501195186 (@ipendlet)

This is one of the reasons we use binary encodings for the source of truth; as long as we don't change the tag numbers we can rename fields without breaking the encoding.

skearnes:
> re: [#669 (comment)](https://github.com/open-reaction-database/ord-schema/pull/669#issuecomment-1501195186) (@ipendlet)
> 
> This is one of the reasons we use binary encodings for the source of truth; as long as we don't change the tag numbers we can rename fields without breaking the encoding.

I've also verified by running `ord_schema/scripts/validate_dataset.py` on several datasets in ord-data.

PR #669, opened 2023-02-21
Better ORM context information

skearnes:
* Better column names: `_polymorphic_type` is now `ord_schema_context` and the values are more readable to match the ORD schema definitions.
* Added a "Database structure" section to the ORM readme.


ipendlet (approved):

ipendlet:
Tests validated and I was able to follow the changes.  I am not sure how these might effect the existing files though.  Is there a way to make sure that the modifications to the schema don't break passed versions of the protobuff? Or is this something that protobuff handles naturally?

skearnes:
> Tests validated and I was able to follow the changes. I am not sure how these might effect the existing files though. Is there a way to make sure that the modifications to the schema don't break passed versions of the protobuff? Or is this something that protobuff handles naturally?

I think you are referring to the changes in #666?

PR #671, opened 2023-04-11
fix: update examples to use latest version

flyi:
The examples in the README.md file were outdated and did not work with the latest version of the project. This commit updates the examples to use the latest version so that they work as expected.


skearnes:
Thanks @flyi!

skearnes (approved):

Issue #683, opened 2023-06-06
Stoichiometry


qai222:
**Is your feature request related to a problem? Please describe.**
Record the stoichiometric coefficients of the main chemical equation (assumed following #682) to have a balanced equation. 

**Describe the solution you'd like**
A field at the `Compound` level. The coefficient should be additive for the same `Compound`.

1. (without modifying the current schema) convention: for the main chemical equation, the stoichiometric coefficient of a `Compound` is recorded as a custom `CompoundIdentifier` whose value is the stringified stoichiometric coefficient.
2. extend `Amount.kind` to include stoichiometric coefficients.

**Describe alternatives you've considered**
Something at the `Reaction` level, ex. a balanced reaction SMILES as a `ReactionIdentifier`. The problem here is although I can have a balanced equation, 
1. I have to have all `Compound`s identified by SMILES.
2. If the same `Compound` appears in multiple `ReactionInput`s, I have no way to know the exact coefficient of one `Compound` in a specific `ReactionInput`.

qai222:
Convention for compound stoichiometry:
- The preferred field for compound stoichiometry is the map `Compound.features`.
- The key should be "stoichiometric_coefficient" or "stoichiometric_ratio".
- The value should be a `Data` mesaage with its `float_value` representing the compound's stoichiometric coefficient or ratio.
(from discussion during OH)

Issue #682, opened 2023-06-06
side product and byproduct


qai222:
**Is your feature request related to a problem? Please describe.**
There should be a way to tell if a non-desired product is a side product or a byproduct.

We already have a univalued `reaction_role` field for each compound, which implies a main chemical equation (based on which `reaction_role`s are assigned). 
- The products of this equation = one desired product + a set of byproducts.
- If a product is not on the right side of the main chemical equation, by definition it is a side product;
  - Side products are excluded in balancing the main chemical equation.

**Describe the solution you'd like**
extend the enum `ReactionRoleType` to include `SIDE_PRODUCT`

```
    // If a ProductCompound has reaction_role == PRODUCT and is_desired_product == True,
    // then it is the desired product of the main chemical equation.
    // If a ProductCompound has reaction_role == PRODUCT and is_desired_product == False,
    // then it is a byproduct product of the main chemical equation.
    // If a ProductCompound has reaction_role == SIDE_PRODUCT,
    // then it is a product not in the main chemical equation.
    SIDE_PRODUCT = 9;
```

PR #684, opened 2023-06-22
reaction role

qai222:
- Extended ReactionRoleType, added validation functions. Related issue: #682 
- Comments about stoichiometry convention. Related issue: #683


proto/reaction.proto
```diff
@@ -261,6 +261,7 @@ message Compound {
   // properties of the compound (e.g., molecular weight), heuristic estimates
   // of physical properties (e.g., ClogP), optimized geometries (e.g., through
   // DFT), and calculated stereoelectronic descriptors.
+  // Convention: also put stoichiometry ratio/coefficient here, see https://github.com/open-reaction-database/ord-schema/issues/683
```
skearnes, line 264: I'd suggest creating a new section in the readme called "conventions" and moving this comment there; that's a better place for notes like this.

proto/reaction.proto
```diff
@@ -287,6 +288,14 @@ message ReactionRole {
     // A product can be any species produced by the reaction, whether desired
     // or undesired.
     PRODUCT = 8;
+    // When there is only one intended chemical equation.
```
skearnes, line 291: I'm having trouble following this comment. Maybe you mean something like:

```
// Use BYPRODUCT to indicate a chemical species that is an expected result of the reaction but is not the product of interest.
// Use SIDE_PRODUCT to indicate the product of a side reaction.
// See https://doi.org/10.1021/op300317g for a discussion of these terms.

File: proto/reaction.proto [hunk 3]
```diff
@@ -887,6 +896,7 @@ message ProductCompound {
   }
   Texture texture = 5;
   // Any features (e.g., calculated descriptors) of the product compound.
+  // Convention: also put stoichiometry ratio/coefficient here, see https://github.com/open-reaction-database/ord-schema/issues/683
```
skearnes, line 899: same comment as above; the proto definition probably isn't the right place for this

proto/reaction.proto
qai222: I feel the clause "When there is only one intended chemical equation" is important here: if there are more than one intended chemical equation and product A is a byproduct, I still don't know which equation A is a part of. One can use desired/side/byproduct roles to balance an equation only "when there is only one intended chemical equation" 

Also, is it possible (and should we allow it) for a product to be a byproduct and a side product at the same time?

skearnes: It's fine with me if you add that clause somewhere; is it common for there to be more than one? I guess I was assuming this was implied. I think that the definitions given in that article are mutually exclusive (assuming a single reaction is intended), so I think a single assignment is appropriate. Also, note that these reaction role assignments are not designed to be gospel truth; they are always going to be somewhat subjective.

skearnes:
Please run the compile_proto_wrappers.sh and format.sh scripts so the tests pass.

qai222:
> Please run the compile_proto_wrappers.sh and format.sh scripts so the tests pass.

`compile_proto_wrappers.sh` made some changes (deletions) that I don't quite understand, still, two tests failed. Using `libprotoc 3.15.8`.

proto/reaction.proto
```diff
@@ -287,6 +287,13 @@ message ReactionRole {
     // A product can be any species produced by the reaction, whether desired
     // or undesired.
     PRODUCT = 8;
+    // When there is one intended chemical equation:
+    // - Set `is_desired_product=True` to indicate a desired product.
+    // - Use BYPRODUCT to indicate a chemical species that is an expected result of the reaction but is not the product of interest.
```
skearnes, line 292: one of the tests is failing because this line is too long; clang-format should fix it (should be called by format.sh, make sure you pay attention to the warnings about whether stuff needs to be installed)

skearnes:
> > Please run the compile_proto_wrappers.sh and format.sh scripts so the tests pass.
> 
> `compile_proto_wrappers.sh` made some changes (deletions) that I don't quite understand, still, two tests failed. Using `libprotoc 3.15.8`.

Make sure you're using the same versions as in this test: https://github.com/open-reaction-database/ord-schema/blob/main/.github/workflows/run_tests.yml#L78

js/ord-schema/proto/reaction_pb.js
```diff
@@ -4814,7 +4814,9 @@ proto.ord.ReactionRole.ReactionRoleType = {
   WORKUP: 5,
   INTERNAL_STANDARD: 6,
   AUTHENTIC_STANDARD: 7,
-  PRODUCT: 8
```
skearnes, line 4817: I think you need to pull in the latest changes from `main` to make sure the proto updates won't conflict with some recent changes upstream.

skearnes: ah ok, false alarm; looks like you already merged everything relevant here

ord_schema/validations.py
```diff
@@ -783,9 +783,18 @@ def validate_reaction_workup(message: reaction_pb2.ReactionWorkup):
...
-    # Can only have one desired product
-    if sum(product.is_desired_product for product in message.products) > 1:
-        warnings.warn("Cannot have more than one desired product!", ValidationError)
+    # *Usually* there should be at most one PRODUCT & is_desired_product
+    ndp = sum(
+        product.is_desired_product
+        for product in message.products
+        if product.reaction_role == reaction_pb2.ReactionRole.ReactionRoleType.PRODUCT
+    )
+    if ndp > 1:
```
skearnes, line 792: i'm going to approve this, but in the future please use better variable names; `ndp` is not meaningful

skearnes (approved):

Issue #693, opened 2023-07-19
Improve unit conversions and representations


connorcoley:
We have a unit parser to help convert strings to messages. For each type of unit, there are multiple possible units (e.g., days, hours, minutes, seconds for time). 

- We might want to parse a string in one format into a message, then convert that message units into a default for that unit type, or a specific target. For example, converting "11 days" to "264 hours". This will require defining conversion factors, as we already do for volumes and flow rates to handle flow rates.
- We also might want to make it easier to interpret a message that defines a value and unit. The default `__repr__` for a message will print the value and unit on two lines. We can easily access `message.value` to get a float and `message.unit` to get an int corresponding to the enum, but this `message.unit` is not very helpful. Conversion from that int to a string requires slightly more familiarity with protocol buffers (and more code) than I'd like. Should we consider overriding the `__repr__` for messages that have (value, unit, precision) fields to be `"{value} \pm {precision} {unit_as_string}"`?

Edit: The second use case is addressed by https://github.com/open-reaction-database/ord-schema/blob/main/ord_schema/units.py#L253 already

PR #694, opened 2023-07-26
Add unit conversion method to UnitResolver

connorcoley:
Addresses #693

### DIFF
--- a/ord_schema/units.py
+++ b/ord_schema/units.py
@@ -14,7 +14,7 @@
 """Helpers for translating strings with units."""
 
 import re
-from typing import Optional, Type
+from typing import Optional, Type, Union, Tuple
 
 import numpy as np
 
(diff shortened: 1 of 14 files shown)


ord_schema/units.py
```diff
@@ -138,6 +138,63 @@
     "m": "ambiguous between meter and minute",
 }
 
+_UNIT_CONVERSIONS = {
```
connorcoley, line 141: Since I didn't add an exhaustive set of test cases, would appreciate another pair of eyes on this (@skearnes, @qai222)

FanwangM (approved):
The unit conversions look good to me.

ord_schema/units.py
```diff
@@ -237,17 +287,109 @@ def resolve(self, string: str, allow_range: bool = False) -> ord_schema.UnitMess
...
+            message: a message with units, e.g., Mass, Length.
+            new_units: the desired units of the new message, expressed
+                either as a string or an integer (ENUM value). Use of a
+                string is recommended due to the ambiguity of using
+                ENUM values; for example, Mass.GRAM == Time.MINUTE.
+
+        Returns:
+            A new message with units, e.g., Mass, Length.
+        """
+        #pylint: disable=too-many-branches
```
qai222, line 331: looks like the missing space after `#` is causing errors in format tests (GitHub actions)

ord_schema/units.py
```diff
@@ -138,6 +138,63 @@
...
+    },
+    reaction_pb2.Current: {
+        reaction_pb2.Current.AMPERE: 1,
+        reaction_pb2.Current.MILLIAMPERE: 0.001,
+    },
+    reaction_pb2.Voltage: {
+        reaction_pb2.Voltage.VOLT: 1,
+        reaction_pb2.Voltage.MILLIVOLT: 0.001,
+    },
+    reaction_pb2.FlowRate: {
```
qai222, line 189: Am I crazy or `1 uL/min` should be `1/60 uL/s`?

connorcoley: My strategy here was to peg everything to a reference of 1 'unit'. So for example, one `VOLT` is 1 unit (V), one `MILLIVOLT` is 0.001 unit (V).

One `MICROLITER_PER_SECOND` is 60x as high of a flowrate as one `MICROLITER_PER_MINUTE`

ord_schema/units.py
connorcoley: Thanks for the catch

qai222 (approved):

ord_schema/units.py
```diff
@@ -294,7 +294,7 @@ def resolve(self, string: str, allow_range: bool = False) -> ord_schema.UnitMess
             return message(value=value, precision=precision, units=unit)
         return message(value=value, units=unit)
 
-    def resolve_unit(self, string_unit: str) -> Tuple[ord_schema.UnitMessage, int]:
+    def resolve_unit(self, string_unit: str) -> Tuple[ord_schema.UnitMessage, ord_schema.Message]:
```
qai222, line 297: I thought this was correct: the type check fail actually comes from `unit_synonyms` in `__init__` (the second key should be an enum value).

skearnes (approved):

ord_schema/units.py
connorcoley: Could you clarify? All tests are passing but I will wait for you to resolve this comment (or help me figure out what's wrong) before merging

qai222: From this [line](https://github.com/open-reaction-database/ord-schema/blob/d23c1c3a7acf40b11ca8c8bca61ab8f625c4c67e/ord_schema/units.py#L220C9-L220C113) the second key in `unit_synonyms` is hinted to be an `ord_schema.Message` (causing the "wrong" return type in `resolve_unit`), but it should be just an ENUM option. So maybe it's better to change type hints there. It doesn't matter much tho.

PR #698, opened 2023-08-21
use `Texture` for `ReactionInput` and `Compound`

qai222:


connorcoley (commented):
Steve and I were talking about this and had some thoughts:
- The `MixtureDescription` field seems a bit unnecessary. I get wanting to understand the "texture/phase" (for lack of a better phrase) from the perspective of processability, so we should include it somehow
- Perhaps `StateOfMatter` could be generalized a bit more to include`SLURRY` and other ways that you would describe a potential input. This would be the singular enum we would use to record this information
- Probably don't need plasma, lol
- Having this kind of texture/state/phase defined at the `Compound`, `CrudeComponent`, and `ReactionInput` level seems okay. We would need to expand the validations for this to (1) ensure details are given if type is CUSTOM, and (2) make sure that single-compound inputs have the same texture/state/phase as the compound when both are defined
- CUSTOM should always be the second enum option after UNSPECIFIED, and all enums should have a `CUSTOM` option (probably including this texture/state/phase
- Is there any world in which we simply expand the product texture field and use it for the `ProductCompound`, `Compound`, `CrudeComponent`, and `ReactionInput` fields?

qai222:
Thanks for the comments. 

> * The `MixtureDescription` field seems a bit unnecessary. I get wanting to understand the "texture/phase" (for lack of a better phrase) from the perspective of processability, so we should include it somehow
> * Perhaps `StateOfMatter` could be generalized a bit more to include`SLURRY` and other ways that you would describe a potential input. This would be the singular enum we would use to record this information
> * Is there any world in which we simply expand the product texture field and use it for the `ProductCompound`, `Compound`, `CrudeComponent`, and `ReactionInput` fields?

My idea was just to expand `texture` but it feels really weird to put "dispersion" or "gas" there. A more suitable term may be `morphology` or simply `form`. But I'm not sure if changing message names would break compatibility. 

Another thing is the current options of `texture` are not disjoint (ex. a `powder`, as a `solid`, can also be a `crystal`). I feel two `enum` each with three disjoint options and clear physical meaning are often better than one `enum` of >9 options.

> * CUSTOM should always be the second enum option after UNSPECIFIED, and all enums should have a `CUSTOM` option (probably including this texture/state/phase
> * Having this kind of texture/state/phase defined at the `Compound`, `CrudeComponent`, and `ReactionInput` level seems okay. We would need to expand the validations for this to (1) ensure details are given if type is CUSTOM, and (2) make sure that single-compound inputs have the same texture/state/phase as the compound when both are defined

Got it, will do this after we settle the design.

> * Probably don't need plasma, lol

Only include for completeness... although this is in a recent JACS [Plasma Electrochemistry for Carbon–Carbon Bond Formation via Pinacol Coupling](https://pubs.acs.org/doi/epdf/10.1021/jacs.3c01779)
image

(hope we don't need to describe excited states in the future)

qai222:
todo: throw a warning for a `ReactionInput` when 
1. there is only one `Compound`, and
2. `ReactionInput.texture` is different from `Compound.texture`

ord_schema/validations.py
```diff
@@ -496,6 +496,35 @@ def validate_reaction_input(message: reaction_pb2.ReactionInput):
...
+        reaction_pb2.ProductCompound.Texture.SOLID: 3,
+    }
+    input_state_code = texture_type_to_state_of_matter[message.texture.type]
+    if input_state_code is not None:
+        components = message.components + message.crude_components
+        component_state_codes = [texture_type_to_state_of_matter[c.texture.type] for c in components]
+        if (
+            component_state_codes
+            and None not in component_state_codes
+            and max(component_state_codes) > input_state_code
```
connorcoley, line 520: Can you document this a little more with in-line comments? My reading is that a warning will be issued if any component is solid but the input is liquid, which would be a common situation for stock solutions.

proto/reaction.proto
```diff
@@ -231,6 +233,7 @@ message CrudeComponent {
   optional bool has_derived_amount = 3;
   // If the entire crude mixture was not used, need to specify an amount.
   Amount amount = 4;
+  ProductCompound.Texture texture = 5;
```
connorcoley, line 236: We should pull `Texture`'s definition outside of `ProductCompound` if it is being used in multiple places

ord_schema/validations.py
```diff
@@ -496,6 +496,35 @@ def validate_reaction_input(message: reaction_pb2.ReactionInput):
...
+        ):
+            warnings.warn(
```
connorcoley, line 522: Would be good to add a few test cases to ensure that validation warnings are being issued when expected (and not otherwise)

ord_schema/validations.py
qai222: my bad, it should have been `max(component_state_codes) < input_state_code` so warnings are issued for G->L (gas component, liquid reaction_input), L->S, GL->S, LL->S. I will add comments on this.

proto/reaction.proto
qai222: Done in the latest commit. Not sure if this would break backward compatibility.

ord_schema/validations.py
qai222: test cases added

qai222:
Moving `Texture` out from `ProductCompound` seems backward compatible.
```
$ python validate_dataset.py --input=ord_dataset-e984b2d4813f44d59867e1771acc9b66.pb.gz
INFO 2023-09-14 07:19:11,320 validate_dataset.py:49: Found 1 datasets
INFO 2023-09-14 07:19:11,320 validate_dataset.py:54: Validating ord_dataset-e984b2d4813f44d59867e1771acc9b66.pb.gz
INFO 2023-09-14 07:19:14,273 validations.py:122: Validation summary for ord_dataset-e984b2d4813f44d59867e1771acc9b66.pb.gz: (720,)/720 successful (0 failures)
```

qai222:
4 tests failed, all seem unrelated to `ord_schema` but come from `npm install jest@^29.3.1` which requires a non-existing version of `babel/types@^7.22.18` (latest `7.22.17`).

qai222:
> 4 tests failed, all seem unrelated to `ord_schema` but come from `npm install jest@^29.3.1` which requires a non-existing version of `babel/types@^7.22.18` (latest `7.22.17`).

This issue has been resolved in `jest@29.7.0`,all checks passed.

proto/reaction.proto
```diff
@@ -209,6 +211,32 @@ message UnmeasuredAmount {
   string details = 2;
 }
 
+/**
+ * This qualitatively describes the apparent size and morphology of a Compound
+ * or a ReactionInput.
+ */
+message Texture {
+  // Note not all of the type options are mutually disjoint.
```
skearnes, line 219: this comment is true of all enums; please remove

proto/reaction.proto
skearnes: Please verify your checks on ord-data datasets before closing this thread.

qai222: > Please verify your checks on ord-data datasets before closing this thread.

verified using `ord_schema/scripts/validate_dataset.py` for small (size < 5 MB) datasets in https://github.com/open-reaction-database/ord-data/tree/main/data

proto/reaction.proto
qai222: > this comment is true of all enums; please remove

removed in https://github.com/open-reaction-database/ord-schema/pull/698/commits/e0891b5b1a796cf295030bef6f72654f00ee5d25

proto/reaction.proto
```diff
@@ -209,6 +211,31 @@ message UnmeasuredAmount {
   string details = 2;
 }
 
+/**
+ * This qualitatively describes the apparent size and morphology of a Compound
+ * or a ReactionInput.
```
skearnes, line 216: include product in this list as well

skearnes (approved):

skearnes:
please update the PR description; only deals with Texture now

ord_schema/__init__.py
```diff
@@ -37,7 +37,7 @@
     reaction_pb2.PressureConditions.Atmosphere,
     reaction_pb2.PressureConditions.PressureMeasurement,
     reaction_pb2.PressureConditions.PressureControl,
-    reaction_pb2.ProductCompound.Texture,
+    reaction_pb2.Texture,
```
skearnes, line 40: keep this list alphabetical

ord_schema/validations.py
```diff
@@ -496,6 +496,35 @@ def validate_reaction_input(message: reaction_pb2.ReactionInput):
                     ValidationWarning,
                 )
 
+    texture_type_to_state_of_matter = {
```
skearnes, line 499: you should probably use an enum here; integer values are not meaningful

skearnes, line 499: document these values; consider creating a new enum with integer values

qai222: added an `IntEnum` in https://github.com/open-reaction-database/ord-schema/pull/698/commits/c03e02687ae6831084669afc9a01eee6e3984d55

ord_schema/__init__.py
qai222: fixed in https://github.com/open-reaction-database/ord-schema/pull/698/commits/c03e02687ae6831084669afc9a01eee6e3984d55

proto/reaction.proto
qai222: fixed in https://github.com/open-reaction-database/ord-schema/pull/698/commits/c03e02687ae6831084669afc9a01eee6e3984d55

PR #699, opened 2023-08-25
Fix problem of generating dataset when having `reaction_id` in the template (#697)

### LINKED ISSUE #697: Enumeration script creates invalid datasets when reaction_id included
If a `reaction_id` is defined in the template Reaction, an enumerated Dataset will have repeated `reaction_id` values and be rejected during submission. I believe the enumeration script should simply check for the existence of the `reaction_id` field in the template and remove it.

(https://github.com/open-reaction-database/ord-data/pull/174)

FanwangM:
The PR is an attempt to fix #697 by filtering out the line with reaction_id.

### DIFF
--- a/ord_schema/scripts/enumerate_dataset_test.py
+++ b/ord_schema/scripts/enumerate_dataset_test.py
@@ -78,6 +78,7 @@ def setup(tmp_path) -> tuple[str, str, str]:
             }
         }
     }
+    reaction_id: "ord-d73a86df4d0d4a32a23007aeff1a94e4"
     """
     template_filename = os.path.join(dirname, "template.pbtxt")
     with open(template_filename, "w") as f:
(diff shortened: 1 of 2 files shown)


connorcoley (changes requested):
Instead of doing string operations on the template pbtxt, it would be cleaner to remove the field from the `dataset` after it is generated by the enumeration, I think. WDYT?

FanwangM:
> Instead of doing string operations on the template pbtxt, it would be cleaner to remove the field from the `dataset` after it is generated by the enumeration, I think. WDYT?

This can be a cleaner solution. New changes are committed here.

One question, do we need to assign a reaction_id for each reaction in the database after the reaction list is constructed?

connorcoley:
> One question, do we need to assign a reaction_id for each reaction in the database after the reaction list is constructed?

No, the submission workflows will automatically assign a `reaction_id` to each

connorcoley (approved):

PR #700, opened 2023-09-04
Update protocol complier usage because of protobuf updates

FanwangM:
Optional fields for proto3 are enabled by default, and no longer require the --experimental_allow_proto3_optional flag,
https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0. We are using 22.3 and it should be safe to remove this option.


skearnes (approved):

FanwangM:
Given that this option does not break anything, I am going to close it.

Issue #711, opened 2024-02-16
Code completion of protobuf classes in VSCode


ottokruse:
**Describe the bug**
In VSCode I want code completion, but VSCode does not recognize the generated Python classes for protobuf. This hampers development speed.

Excuse me for filing here first––I'm not sure whose bug this is, might also be VSCode of Protobuf. But I assume "normally" code completion with protobufs work in VSCode, as VSCode and protobufs are such common tools.

And maybe this is not a bug, and I just don't know how to set it up right––any hints would we appreciated!

**Screenshots**
<img width="760" alt="image" src="https://github.com/open-reaction-database/ord-schema/assets/6275520/3264ffaa-9cb0-4564-a2df-15b7220420f5">

**Desktop (please complete the following information):**
 - MacOS
 - ORD Schema v0.3.71

qai222:
It's probably not VSCode's problem. I had the same issue with Pycharm, and I found two solutions.
1. In `compile_proto_wrappers.sh` you can find the command used to generate `*.pb2` files, simply add `--pyi_out=.`, so you get
    ```
    protoc \
      --proto_path=.. \
      --python_out=. \
      --js_out=import_style=commonjs,binary:js \
      ../ord-schema/proto/*.proto --pyi_out=.
    ```
    now run the script and you will get two `*.pyi` files in `ord_schema/proto/`. These are the "stubs" files like this
    ```
    class Reaction(_message.Message):
        __slots__ = ["identifiers", "inputs", "setup", "conditions", "notes", "observations", "workups", "outcomes", "provenance", "reaction_id"]
        class InputsEntry(_message.Message):
            __slots__ = ["key", "value"]
            KEY_FIELD_NUMBER: _ClassVar[int]
            VALUE_FIELD_NUMBER: _ClassVar[int]
            key: str
            value: ReactionInput
    ````
    and your IDE should recognize them.
2. Generate actual python classes using `betterproto`. Here is a guide I wrote a while ago https://github.com/qai222/python-betterproto-ord/tree/master/ord_betterproto.

ottokruse:
Thanks will give that shot!

A development opportunity for the library here, to make such jumping through hoops for proper intellisense unnecessary 🙌

qai222:
> Thanks will give that shot!
> 
> A development opportunity for the library here, to make such jumping through hoops for proper intellisense unnecessary 🙌

No problem. Let me know if solution 1 works for you, and I can make a PR to include `pyi` files.

[...]

PR #713, opened 2024-02-19
add compiler option to export pyi files

qai222:
add pyi stub files to support IDE auto-completion 

see solution 1 in #711

### DIFF
--- a/compile_proto_wrappers.sh
+++ b/compile_proto_wrappers.sh
@@ -20,4 +20,4 @@ protoc \
   --proto_path=.. \
   --python_out=. \
   --js_out=import_style=commonjs,binary:js \
-  ../ord-schema/proto/*.proto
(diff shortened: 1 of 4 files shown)


compile_proto_wrappers.sh
```diff
@@ -20,4 +20,4 @@ protoc \
   --proto_path=.. \
   --python_out=. \
   --js_out=import_style=commonjs,binary:js \
-  ../ord-schema/proto/*.proto
+  ../ord-schema/proto/*.proto --pyi_out=.
```
skearnes, line 23: please put this flag after --python_out

qai222: fixed in https://github.com/open-reaction-database/ord-schema/pull/713/commits/0313b8dcdd5bc44586f266637464444eeeb8a300

PR #722, opened 2024-06-14
Update ReactionRole use in get_reaction_smiles

skearnes:
* Updates `get_reaction_smiles` to accept `ReactionRole.UNSPECIFIED` as a marker for reactants and/or products. The test dataset does not have product reaction roles set, and this was causing reaction SMARTS queries to fail since reaction SMILES were not being generated as part of the dataset creation (and thus the rdkit.reactions table was empty).
* Adds examples of additional chemistry-aware search functions.
* Makes `RDKitMol` table fingerprint columns explicit in the ORM mapper for improved readability and IDE inference.
* Silence RDKit logging when the ORM is imported.
* Set the default log level to DEBUG.

### DIFF
--- a/ord_schema/message_helpers.py
+++ b/ord_schema/message_helpers.py
@@ -352,6 +352,7 @@ def get_reaction_smiles(
     message: reaction_pb2.Reaction,
     generate_if_missing: bool = False,
     allow_incomplete: bool = True,
+    allow_unspecified_roles: bool = True,
     validate: bool = False,
     canonical: bool = True,
 ) -> Optional[str]:
@@ -364,6 +365,8 @@ def get_reaction_smiles(
         allow_incomplete: Boolean whether to allow "incomplete" reaction SMILES
             that do not include all components (e.g. if a component does not
             have a structural identifier).
+        allow_unspecified_roles: If True, reactants and products with the UNSPECIFIED reaction role will be included
+            when generating a reaction SMILES.
         validate: Boolean whether to validate the reaction SMILES with rdkit.
             Only used if allow_incomplete is False.
         canonical: Boolean whether to return a canonicalized reaction SMILES.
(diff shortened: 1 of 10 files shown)


qai222:
>Updates get_reaction_smiles to accept ReactionRole.UNSPECIFIED as a marker for reactants and/or products. The test dataset does not have product reaction roles set, and this was causing reaction SMARTS queries to fail since reaction SMILES were not being generated as part of the dataset creation (and thus the rdkit.reactions table was empty).

Does this only affect the test dataset?

skearnes:
> > Updates get_reaction_smiles to accept ReactionRole.UNSPECIFIED as a marker for reactants and/or products. The test dataset does not have product reaction roles set, and this was causing reaction SMARTS queries to fail since reaction SMILES were not being generated as part of the dataset creation (and thus the rdkit.reactions table was empty).
> 
> Does this only affect the test dataset?

No, there are about 1k inputs and 250k outcomes in the database that have an UNSPECIFIED reaction role. In cases where there was no explicit REACTANT or PRODUCT, these reactions will now be able to generate reaction SMILES.

skearnes:
@qai222 could you please approve?

ord_schema/message_helpers.py
```diff
@@ -386,6 +389,11 @@ def get_reaction_smiles(
 
     reactants, agents, products = set(), set(), set()
     roles = reaction_pb2.ReactionRole
+    reactant_roles = [roles.REACTANT]
+    product_roles = [roles.PRODUCT]
```
qai222, line 393: Is it intentional to exclude BYPRODUCT or SIDE_PRODUCT?

qai222 (approved):

ord_schema/message_helpers.py
skearnes: Yes; we are only including the desired product when generating reaction SMILES. Happy to discuss if you think we should change that, but maybe in a new issue?

PR #723, opened 2024-06-16
Move database query classes/functions to ord-schema

skearnes:
Migration from https://github.com/open-reaction-database/ord-interface/blob/bd9ff55a7d9aa6f3161ab85e5fcc92c75973c311/ord_interface/client/query.py

Updates:
* Simpler reaction query class
* A single query is executed that `INTERSECT`s all of the component queries
* Query testing does not require docker

### DIFF
--- a/ord_schema/orm/conftest.py
+++ b/ord_schema/orm/conftest.py
@@ -16,7 +16,9 @@
 import os
 from typing import Iterator
 
+import psycopg2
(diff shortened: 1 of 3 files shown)


skearnes:
Actually I think it's better to keep this in ord-interface since it's going to be less stable than the overall ORM.

PR #724, opened 2024-06-16
Add num_reactions to dataset table

skearnes:


ord_schema/orm/mappers.py
```diff
@@ -264,6 +266,8 @@ def from_proto(  # pylint: disable=too-many-branches
             kwargs[field.name] = value
     if isinstance(message, dataset_pb2.Dataset):
         kwargs["md5"] = md5(message.SerializeToString(deterministic=True)).hexdigest()
+        assert hasattr(message, "reactions") and hasattr(message, "reaction_ids")  # Type hints.
+        kwargs["num_reactions"] = len(message.reactions) or len(message.reaction_ids)
```
qai222, line 270: I'm not sure if they can be different. Also, can `[r.reaction_id for r in dataset.reactions]` be different from `dataset.reaction_ids`?

qai222 (approved):

ord_schema/orm/mappers.py
qai222: (somewhat a random question, feel free to resolve it)

skearnes: A dataset is composed of _either_ Reaction objects or a list of reaction IDs; this line just covers both cases. See the Dataset definition here: https://github.com/open-reaction-database/ord-schema/blob/main/proto/dataset.proto#L39

PR #725, opened 2024-06-20
Simplify ORM dataset ingestion

skearnes:
Fixes #721

Issue #672, opened 2023-04-15
`add_rdkit` error out for a particular dataset


qai222:
**Describe the bug**
`ord_schema.orm.database.add_rdkit` error out for a particular dataset.

**To Reproduce**
```python
from ord_schema.message_helpers import fetch_dataset
from ord_schema.orm.database import add_dataset, add_rdkit
from sqlalchemy import create_engine
from sqlalchemy.orm import Session

connection_string = f"postgresql://127.0.0.1:5432/ord"

dataset_id = "ord_dataset-b195433d5c354ddfb6cde0d53c41910f"
dataset = fetch_dataset(dataset_id)
engine = create_engine(connection_string, future=True)

with Session(engine) as session:
    add_dataset(dataset, session)
    session.flush()
    session.commit()
# INFO 2023-04-14 19:55:36,833 database.py:70: Adding dataset ord_dataset-b195433d5c354ddfb6cde0d53c41910f
# INFO 2023-04-14 19:56:51,821 database.py:73: from_proto() took 74.98789548873901s
# INFO 2023-04-14 19:57:05,244 database.py:76: session.add() took 13.422787189483643s

with Session(engine) as session:
    add_rdkit(session)
    session.commit()
# INFO 2023-04-14 20:00:17,986 database.py:89: Populating RDKit reaction columns
# INFO 2023-04-14 20:00:38,573 database.py:98: Adding reaction took 20.587525844573975s
# INFO 2023-04-14 20:00:38,574 database.py:99: Populating RDKit mol columns
# Traceback (most recent call last):
#   File "/home/qai/local/miniconda3/envs/LLM_organic_synthesis__ord_data/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1964, in _exec_single_context
#     self.dialect.do_execute(
#   File "/home/qai/local/miniconda3/envs/LLM_organic_synthesis__ord_data/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 748, in do_execute
#     cursor.execute(statement, parameters)
# psycopg2.errors.InternalError_: molFromPickle: invalid value in pickle
# 
# 
# The above exception was the direct cause of the following exception:
# 
# Traceback (most recent call last):
#   File "/home/qai/workplace/LLM_organic_synthesis/setup_db.py", line 66, in <module>
#     ord_add_rdkit()
#   File "/home/qai/workplace/LLM_organic_synthesis/setup_db.py", line 44, in ord_add_rdkit
[...]

PR #726, opened 2024-06-20
Explicit tests for #672

skearnes:
@qai222 the edits I made to the test data are not an adequate test for regression on #672, but we have added some explicit skips for those cases in the code. Thanks for flagging this on #725.

This PR adds another test that verifies that the offending SMILES is in the `ord.compound` table but not the `rdkit.mols` table.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -34,7 +34,7 @@ jobs:
         run: |
           # NOTE(skearnes): conda is only used for postgres (not python).
           # NOTE(skearnes): rdkit-postgresql may not be available for ARM.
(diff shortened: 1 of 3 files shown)


skearnes:
@qai222 I guess we have a decision to make. I think that the errors are correctly being caught and ignored on `main`; I might prefer that over going to an older version. Let me do some more testing...

skearnes:
@qai222 haha I forgot that I already added an explicit skip for this case: https://github.com/open-reaction-database/ord-schema/blob/main/ord_schema/orm/database.py#L162. I'm still seeing that disappear from the `Compound` table though, which shouldn't be happening. I'll try to fix that in this PR.

skearnes:
Closing this; this was mostly a demonstration that the current ORM is behaving as expected.

PR #734, opened 2024-07-10
added pump enums

bdeadman:


proto/reaction.proto
```diff
@@ -154,6 +154,10 @@ message ReactionInput {
       ADDITION_FUNNEL = 5;
       PIPETTE = 6;
       POSITIVE_DISPLACEMENT_PIPETTE = 7;
+	  PISTON_PUMP = 8;
+	  SYRINGE_PUMP = 9;
+	  PERISTALTIC_PUMP = 10;
+	  OTHER_PUMP;
```
skearnes, line 160: all entries need a tag number, and should be indented the same as the other entries

skearnes: Also, we use CUSTOM in all enums as a catch-all; there should be no need for OTHER_PUMP

bdeadman: Sorry, didn't realise this PR had been created. I've been wrestling with compiling the protocol buffers before I intended to  do the PR.

I caught the numbering and indent issues on my fork so that's already fixed - just need to close this or work out how to update the PR. If @skearnes is at OH I'll ask then.

bdeadman:
Closed this PR as it has been resubmitted (with fixes) to a branch in #736

Issue #735, opened 2024-07-10
Error while compiling protocol buffers following edits to the schema


bdeadman:
Encountered the following error while trying to compile the protocol buffers following edits to schema enums. Windows 11 system running python in a Conda environment with libprotoc 3.12.4.

(ord-dev) bdeadman@BJDFrameworkWin:~/ord-schema$ ./compile_proto_wrappers.sh
+ protoc --proto_path=.. --python_out=. --pyi_out=. --js_out=import_style=commonjs,binary:js ../ord-schema/proto/dataset.proto ../ord-schema/proto/reaction.proto ../ord-schema/proto/test.proto
ord-schema/proto/reaction.proto: This file contains proto3 optional fields, but --experimental_allow_proto3_optional was not set.

skearnes:
@bdeadman I think you might be running an old version of protoc; that flag was deprecated a while ago. You should be on libprotoc 22.3: https://github.com/open-reaction-database/ord-schema/blob/main/.github/workflows/run_tests.yml#L87

skearnes:
https://github.com/protocolbuffers/protobuf/releases/tag/v22.3

bdeadman:
Thanks @skearnes. I can confirm that libprotoc 22.3 works after some tweaks. The pre-built binaries seem to have lost protoc-gen-js in v21. If you compiled from source then it may be different, but not something I want to tackle at the moment.

The [npm package](https://www.npmjs.com/package/protoc-gen-js) allowed me to get protoc-gen-js going in WSL. I'll close this issue now, but keeping this note here for future reference.

skearnes:
Note that you have to install the JS side separately: https://github.com/open-reaction-database/ord-schema/blob/main/.github/workflows/run_tests.yml#L89

PR #736, opened 2024-07-11
Added piston, syringe and peristaltic pump options for addition devices. 

bdeadman:
Added piston, syringe and peristaltic pump options for addition devices. This will set us up to better capture flow chemistry reactions.


bdeadman:
Note that the protocol buffers have not been compiled so that will need to be done first before merging into main.

bdeadman:
This is PR #734 after fixes and redirecting to a branch.

bdeadman:
Waiting for issue #735 before compiling.

skearnes (approved):

PR #757, opened 2024-12-12
Added typescript definitions to js module

daniil-sloboda:
* Added typescript definitions generation to output of the compilation script.     
* Updated package.json version

### DIFF
--- a/compile_proto_wrappers.sh
+++ b/compile_proto_wrappers.sh
@@ -21,4 +21,6 @@ protoc \
   --python_out=. \
   --pyi_out=. \
   --js_out=import_style=commonjs,binary:js \
-  ../ord-schema/proto/*.proto
+  ../ord-schema/proto/*.proto \
+  --plugin="protoc-gen-ts=./js/ord-schema/node_modules/.bin/protoc-gen-ts" \
+  --ts_out=./js/ \
--- a/compile_proto_wrappers.sh
+++ b/compile_proto_wrappers.sh
@@ -22,5 +22,4 @@ protoc \
   --pyi_out=. \
   --js_out=import_style=commonjs,binary:js \
   ../ord-schema/proto/*.proto \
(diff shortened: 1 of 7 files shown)


daniil-sloboda:
@skearnes Hello, can you please take a look at the PR when you have some spare time?

js/ord-schema/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ord-schema",
-  "version": "0.3.94",
+  "version": "0.3.95",
```
skearnes, line 3: you should not update the version manually; bump2version handles this: https://github.com/open-reaction-database/ord-schema/blob/main/.bumpversion.cfg

compile_proto_wrappers.sh
```diff
@@ -21,4 +21,6 @@ protoc \
   --python_out=. \
   --pyi_out=. \
   --js_out=import_style=commonjs,binary:js \
-  ../ord-schema/proto/*.proto
+  ../ord-schema/proto/*.proto \
+  --plugin="protoc-gen-ts=./js/ord-schema/node_modules/.bin/protoc-gen-ts" \
```
skearnes, line 25: If you can, please update this to remove `--plugin` and assume that protoc-gen-ts is in the path, like we do with JS: https://github.com/open-reaction-database/ord-schema/blob/main/.github/workflows/run_tests.yml#L88 (and update the GitHub workflow to add it)

daniil-sloboda: Thanks a lot for the explanation, I will update the PR tomorrow then

js/ord-schema/package.json
daniil-sloboda: Okay, will revert the changes

skearnes:
Thanks @daniil-sloboda this looks great!

skearnes (approved):

PR #759, opened 2025-01-14
Remove CrudeComponent.texture

skearnes:
This seems like it should be specified as ProductCompound.texture on the source reaction?


skearnes:
Note that I only see `texture` being used on ReactionInput and ProductCompound in the database, so it should be safe to remove this.

Issue #766, opened 2025-03-14
Subset type for Reaction Identifier


bdeadman:
**Is your feature request related to a problem? Please describe.**
There are times where a dataset includes subsets of reactions which the author may want to designate, for example two tables grouped into 1 dataset, different stages of a DOE campaign.

**Describe the solution you'd like**
Add a Reaction Identifier type for SUBSET.

**Describe alternatives you've considered**
The alternative is to have a separate dataset for each subset, but this loses the context that ties them together. This is the approach that has been used in some existing datasets (e.g. ord_dataset-2038a3c967db4a32a8fbce288437e929, ord_dataset-2be30f5d8dcd471aa6ad410bdee05902, ord_dataset-31989f1b2b9d4885b1dd2d9982da4517, ord-0defb78dc9b047179947fa8bf0330132 are 4 datasets from http://doi.org/10.1039/C6RE00153J).

**Additional context**
I'll setup the PR for ord-schema, but marking it here for discussion.

bdeadman:
An INDEX type could also be useful, although possibly redundant with the natural order of the reactions in the dataset. Having an INDEX reaction identifier would allow authors to explicitly define the order of reactions in a dataset when this has a specific meaning (defined in the associated details). In contrast, the order of reactions in a dataset does not mean that the data was definitely collected in that order.

PR #767, opened 2025-03-19
Added subset and index types to reaction identifier

bdeadman:
As proposed in #766.

Added SUBSET and INDEX types to reaction identifiers. This allows grouped and ordered relationships between reactions to be described.

Example usage:
A dataset containing several stages [SUBSET] of a DOE campaign, where the run number [INDEX] is explicitly documented.

An HTE dataset containing multiple plates [SUBSET] with well ids [INDEX] designating the spatial relationship of reactions in the plate.


connorcoley:
In the vessel message, we already have position/row/column to fill the role
of the “index” you’re proposing. I think those are redundant but please
correct me if not!

bdeadman:
I agree that reactor position/row/column is a better place to put the positional identifiers. I'll have to start using and recommending that label! I do however still think there is a use case for an index or order identifier for reactions performed in an explicit order, but where the exact times are not known. I'm currently making CUSTOM identifiers for 'run number' in a DOE and transient flow datasets.

PR #788, opened 2026-04-12
Drop Werkzeug dependency

skearnes:
## Summary

The only use of Werkzeug in `ord-schema` was `werkzeug.security.safe_join` in [`id_filename()`](ord_schema/message_helpers.py), which builds a `data/{shard}/{basename}` path for dataset storage. Pulling in a web framework for one `join` call is excessive — replace it with `posixpath.join` plus an `isalnum()` check on the shard.

- Removes `Werkzeug>=3.0.3` from `[project.dependencies]`
- `id_filename()` now uses `posixpath.join("data", shard, basename)` after verifying the shard is alphanumeric (stricter than `safe_join` and matches our hex-like dataset ID shape)

Werkzeug remains available transitively via `tensorboard` for the `examples` extra, so anyone who needs it for experimentation still gets it — it just isn't a required runtime dep of `ord-schema` anymore.

## Test plan

- [x] `pytest ord_schema/message_helpers_test.py ord_schema/scripts/process_dataset_test.py` — 94 passed locally
- [ ] CI green

### DIFF
--- a/ord_schema/message_helpers.py
+++ b/ord_schema/message_helpers.py
@@ -17,6 +17,7 @@ import enum
 import functools
 import gzip
 import os
+import posixpath
 import re
 import urllib.parse
 import warnings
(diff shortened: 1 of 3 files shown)


skearnes:
Folded into #779 since it's a tiny change in the same spirit as the rest of the dependency cleanup.

PR #790, opened 2026-04-22
Add Parquet-based Dataset serialization [bump minor]

skearnes:
## Summary

Introduces a Parquet-based serialization for `Dataset` messages alongside the existing `.pb.gz` format, to address two limitations of the current on-disk representation:

- No random access — the whole `Dataset` must be deserialized to read any part of it.
- Whole-dataset memory footprint — which has forced manual sharding of large datasets in [ord-data](https://github.com/open-reaction-database/ord-data) purely for memory reasons.

### Format

Each Parquet file **is** a serialized `Dataset`:

- Rows are `Reaction` messages with columns `reaction_id` (string) and `reaction` (serialized wire-format bytes).
- `Dataset.name` and `Dataset.description` are required and live in the Parquet footer key-value metadata under `ord.name` / `ord.description`. `Dataset.dataset_id` is optional (assigned during submission) and stored under `ord.dataset_id` when present. `ord.schema_version` is always written for forward compatibility.
- `Dataset.reaction_ids` is not persisted — the `reaction_id` column is the source of truth for a Parquet-serialized dataset.

Row-group-level random access and streaming iteration keep memory bounded regardless of dataset size. Row group size defaults to 1000.

### What's in this PR

- `ord_schema/parquet_dataset.py` — new module with:
  - `DatasetWriter` — streaming context manager. Requires non-empty `name` and `description`; `dataset_id` is optional.
  - `write_dataset(dataset, path)` — convenience that wraps `DatasetWriter` for a full `Dataset` message.
  - `read_dataset(path)` — reassembles a full `Dataset`. Prefer the streaming APIs below for large files.
  - `read_metadata(path)` — reads only the footer scalars; no column data. Cheap.
  - `iter_reactions(path, reaction_ids=None)` — row-group streaming with an optional id filter. Raises on an explicitly empty filter since that's almost always a bug.
  - `read_reaction(path, reaction_id)` — single-reaction lookup; row-group pruning when `reaction_id` values are clustered in groups, falls back to a full column scan otherwise.
- `ord_schema/scripts/pb_to_parquet_dataset.py` — migration script. Accepts one or more `.pb` / `.pb.gz` / `.pbtxt` / `.json` Dataset inputs and produces a single `.parquet`. Supports **un-sharding** (merging previously split datasets) via multi-input concatenation. Metadata comes from the first input by default; overridable via `--name` / `--description` / `--dataset-id`, and the script warns on non-overridden mismatches. Inputs are streamed one file at a time so peak memory is bounded by the largest single input.
- Tests for both modules (28 tests).
- `pyarrow>=14` added to `pyproject.toml` / `uv.lock`.

### Explicitly not in this PR

These are planned follow-ups, each sized for an independent review:

1. **Integrate `.parquet` into `message_helpers.load_message` / `write_message`.** Today the new API is deliberately separate to keep the seam obvious; once it stabilizes, routing `.parquet` through the generic helpers lets downstream tools pick it up transparently.
2. **Convert `validate_dataset.py` and `process_dataset.py` to streaming.** Both currently load full datasets; with the new API they can stream reactions via `iter_reactions`, which is the real memory win for validation of large datasets.
3. **Stream ORM ingestion in `ord_schema/orm/scripts/add_datasets.py`.** Same motivation — today it calls `load_message(..., Dataset)`, which is the biggest memory offender in the ingest path.
4. **Bulk-convert `../ord_data`.** Walk the existing `data/{shard}/ord-*.pb.gz` tree, run `pb_to_parquet_dataset.py` per dataset (opportunistically un-sharding any datasets that were previously split for memory reasons), and produce a parallel `.parquet` tree. This is a one-shot data-repo migration, not schema code.
5. **Decide the cutover story.** Once (1)–(4) land, whether `.parquet` becomes the primary format (with `.pb.gz` as a compatibility export) or both coexist indefinitely.
6. **Atomic writes for all `ord_schema` writers** ([#791](https://github.com/open-reaction-database/ord-schema/issues/791)). Neither `write_message` nor `DatasetWriter` is crash-safe today; a partial `.parquet` in particular is hard to distinguish from a valid one. Filed separately so it can be reviewed on its own.

## Test plan

- [x] `pytest ord_schema/parquet_dataset_test.py ord_schema/scripts/pb_to_parquet_dataset_test.py` — 28/28 pass locally
- [x] `pytest ord_schema/message_helpers_test.py ord_schema/scripts/` — 109/109 pass (no regressions)
- [x] `ruff check` + `ruff format --check` clean on new files
- [x] `ty` clean (enforced by pre-commit)
- [ ] Manual smoke test: migrate one real dataset from `../ord_data` and verify `read_dataset` round-trips the reactions (reviewer spot-check)


bdeadman (approved):

skearnes:
Follow-up note: **dropping planned follow-up #1** ("Integrate `.parquet` into `message_helpers.load_message` / `write_message`").

The write side already has `message_helpers.write_dataset` (#794), which is a thin suffix-dispatching wrapper. Going further — routing `.parquet` through `load_message` itself — would force one of two bad outcomes:

- Materialize the whole Parquet file into a `Dataset` proto, defeating the memory win and hiding a potentially 100× cost behind an innocent-looking `load_message(path, Dataset)` call.
- Return a `DatasetView` instead of a `Dataset`, silently breaking the type contract — existing callers would lose `SerializeToString()`, mutability, etc.

Two-seam world is more honest: `message_helpers` for the proto formats, `parquet_dataset` (or `DatasetView` / `iter_reactions`) when you want streaming or random access. Suffix-dispatch on the read side would paper over that distinction.

Updating the list — remaining open follow-ups are #3 (stream ORM ingestion), #4 (bulk-convert ord_data), and #5 (cutover story); #2 and #6 shipped via #794/#796/#801.

PR #793, opened 2026-04-24
Route Parquet through the submission pipeline end-to-end

skearnes:
## Summary

Follow-ups 1–3 from [#790](https://github.com/open-reaction-database/ord-schema/pull/790), bundled in one PR:

1. **`.parquet` in `message_helpers.load_message` / `write_message`.** Dispatches to `parquet_dataset` when the suffix is `.parquet`; Dataset-only and `.parquet.gz` is rejected.
2. **Streaming validation + end-to-end streaming `process_dataset.py`.**
3. **Streaming ORM ingestion in `add_datasets.py`.**

Everything on the `.parquet` path deserializes one Reaction at a time; peak memory stays bounded regardless of dataset size. Non-Parquet inputs are unchanged.

## Key pieces

- `validations.validate_parquet_datasets` — streams Reactions, accumulates cross-reference state in a single pass, writes per-file `.error` output when `write_errors=True`.
- `parquet_dataset` additions:
  - `iter_reaction_ids(source)` — streams just the ID column (no Reaction deserialization).
  - `ParquetSource` on `read_metadata` / `iter_reactions` (`str | bytes | IO`) so a `git show` blob is usable without a tempfile.
  - `streaming_md5(path)` — canonical hash over scalars + serialized Reactions in file order, for change detection during ingest.
- `process_dataset.py` for Parquet inputs:
  - Streaming validation + per-Reaction `--max_size` check.
  - Streaming diff stats via `iter_reaction_ids` on both the current file and the base (`git show`) bytes.
  - Two-pass streaming update pipeline: pass 1 assigns Reaction IDs and writes to a temporary Parquet file next to the output; pass 2 rewrites cross-references with the substitution map and writes the final file.
  - `--output_format .parquet` is required for Parquet inputs when `--update` is set (documented in the error message).
- `database.add_parquet_dataset` — inserts the `Dataset` row up front, streams Reactions in `flush` + `expunge` batches of 200, patches `md5` / `num_reactions` at the end. `add_datasets.py` uses `streaming_md5` for change detection and routes to the streaming path for `.parquet` inputs.

### MD5 caveat for ORM ingestion

[...]

PR #794, opened 2026-04-24
Stream Parquet validation via a DatasetView wrapper; add message_helpers.write_dataset

skearnes:
## Summary

Streams validation for Parquet-serialized Datasets, plus a top-level `write_dataset` helper that dispatches by filename suffix. This is the `validate_dataset.py` piece of follow-up #2 from [#790](https://github.com/open-reaction-database/ord-schema/pull/790).

### What's in here

- **`parquet_dataset.DatasetView`** — a read-only, streaming stand-in for `dataset_pb2.Dataset`. Exposes `name`, `description`, `dataset_id` from the Parquet footer (read once on construction); `reaction_ids` is always empty (Parquet doesn't persist it); `reactions` is a read-only `@property` over a `_ReactionStream` that:
  - opens a fresh read on each iteration so callers stay memory-bounded,
  - reports `len()` / `bool()` from the footer's `num_rows`, so emptiness checks behave like a list instead of being always-truthy the way a bare generator would be.
- **`_validate_datasets`** now calls `validate_dataset` directly inside `catch_warnings` for the Dataset-level check instead of going through `validate_message(dataset, recurse=False)`. `validate_message` insists on a proto type (`_VALIDATOR_SWITCH` lookup, `DESCRIPTOR` access) and would reject a non-proto stand-in like `DatasetView`; the direct call preserves the same warning-capture behavior for both real `Dataset`s and `DatasetView`s.
- **`validate_dataset.py::run`** wraps `.parquet` inputs in `DatasetView` before handing off to `validate_datasets`, so the streaming path shares the whole validation body with in-memory `Dataset`s.
- **`message_helpers.write_dataset(dataset, path)`** — dispatches by suffix: `.parquet` routes to `parquet_dataset.write_dataset`, everything else to `write_message`. Gives callers one place to write a Dataset regardless of format.

### Not in this PR

- **Routing `.parquet` through `message_helpers.load_message`.** `DatasetView` isn't a proto, so returning one from `load_message` would break its generic `MessageType` contract for callers that mutate or `SerializeToString`. Callers that want streaming use `parquet_dataset.DatasetView` directly.
- **`process_dataset.py` Parquet support** (rest of follow-up #2) — deferred; will land as a separate PR.

## Test plan

- [x] `uv run pytest ord_schema/` — 312/312 pass (Postgres on `PATH` for ORM tests).
- [x] `test_dataset_view_exposes_all_dataset_fields` — regression guard: fails if a Dataset proto field is added without a matching attribute on `DatasetView`.
- [x] `test_dataset_view_empty_parquet_is_falsy` — guards the `__len__`/`__bool__` contract so `validate_dataset`'s "Dataset requires reactions or reaction_ids" check still fires for empty Parquet.
- [x] `test_dataset_view_reactions_is_read_only` — pins the read-only `@property` so accidental rebinding raises.
- [x] `test_dataset_view_values_round_trip` — scalars, `len(reactions)`, full iteration, and re-iteration.
- [x] `validate_dataset_test::setup` parametrized over `(.pbtxt, .parquet)` — existing `test_simple` / `test_filter` / `test_validation_errors` each run against both formats.
- [x] `test_write_dataset` parametrized over all 5 write formats, reading `.parquet` back via `DatasetView`.
- [x] `ruff check` + `ruff format --check` clean; `ty` clean via pre-commit.

Supersedes the over-scoped #793 (closed).

### DIFF
--- a/ord_schema/message_helpers.py
+++ b/ord_schema/message_helpers.py
@@ -35,7 +35,7 @@ from rdkit import Chem
 from rdkit.Chem import rdChemReactions
 
 import ord_schema
-from ord_schema import parquet_dataset, units
+from ord_schema import units
 from ord_schema.proto import dataset_pb2, reaction_pb2
 
 _COMPOUND_IDENTIFIER_LOADERS = {
@@ -727,7 +727,6 @@ class MessageFormat(enum.Enum):
     BINARY = ".pb"
     JSON = ".json"
     PBTXT = ".pbtxt"
-    PARQUET = ".parquet"
 
 
 def fetch_dataset(dataset_id: str, timeout: float = 10.0) -> dataset_pb2.Dataset:
@@ -776,12 +775,6 @@ def load_message(filename: str, message_type: type[MessageType]) -> MessageType:
         this_open = open
         _, extension = os.path.splitext(filename)
     input_format = MessageFormat(extension)
-    if input_format == MessageFormat.PARQUET:
-        if filename.endswith(".gz"):
-            raise ValueError(f"Parquet files are not gzip-wrapped: {filename}")
-        if message_type is not dataset_pb2.Dataset:
-            raise ValueError(f"Parquet is only supported for Dataset messages, not {message_type.__name__}")
-        return parquet_dataset.read_dataset(filename)  # ty: ignore[invalid-return-type]
     if input_format == MessageFormat.BINARY:
         mode = "rb"
     else:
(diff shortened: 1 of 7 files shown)


ord_schema/message_helpers.py
```diff
@@ -826,6 +826,18 @@ def write_message(message: ord_schema.Message, filename: str):
...
+def write_dataset(dataset: dataset_pb2.Dataset, filename: str) -> None:
+    """Writes a Dataset to disk, dispatching on filename suffix.
+
+    ``.parquet`` routes to ``parquet_dataset.write_dataset``; other suffixes
+    go through ``write_message``.
+    """
+    if filename.endswith(".parquet"):
+        parquet_dataset.write_dataset(dataset, filename)
+        return
+    write_message(dataset, filename)
```
bdeadman, line 838: Is this correct or should the 'write_message' call be under an 'else' statement? It looks like writing a '.parquet' dataset will route through both 'parquet_dataset.write_dataset' and 'write_message'. The 'write_message' should then return a ValueError due to the unexpected suffix.

bdeadman (approved):
I've put a comment to check on the write_message helper. I may be misunderstanding how the helper works now so please ignore it if I'm wrong.

Otherwise happy to approve PR.

ord_schema/message_helpers.py
skearnes: Good question; the `return` will break out of the function so an else wouldn't add anything.

PR #814, opened 2026-05-12
Dataset Rewards program docs page

bdeadman:
- Adds the Dataset Rewards program page to docs site.
- Adds a placeholder page for newsletter with link to LinkedIn newsletter.


docs/events/rewards.rst
```diff
@@ -0,0 +1,315 @@
+####################
```
skearnes, line 1: I'm wondering if we should have docs like this live on LinkedIn so they can be easily updated, receive comments, etc. and just have the pointer above in the repo?

bdeadman: I agree that very short term events are more easily kept on LinkedIn, but for longer running events I don't think LinkedIn does a good job of keeping them visible and findable. I find LinkedIn posts get lost in the organisation history. Newsletter editions have more of a presence and are easier to find later on, but I'd like to send out the newsletter more regularly so there is a risk that a Dataset Rewards newsletter will be less visible.

I'm expecting the Dataset Rewards event page to be online (even if only to advise that it is closed to new applications) for 2026 at least. I would rather periodically advertise it on LinkedIn and push any interest to this ORD docs page, where we can use version control to update the info as and when required. This also has the advantage of exposing viewers to the ORD website and documentation which will be helpful to them as they jump into preparing and submitting their datasets.

The newsletter page will be just a link pointing to LinkedIn for now. I'm toying with the idea of periodically archiving a list of links to specific LinkedIn articles on the Newsletter docs page.

skearnes (approved):

docs/events/rewards.rst
```diff
@@ -0,0 +1,315 @@
...
+Application Process
+************************
+
+To participate in the Dataset Reward program please do the following:
+
+1. Check that you meet the :ref:`applicant-eligibility` criteria
+2. Select a suitable paper or source dataset that you will turn into an ORD dataset. The paper/dataset must meet the :ref:`dataset-eligibility` criteria.
+3. Complete a Dataset Reward application form (`link <https://forms.gle/uAeqGwVak1bti7EC9>`__). As part of the application you will be required to agree to the following:
+
+   - Independent Contractor Agreement (`link <https://github.com/open-reaction-database/ord-schema/blob/325d4c67c57e30f3d10a32462bf765c48ca63136/docs/documents/STF-%20Data%20Curators%20Agreement.pdf>`__)
```
skearnes: let's put this in google drive

bdeadman: Agreed, and done.

PR #820, opened 2026-06-12
Speed up incremental RDKit phase in add_datasets

skearnes:
## Summary

Speeds up the incremental "Adding RDKit functionality" phase, which dominated ORM backend updates — a no-op dataset (everything already loaded) was taking ~60s+ almost entirely on work to re-confirm there was nothing to do.

### `ord_schema/orm/database.py` — query rewrites
1. **`EXCEPT` → `NOT EXISTS`** in `_update_rdkit_reactions`/`_update_rdkit_mols`: per-candidate probes of the unique `smiles`/`reaction_smiles` indexes instead of scanning all of `rdkit.{mols,reactions}`.
   - **Materialization barrier (mols):** the new SMILES are resolved in a `WITH new_smiles AS MATERIALIZED (...)` CTE so the cheap anti-join runs first and the expensive `mol_from_smiles`/Morgan-fingerprint calls run only on survivors. (Without it, the `WHERE mol IS NOT NULL` guard let the planner run those functions on every candidate → ~240s; reactions doesn't need it because `SELECT DISTINCT` already barriers.)
   - **NULL keys:** `reaction_smiles` is nullable; `NOT EXISTS` (unlike `EXCEPT`) never matches NULL, so the reactions query adds `AND reaction_smiles IS NOT NULL` to stay idempotent. Mols already drops NULL via `NOT LIKE '%[Ti+5]%'`.
2. **Flatten the `update_rdkit_ids` UPDATEs** to `UPDATE ... FROM` (drops a materialize + self-rejoin-by-id; unique join keys → no row multiplication).
3. **NULL cartridge guards** — `WHERE mol IS NOT NULL` / `WHERE reaction IS NOT NULL` keep junk NULL rows out and NULL out of the fingerprint functions (symmetric).

### `ord_schema/orm/mappers.py` — partial indexes (the no-op short-circuit)
After the query rewrites, an already-loaded dataset still spent ~12s (mols anti-join) + ~11s (ID `UPDATE`s) just confirming "nothing to link" — inherent if you must scan every reaction/compound in the dataset. Add **partial indexes over only the unlinked rows**, keyed by the FK used to reach a dataset:

```
reaction_unlinked_index         ON ord.reaction (dataset_id)                  WHERE rdkit_reaction_id IS NULL
compound_unlinked_index         ON ord.compound (reaction_input_id)           WHERE rdkit_mol_id IS NULL
product_compound_unlinked_index ON ord.product_compound (reaction_outcome_id) WHERE rdkit_mol_id IS NULL
```

They stay tiny (only in-flight datasets have unlinked rows), so the no-op case becomes ~O(unlinked) instead of O(dataset size). Chosen over an outright "skip add_rdkit for unchanged datasets" because it's **always correct** — the queries are unchanged, so self-healing after a failed prior `add_rdkit` is preserved. `create_all` builds them on new databases.

### Existing databases — manual migration
`create_all` does **not** add indexes to existing tables, so a live database needs them created manually, online:

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS reaction_unlinked_index
    ON ord.reaction (dataset_id) WHERE rdkit_reaction_id IS NULL;
CREATE INDEX CONCURRENTLY IF NOT EXISTS compound_unlinked_index
    ON ord.compound (reaction_input_id) WHERE rdkit_mol_id IS NULL;
CREATE INDEX CONCURRENTLY IF NOT EXISTS product_compound_unlinked_index
    ON ord.product_compound (reaction_outcome_id) WHERE rdkit_mol_id IS NULL;
ANALYZE ord.reaction; ANALYZE ord.compound; ANALYZE ord.product_compound;
```
`CONCURRENTLY` keeps the tables writable during the build and **cannot run inside a transaction block** — run each statement standalone. If a build is interrupted it leaves an `INVALID` index; `DROP INDEX CONCURRENTLY` and retry.

## Testing

ORM suite (RDKit cartridge + Postgres) passes: `database_test`, `mappers_test`, `updates_test`, `rdkit_mappers_test`, `add_datasets_test`. Highlights: `test_update_rdkit_tables_idempotent` (re-run inserts nothing, no NULL cartridge rows — fails without the NULL-`reaction_smiles` guard), `test_unlinked_partial_indexes` (the partial indexes exist with the expected predicates), `test_product_compound_rdkit_link` (completeness of the product_compound link path).

Live timing on a no-op (already-loaded) dataset, confirmed on the live DB, with all three layers (NOT EXISTS + materialization barrier + partial indexes): the whole RDKit phase drops to ~0.3s — reactions ~0.003s, mols ~0.26s (was ~240s mid-PR), IDs ~0.027s (was ~11s). The ID step confirms the planner drives from the partial indexes (no per-dataset compound scan).

<h3>Greptile Summary</h3>

This PR rewrites the incremental RDKit linking phase in `add_datasets` to eliminate the bottleneck that caused a no-op dataset (already fully loaded) to spend 60s+ confirming nothing had changed. The changes are semantically equivalent but query-plan aware.

- **`EXCEPT` → `NOT EXISTS`** in both `_update_rdkit_reactions` and `_update_rdkit_mols`: probes the unique `reaction_smiles`/`smiles` indexes per candidate instead of scanning all of `rdkit.{reactions,mols}`; a `MATERIALIZED` CTE in the mols path forces the cheap anti-join to complete before the expensive `mol_from_smiles`/fingerprint calls run. NULL guards (`reaction_smiles IS NOT NULL`, `WHERE reaction IS NOT NULL`, `WHERE mol IS NOT NULL`) keep junk rows out of the cartridge tables and were an open correctness gap in the old code.
- **Flatten `update_rdkit_ids`** from a materialize-then-rejoin subquery to a flat `UPDATE … FROM`, extracting the shared compound/product_compound pattern into `_link_mol_ids`. The three partial indexes in `mappers.py` (`reaction_unlinked_index`, `compound_unlinked_index`, `product_compound_unlinked_index`) index only in-flight (unlinked) rows, so the no-op case becomes O(unlinked) ≈ 0 instead of O(dataset size).
- Migration guidance (manual `CREATE INDEX CONCURRENTLY` for existing databases) is documented in `README.md` and kept in sync with the `Index()` declarations.

Safe to merge. The query rewrites are semantically equivalent to the old code; the correctness gaps (NULL-mol/reaction rows in cartridge tables) are fixed, not introduced. The partial indexes are additive and create_all handles them automatically for new databases.

All changed queries have been analyzed for row-multiplication (none — join keys are unique-indexed), idempotency (the new IS NOT NULL guards and WHERE mol/reaction IS NOT NULL correctly prevent junk re-insertion), and the MATERIALIZED barrier correctly prevents premature fingerprint evaluation. The new tests directly exercise the fixed corner cases. Existing databases require a manual CONCURRENTLY migration, which is clearly documented.

No files require special attention. Reviewers comfortable with PostgreSQL UPDATE…FROM semantics should verify the flat FROM clause in update_rdkit_ids and _link_mol_ids — the key insight is that all join keys (reaction_smiles, smiles, dataset_id) carry unique constraints, so each target row matches at most one tuple from the FROM set.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| ord_schema/orm/database.py | Core query rewrites: EXCEPT→NOT EXISTS with MATERIALIZED CTE barrier, NULL-reaction/mol guards, and flat UPDATE…FROM. No row-multiplication risk (join keys are unique-indexed). _link_mol_ids uses f-string interpolation for trusted hardcoded table/column identifiers with dataset_id as a bind param. |
| ord_schema/orm/mappers.py | Adds three partial indexes (reaction/compound/product_compound unlinked rows) registered on Base.metadata so create_all builds them on new databases. Indexes correctly keyed on the FK used to reach a dataset and predicated on rdkit_*_id IS NULL. |
| ord_schema/orm/database_test.py | Adds test_update_rdkit_tables_idempotent (guards the NOT EXISTS + IS NOT NULL rewrite) and test_unlinked_partial_indexes (verifies the three partial indexes exist with correct predicates). Both are meaningful regression tests for the new behaviour. |
| ord_schema/orm/rdkit_mappers_test.py | Adds test_product_compound_rdkit_link to cover the product_compound linking path that was missing from operator tests. Completeness check confirms no linkable product_compound rows are left unlinked. |
| ord_schema/orm/README.md | Adds migration section documenting CREATE INDEX CONCURRENTLY statements and cross-referencing the Index() declarations in mappers.py. Documentation is accurate and the CONCURRENTLY caveat (cannot run in a transaction) is correctly noted. |

</details>

</details>

<details><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `ord_schema/orm/database.py`, line 209-244 ([link](https://github.com/open-reaction-database/ord-schema/blob/f2c5509861b2a5f55c324086834e710e5415d11c/ord_schema/orm/database.py#L209-L244)) 

   <a href="#">P2</a> **Missing NULL-mol guard before fingerprint computation**

   `mol_from_smiles` returns `NULL` for any SMILES string it cannot parse. The outer `SELECT` immediately feeds that `NULL` into `morganbv_fp(mol)` and `morgan_fp(mol)` without a `WHERE mol IS NOT NULL` guard. Passing a NULL `mol` to those functions may silently propagate NULLs into `morgan_bfp`/`morgan_sfp` columns (or raise an error depending on the cartridge build), and the `ON CONFLICT (smiles) DO NOTHING` clause will not prevent the bad row from being inserted if the SMILES is genuinely new.

   This is a pre-existing issue in the old code too, but the refactor flattened the three subquery layers into two, making it a good moment to add the guard. Consider adding `WHERE mol IS NOT NULL` in `mol_subquery` between the `NOT EXISTS` filter and the outer fingerprint computation.
</details>

<sub>Reviews (9): Last reviewed commit: ["Merge branch &#39;main&#39; into speedup-add-rdk..."](https://github.com/open-reaction-database/ord-schema/commit/97e829753c0171eb7407adcb7bd6656b246e996d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37145298)</sub>


skearnes:
Addressed the Greptile P2 (NULL-mol guard) in da0cef4 — added `WHERE mol IS NOT NULL` in `_update_rdkit_mols` so an unparseable, genuinely-new SMILES no longer inserts a NULL-mol row or feeds NULL into `morganbv_fp`/`morgan_fp`. As noted, it was pre-existing; the `NOT EXISTS` refactor just made it visible.

skearnes:
@greptileai review

skearnes:
@greptileai review

skearnes:
@greptile review

bdeadman (approved):

skearnes:
Re: Greptile's P2 "Missing NULL-mol guard before fingerprint computation" (Comments Outside Diff) — this is already handled in the current revision (14e8445).

In `_update_rdkit_mols`, `mol_from_smiles` is computed in the inner `computed` subquery, and the outer `INSERT ... SELECT` that calls `morganbv_fp(mol)`/`morgan_fp(mol)` carries `WHERE mol IS NOT NULL`:

```sql
SELECT smiles, mol, morganbv_fp(mol) AS morgan_bfp, morgan_fp(mol) AS morgan_sfp
FROM (
    SELECT smiles, mol_from_smiles(smiles::cstring) AS mol FROM new_smiles
) computed
WHERE mol IS NOT NULL
ON CONFLICT (smiles) DO NOTHING
```

Postgres applies `WHERE` before evaluating the select list, so the fingerprint functions never receive a NULL `mol`, and an unparseable SMILES can't insert a junk NULL-mol row (the case `ON CONFLICT (smiles) DO NOTHING` wouldn't catch for a genuinely new SMILES). This matches Greptile's own summary, which notes the concern "has been addressed in this revision." No change needed.

PR #915, opened 2026-07-30
Run the Python matrix in one job per OS, and fix CI bugs found alongside

skearnes:
## Why

The Python matrix fanned out to **14 jobs** (8 `test_ord_schema` + 6 `test_notebooks`), each one repeating the conda `rdkit-postgresql` install and a `uv sync`. Measured on the last `main` run: **21 jobs, 47 runner-minutes, 6.2 min wall clock.**

## Change

Loop over the interpreters inside **one job per OS**. `uv sync --python X` fetches and switches interpreters, so `actions/setup-python` is no longer needed.

|  | before | after |
|---|---|---|
| PR jobs | 21 | **11** |
| runner-minutes | ~47 | **~28** |
| wall clock | 6.2 min | ~8 min |

The wall clock goes *up* slightly, because the versions now run serially instead of in parallel — that is the trade. Queue wait improves and no test coverage is lost. `fail-fast` is off now that one job carries every version's result, and the loop records failures instead of stopping at the first, so one bad interpreter no longer hides the rest.

Also added `concurrency` so a superseded PR run is cancelled rather than running a full stale matrix to completion, and `permissions: contents: read` at the workflow level. `main` runs are never cancelled — each push's result is the record for that commit.

## Bugs found while auditing the other jobs

Each of these was verified locally rather than assumed:

- **`test_ord_schema` reported success when tests failed.** A bare `shell:` string is used verbatim, so GitHub does not add `-eo pipefail`; `uv run coverage report` ran last and set the step's exit code. Reproduced: `bash -l -c 'true; false; true'` exits **0**. Failures are now captured explicitly, and `Setup PostgreSQL` sets `-eo pipefail` so a failed conda install cannot be masked by `initdb`.
- **The proto-wrapper drift check could not see a wrapper generated for a new `.proto`.** `git diff` reports only tracked modifications, so a newly generated (untracked) file read as *no drift*. Now stages first and diffs `--cached`; confirmed `-I` still ignores copyright-year-only changes under `--quiet`.
- **`check_markdown` never linted `.claude/` or `.github/`.** `**/*.md` is expanded by globby, which skips dotfiles by default — so `SKILL.md` and both issue templates were unchecked in CI, while the pre-commit hook (which receives explicit filenames) did check them. Now takes the list from `git ls-files`, which also excludes `.venv/` for free. This brings 3 previously-invisible files under CI.
- **`addlicense` floated on `@latest`** in `test_proto_wrappers` while `cleanup.yml` pins `v1.2.0` — in the job that *writes* the headers `cleanup.yml` *checks*. Now pinned to match.
- **`PGDATA` was a literal `$GITHUB_WORKSPACE/...`**, since `env:` values expand only `${{ }}`. `initdb` was creating a directory named after the variable.

`check_proto` was suspect for the same reason (`find -exec … {} +`) but is **sound** — the `+` form does propagate a non-zero exit; only the `\;` form swallows it.

## Note

`check_markdown` may now fail on files it was never linting. If so that is the fix working, and I will clean up whatever it reports.

This PR consolidates Python CI matrices by OS and adds flat Parquet reaction views.
- Serializes supported Python versions within each OS job while preserving per-version failure reporting.
- Tightens workflow permissions, cancellation, PostgreSQL setup, Markdown coverage, and generated-wrapper drift detection.
- Adds canonical unit conversions for concentration and mercury pressure units.
- Adds atomic, stamped Parquet view generation with a glob-driven CLI and tests.
</details>

<h3>Confidence Score: 4/5</h3>

The unsupported-unit failure should be fixed before merging because one forward-compatible enum value can abort an entire view derivation.

The new projection directly invokes dictionary-backed unit conversion without handling unknown enum values, contradicting its null-on-unconvertible contract and turning one incompatible measurement into a dataset-wide failure.

**Files Needing Attention:** ord_schema/views.py

| Filename | Overview |
|----------|----------|
| .github/workflows/run_tests.yml | Consolidates Python-version matrices into per-OS loops and fixes failure propagation and generated-wrapper drift checks. |
| .github/workflows/cleanup.yml | Adds constrained permissions and concurrency while expanding Markdown linting to tracked dot-directory files. |
| ord_schema/units.py | Adds concentration conversion factors and torr/mmHg parsing and conversion support. |
| ord_schema/views.py | Adds atomic flat-view derivation and freshness stamps, but an unsupported unit aborts derivation instead of producing the documented null. |
| ord_schema/scripts/derive_views.py | Adds a tested CLI that maps source globs into mirrored output paths and skips current views. |

</details>

```mermaid
flowchart LR
  Source[Source Parquet dataset] --> Hash[Compute content MD5]
  Hash --> Current{Existing view current?}
  Current -->|Yes| Skip[Skip derivation]
  Current -->|No| Stream[Stream source row groups]
  Stream --> Project[Project reactions to flat rows]
  Project --> Temp[Write temporary Parquet view]
  Temp --> Stamp[Attach source and version stamps]
  Stamp --> Publish[Atomically replace destination]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["Unpack protoc outside the checkout so th..."](https://github.com/open-reaction-database/ord-schema/commit/c4a36f571a4d4eaa6cbdaca46fc5694fde20ad93) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48548360)</sub>

> Greptile also left **1 inline comment** on this PR.

<details><summary><h4>Context used (3)</h4></summary>

- Knowledge Base — [CI Tooling and Repository Automation Scripts](https://app.greptile.com/open-reaction-database/-/custom-context/knowledge-base/open-reaction-database/ord-schema/-/docs/ci-tooling.md)
- Knowledge Base — [Data I/O: proto serialization, Parquet, and units](https://app.greptile.com/open-reaction-database/-/custom-context/knowledge-base/open-reaction-database/ord-schema/-/docs/data-io.md)
- Knowledge Base — [Dataset Scripts](https://app.greptile.com/open-reaction-database/-/custom-context/knowledge-base/open-reaction-database/ord-schema/-/docs/dataset-scripts.md)
</details>


skearnes:
Superseded by #916. This branch was cut from `views` rather than `main`, so it carried four unrelated files (`ord_schema/views.py`, `views_test.py`, `scripts/derive_views*.py`) and the views commit history. #916 is the same two workflow files committed directly on top of `main`.

PR #925, opened 2026-08-02
Parse each structural identifier once, not four times

skearnes:
## Summary

Profiling ord-data's corpus sweep found half of validation sitting in `Chem.MolFromInchi` — not because InChI parsing is slow, but because the same string is parsed twice per reaction by two call sites that don't know about each other, and the same molecules recur across a dataset's reactions. On the uspto grants file that's 15,306 `MolFromInchi` calls for 5,336 InChI identifiers. DateTimes have the identical shape: parsed by the per-message walk in `validate_date_time`, then again by `validate_reaction_provenance` to order them.

Memoizing by string is **1.785 → 0.670 ms/reaction, 2.66×**, measured on `data/11/ord_dataset-1158e351757f315b93cbcbe7bc55f38e.parquet`. Full analysis in [the logbook entry](https://github.com/open-reaction-database/ord-logbook/blob/main/entries/2026-08-02-validation-performance.md).

## Changes

- `canonical_smiles_for_identifier(type, value)` memoizes the derived SMILES; `validate_compound_identifier` and `check_compound_identifiers` both read it instead of parsing separately.
- The cached value is a **string, never the `Mol`** — an RDKit `Mol` is mutable, so handing the same one to every caller would let any of them corrupt the entry.
- `identifier_parses_unsanitized` is a separate cache for the `sanitize=False` retry, so that extra parse happens only for identifiers that already failed.
- `_parse_date_time` does the same for DateTime strings. Caching is safe outright here because `datetime` is immutable.
- Caches are bounded: a large dataset's distinct set doesn't saturate (uspto grows ~1.3 new molecules per reaction), and these hold strings, so the ceiling is tens of MB.

## Testing

`pytest -n auto` — 612 passed (excluding `orm/`, which needs Postgres and fails the same way on `main`). ruff check/format, `ty` clean on the touched files with the 41 repo-wide diagnostics unchanged, `pre-commit` on all four files.

Behavior preservation leans on the existing tests, which already pin "could not validate", "could not sanitize", and identifier inconsistency, and which pass unchanged. Two equivalences I verified rather than assumed: `not mol` is exactly `mol is None` for RDKit (no `__bool__`/`__len__`, so an empty `Mol` is truthy), and `datetime.ctime()` cannot raise anywhere in dateutil's parseable year range, so dropping it from `validate_date_time` changes nothing.

New tests cover what the cache newly makes breakable: that the memoized result matches parsing by hand, that the key includes the identifier type (an InChI string must not answer for a SMILES lookup), that the two checks share one parse rather than repeating it, and that a warm cache reports the same findings as a cold one.

## Notes

`validate_reaction_provenance` keeps its short-circuit: the original `try` block stops at the first unparseable timestamp and skips the ordering comparisons after it. My first pass made it stricter by parsing all three; since this PR is meant to be behavior-preserving, I restored the original semantics with a small raising wrapper. Making it stricter is worth doing, but deliberately and separately.

The PR memoizes structural-identifier canonicalization and DateTime parsing so validation call sites can reuse deterministic parse results.
- Adds bounded caches for canonical structural identifiers and unsanitized parsing results.
- Routes compound validity and consistency checks through the shared identifier cache.
- Adds a fixed DateTime parsing default and shares cached timestamps with provenance ordering.
- Adds cache-key, behavior-preservation, and parse-reuse tests.
</details>

The PR appears safe to merge.

No blocking failure remains within the eligible follow-up review scope.

| Filename | Overview |
|----------|----------|
| ord_schema/message_helpers.py | Adds bounded identifier parsing caches and reuses canonicalized strings during compound consistency checks. |
| ord_schema/validations.py | Reuses cached identifier and DateTime parsing while preserving existing validation and provenance short-circuit paths. |
| ord_schema/message_helpers_test.py | Tests canonicalization equivalence, type-sensitive cache keys, invalid identifiers, and unsanitized fallback behavior. |
| ord_schema/validations_test.py | Tests cache reuse, warm-cache equivalence, and deterministic handling of partial timestamps. |

</details>

```mermaid
flowchart LR
  Identifier[Structural identifier] --> IdentifierCache[Canonical identifier cache]
  IdentifierCache --> Validity[Identifier validity check]
  IdentifierCache --> Consistency[Compound consistency check]
  DateTime[DateTime string] --> DateCache[DateTime parse cache]
  DateCache --> DateValidation[DateTime validation]
  DateCache --> Provenance[Provenance ordering]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["Distill the comments added for the DateT..."](https://github.com/open-reaction-database/ord-schema/commit/2cfb95137680a72ad7ea46114c3213cb212d9d0b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49528497)</sub>

**Context used:**

- Knowledge Base — [Validation, Resolution, and Update Pipeline](https://app.greptile.com/open-reaction-database/-/custom-context/knowledge-base/open-reaction-database/ord-schema/-/docs/validation-pipeline.md)

### DIFF
--- a/ord_schema/message_helpers.py
+++ b/ord_schema/message_helpers.py
@@ -67,6 +67,10 @@ STRUCTURAL_IDENTIFIER_TYPES = frozenset(
 # Bounded because a large dataset's distinct set does not saturate (uspto grows
 # by ~1.3 new molecules per reaction); this holds strings, so the ceiling is
 # tens of MB.
+#
+# Entries assume RDKit parses and writes the same way for the life of the process.
+# A caller that flips a global mid-run (Chem.SetUseLegacyStereoPerception and the
+# like) has to call cache_clear() on both functions below.
 _IDENTIFIER_CACHE_SIZE = 100_000
 
 MessageType = TypeVar("MessageType")  # Generic for setting return types
--- a/ord_schema/message_helpers.py
+++ b/ord_schema/message_helpers.py
@@ -15,6 +15,7 @@
 
 import contextlib
 import enum
+import functools
 import gzip
 import os
 import pathlib
@@ -56,7 +57,60 @@ STRUCTURAL_IDENTIFIER_TYPES = frozenset(
     reaction_pb2.CompoundIdentifier.CompoundIdentifierType.Name(identifier_type)
     for identifier_type in _COMPOUND_IDENTIFIER_LOADERS
 )
+# Structural identifiers repeat heavily -- shared solvents, reagents and
+# catalysts recur across the reactions of a dataset -- and validation parses
+# each one twice, once to check that it parses and once to canonicalize it for
+# the consistency check. Memoizing the derived SMILES collapses both.
+#
+# The cached value is a string, never the Mol: an RDKit Mol is mutable, so
+# handing the same one to every caller would let any of them corrupt the entry.
+# Bounded because a large dataset's distinct set does not saturate (uspto grows
+# by ~1.3 new molecules per reaction); this holds strings, so the ceiling is
+# tens of MB.
+_IDENTIFIER_CACHE_SIZE = 100_000
+
 MessageType = TypeVar("MessageType")  # Generic for setting return types
+
+
+@functools.lru_cache(maxsize=_IDENTIFIER_CACHE_SIZE)
+def canonical_smiles_for_identifier(identifier_type: int, value: str) -> str | None:
+    """Canonicalizes a structural identifier, or returns None if it will not parse.
+
+    Args:
+        identifier_type: CompoundIdentifier type enum value.
+        value: The identifier value.
+
+    Returns:
+        Canonical SMILES, or None if ``value`` is not a valid identifier of that
+        type. Returns None for types no Mol can be built from.
+    """
+    loader = _COMPOUND_IDENTIFIER_LOADERS.get(identifier_type)
+    if loader is None:
+        return None
+    mol = loader(value)
+    if mol is None:
+        return None
+    return canonical_smiles(mol)
+
+
+@functools.lru_cache(maxsize=_IDENTIFIER_CACHE_SIZE)
+def identifier_parses_unsanitized(identifier_type: int, value: str) -> bool:
+    """Tests whether an identifier parses at all once sanitization is skipped.
+
+    Separate from :func:`canonical_smiles_for_identifier` so the extra parse
+    happens only for the identifiers that failed, which are rare.
+
+    Args:
+        identifier_type: CompoundIdentifier type enum value.
+        value: The identifier value.
+
+    Returns:
+        True if ``value`` parses with ``sanitize=False``.
+    """
+    loader = _COMPOUND_IDENTIFIER_LOADERS.get(identifier_type)
+    if loader is None:
+        return False
+    return loader(value, sanitize=False) is not None
 
 
 def _resolve_enum_value(descriptor: Descriptor, field_name: str, key: str) -> int:
@@ -421,15 +475,14 @@ def check_compound_identifiers(
     """
     smiles = set()
     for identifier in compound.identifiers:
-        if identifier.type in _COMPOUND_IDENTIFIER_LOADERS:
-            mol = _COMPOUND_IDENTIFIER_LOADERS[identifier.type](identifier.value)
-        else:
+        if identifier.type not in _COMPOUND_IDENTIFIER_LOADERS:
             continue
-        if not mol:
+        canonical = canonical_smiles_for_identifier(identifier.type, identifier.value)
+        if canonical is None:
             raise ValueError(
                 f"invalid structural identifier for Compound: {identifier}"
             )
-        smiles.add(canonical_smiles(mol))
+        smiles.add(canonical)
     if len(smiles) > 1:
         raise ValueError(f"structural identifiers are inconsistent: {smiles}")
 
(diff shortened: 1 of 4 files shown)


skearnes (commented):
## Overview

Memoizes the two expensive parses in validation — structural identifiers and DateTime strings — so the call sites that previously parsed independently share one `lru_cache` entry. The design is sound. Caching the *string* rather than the mutable `Mol` is exactly right, and splitting the `sanitize=False` retry into its own cache keeps the rare path off the hot path.

Rather than take the equivalence claims on faith, I diffed the old and new logic directly in a worktree of the branch.

## Verification

- **Identifier equivalence: 100 differential comparisons** (4 types × 25 values — valid SMILES/CXSMILES/InChI/MolBlock, `""`, whitespace, malformed InChI, pentavalent nitrogen, a 200-char chain, stereo, salts), comparing the deleted `parse_func` dict against the new cached helpers. **0 mismatches, 0 new-side exceptions** — the `canonical_smiles` call newly added to the validity path never raised.
- **`not mol` ≡ `mol is None`: confirmed.** `Chem.Mol` defines neither `__len__` nor `__bool__`, so the empty Mol from `MolFromSmiles("")` is truthy and both paths yield `""`.
- **Dropping `.ctime()`: confirmed safe.** No failures for years 1 through 9999.
- **Provenance short-circuit: preserved.** `_parse_date_time_or_raise` reproduces dateutil's lazy `%s` formatting, and the first-failure exit is intact.

## Finding: `_parse_date_time` memoizes an impure function

`parser.parse` fills missing fields from *today*, so it is not a pure function of its input:

```
'10:30'    -> 2026-08-02 10:30:00   (date from now())
'March 5'  -> 2026-03-05 00:00:00   (year from now())
```

`validate_date_time` only checks parseability, so it is unaffected. But `validate_reaction_provenance` compares the parsed values by sign. Priming the cache on a different day reproduces a spurious error:

```
cold cache, both parsed today   : no ordering error
warm cache primed on another day: SPURIOUS ERROR -> Record creation time should be after experiment
```

This needs a partial timestamp *and* a process whose cache spans a day boundary — a multi-hour corpus sweep, or a long-lived server. Every DateTime in the test suite is a full date (`"2020-01-02"`, `"2023-07-01"`), so real ORD data almost certainly never hits it. Low severity, real mechanism, and worth recording either way since memoizing an impure function is the kind of thing that gets copied.

The clean fix is `parser.parse(value, default=<fixed epoch>)`, which makes the function genuinely pure — but that changes what partial values resolve to, so it belongs in a deliberate follow-up rather than in a behavior-preserving PR. A sentence in the `_parse_date_time` docstring would cover it for now.

## Test coverage gaps

- **The DateTime cache has no new test.** `test_a_warm_cache_does_not_change_what_validation_reports` calls `validations._parse_date_time.cache_clear()`, but its fixture is `_compound_with_two_identifiers()` — a Compound with no timestamps. That clear is dead code, and the datetime half of the PR rests entirely on pre-existing tests that never exercise a warm cache. A provenance-based cold-vs-warm assertion would close it, and would have caught the finding above.
- **MOLBLOCK is absent from the parametrize list** in `test_canonical_smiles_for_identifier_matches_parsing_directly` (SMILES, CXSMILES, INCHI only). It is the loader with the most distinct semantics and the only one now unpinned. The differential covered it and it is fine, but that coverage does not live in the suite.

## Smaller notes

- `test_canonical_smiles_for_identifier_returns_none_when_unparseable` ends up as `... ) is (None)` after ruff wrapping. Correct, but reads oddly; assigning to a local first would format cleanly.
- The caches are process-global and never cleared in library use. If a consumer flips an RDKit global mid-process (`SetUseLegacyStereoPerception` and friends), cached canonical SMILES go stale. Exotic enough to be worth at most a comment.
- `test_the_two_identifier_checks_share_one_parse` pins `misses == 2, hits == 2`. Brittle by design, and correctly so — worth knowing it will also fail if someone legitimately adds a third call site.

## Verdict

Looks good to merge, with the datetime docstring caveat. The identifier half is verified equivalent by differential testing and carries the performance win. The datetime half is correct for all realistic inputs but rests on an unstated purity assumption that no test currently guards.

_Analysis performed with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)._

skearnes:
Addressed the review findings in c84cc0c and 2cfb951.

**The impurity is fixed, not just documented.** `_parse_date_time` now passes a fixed `default=` to `parser.parse`, which makes it a pure function of its input. That is a narrower change than the fixed-epoch swap I proposed in the review: full timestamps supply every field and are untouched, and partial ones already had an arbitrary date, so the ordering comparisons that provenance actually performs are unchanged. The default is a leap year so `"Feb 29"` still parses, and naive so a partial value never comes back timezone-aware when a full one would not.

**The new test discriminates.** I checked it both ways rather than assuming: `test_partial_timestamps_do_not_inherit_the_current_date` fails against the unfixed code and passes with the fix. My first attempt at it did not — seeding the cache in the same run reproduces nothing, because the bug needs the clock to differ — so the test asserts the purity property directly instead of simulating a midnight crossing.

Also added: provenance ordering and shared-parse coverage for the DateTime cache (which no test reached), MOLBLOCK in the canonicalization parametrization, a note that the identifier caches assume RDKit behaves consistently for the life of the process, and the `is (None)` formatting cleanup. 616 passed, ruff and `ty` clean.

_Analysis performed with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)._

Issue #937, opened 2026-08-07
Rebuild derived.reaction_smiles and rdkit.reactions after the agent-free normalization


skearnes:
#935 changed what `_update_derived_tables` writes to `derived.reaction_smiles`, so the table and the `rdkit.reactions` index built from it are stale wherever they were populated before that commit.

What changed: the ORM now calls `message_helpers.derived_reaction_smiles` instead of `get_reaction_smiles`. For an existing row that means

- **agents are gone.** A reaction stored as `A>B>C` is now stored as `A>>C`.
- **recorded values are canonicalized** rather than written as deposited, so atom ordering is normalized. Atom mapping is preserved.
- **a previously-truncated value is fixed.** The old code stripped a CXSMILES extension with `reaction_smiles.split()[0]`, which truncates at *any* whitespace; ORD records SMILES with trailing non-breaking spaces, and those were being cut short. It now uses `split_cxsmiles_extension`, which only strips a block beginning with `|`.
- **some rows appear that previously did not.** Generation used to raise for any reaction with an enhanced-stereo component, and those reactions were skipped; that crash is fixed.

Two consequences to plan around:

**Reaction substructure search changes.** `derived.reaction_smiles` feeds `rdkit.reactions` through `reaction_from_smiles`, so a query that matched on a catalyst or solvent sitting in the agents slot will no longer match. This is intended — an empty agent block is idiomatic in reaction SMILES, and excluding agents is what stops a ligand recorded only by name from deciding whether a reaction is searchable at all — but it is a user-visible search behavior change, not just a storage detail.

**Canonicalizing recorded values should improve dedup** in `rdkit.reactions`, which is keyed by SMILES: reactions that differed only in atom ordering or agent placement now collapse to one row. Worth measuring before and after rather than assuming.

Scope: rebuild `derived.reaction_smiles` and re-link `rdkit.reactions` for any database loaded before #935. The `.claude/skills/orm-database-load` skill covers the load and the pre-cutover verification steps.

Related: #936, which would require its own rebuild of `derived.compound_smiles`. Worth doing both in one pass if #936 lands first.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

skearnes:
Scope has grown since this was filed, and the tooling to do it did not exist. Recording both.

**What now needs rebuilding**, beyond the agent-free normalization this issue was opened for:

[...]

PR #946, opened 2026-08-08
Rederive by updating rows in place instead of deleting them

skearnes:
## Summary

`--rederive` (#944) deleted a dataset's derived rows before recomputing them. That made it unusable against the live database: search inner-joins through `derived.compound_smiles`, `derived.product_compound_smiles`, and `derived.reaction_smiles` into `rdkit.*`, so emptying those tables does not degrade search — it returns **zero results** for every structure and reaction query until the rebuild finishes.

```sql
JOIN derived.compound_smiles ON compound_smiles.compound_id = compound.id
JOIN rdkit.mols ON rdkit.mols.id = compound_smiles.rdkit_mol_id
```

The delete was only ever a way to defeat the `NOT EXISTS` guards. Updating in place achieves the same rebuild with no window where a row is missing.

## Changes

- Rederiving selects every row rather than only the missing ones, and the inserts become upserts with `DO UPDATE` conditional on the value actually differing.
- A row whose SMILES changed has its `rdkit_mol_id` / `rdkit_reaction_id` cleared, handing it to the linking pass. Those rows alone are briefly unsearchable, rather than the whole corpus.
- A reaction or compound that no longer derives anything has its stale value cleared in place, but still gets no row created — an underivable one is re-attempted next run rather than recorded as derived-to-NULL.
- `delete_derived_data` is removed along with the approach that needed it.
- The `orm-database-load` skill documents the new behavior and why it is live-safe.

## Testing

- `uv run pytest -n auto`: 886 pass, 2 skipped, including the 56 ORM tests against a live Postgres. `ruff`, `ty`, and all pre-commit hooks clean; `ty` reports the same 42 diagnostics as `main`.
- `test_rederive_never_leaves_a_row_absent` corrupts a value, rederives, and asserts the row counts are unchanged, the value is corrected, and the RDKit link is cleared.
- `test_rederive_is_a_no_op_when_nothing_changed` rederives an already-current database and asserts no link was cleared. Since `DO UPDATE` always clears the link when it fires, surviving links prove the conditional held and no row was rewritten.

## Notes

Beyond being safe, this is faster: the previous version rewrote every row unconditionally, where this writes only the ones that changed — on the corpus, roughly 2,046 of 198,421 compound identifiers from the #939 coordinate-bond fix.

`--prune_rdkit` (#945) is unaffected and still wanted: a changed SMILES links to a new structure and leaves the old one referenced by nothing.

This means the pending #937 rebuild can run in place against `ord_20260702` rather than needing a fresh database and a cutover.

The PR replaces delete-and-rebuild rederivation with conditional in-place upserts and removes stale derived rows only when an entity no longer yields SMILES.
- Recomputes all entities during `--rederive` while preserving unchanged rows and RDKit links.
- Clears RDKit links only when a derived value changes so the linking stage can rebuild them.
- Deletes stale derived rows for compounds, product compounds, and reactions that become underivable.
- Adds regression coverage for row preservation, unchanged-value no-ops, and underivable entities.
</details>

The PR appears safe to merge.

The previously reported stale-link failure is addressed for compounds, product compounds, and reactions, and no blocking failure remains.

| Filename | Overview |
|----------|----------|
| ord_schema/orm/database.py | Reworks rederivation around conditional upserts, targeted stale-row deletion, and RDKit relinking without leaving the previously reported stale links. |
| ord_schema/orm/database_test.py | Adds focused regression tests for in-place correction, no-op rederivation, and removal of values that no longer derive. |
| .claude/skills/orm-database-load/SKILL.md | Documents the updated rederive lifecycle, its temporary per-row search impact, and subsequent RDKit relinking requirements. |

</details>

```mermaid
flowchart TD
    A[Run rederive] --> B[Recompute derived SMILES]
    B --> C{Value produced?}
    C -->|No| D[Delete existing stale derived row]
    C -->|Yes| E{Value changed?}
    E -->|No| F[Preserve row and RDKit link]
    E -->|Yes| G[Update row in place]
    G --> H[Clear RDKit link]
    H --> I[RDKit insertion and linking passes]
    I --> J[Searchable under new structure]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["Drop a derived row that no longer derive..."](https://github.com/open-reaction-database/ord-schema/commit/448aa263039ec334646baa0fb3416a3f248a7355) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51366515)</sub>

**Context used:**

- Knowledge Base — [ORM and Database Loading](https://app.greptile.com/open-reaction-database/-/custom-context/knowledge-base/open-reaction-database/ord-schema/-/docs/orm-database.md)

### DIFF
--- a/.claude/skills/orm-database-load/SKILL.md
+++ b/.claude/skills/orm-database-load/SKILL.md
@@ -75,8 +75,7 @@ also means an existing row is **never revisited**: after a change to how SMILES 
 the command above writes nothing and exits clean, leaving every stale row in place. Silence
 here is not success.
 
-`--rederive` deletes a dataset's derived SMILES rows before deriving them, which is what turns
-the idempotent passes into a rebuild:
+`--rederive` recomputes every row and updates the ones whose value changed:
 
 ```sh
 python -u -m ord_schema.orm.scripts.add_datasets \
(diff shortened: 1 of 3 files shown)


skearnes:
Both the derived-to-NULL question and the Greptile summary point at the same defect, fixed in 448aa26.

**The reaction path was writing the very state the comment beside it argues against.** Clearing a stale row to NULL still leaves a row present, and a row present means "derived" — so the reaction would sit behind the `NOT EXISTS` guard and no later pass would retry it. That is the property I claimed to be protecting by not recording derived-to-NULL in the first place.

**The compound path had no clearing at all**, which is the gap the summary flags. A compound that stops deriving kept both its stale SMILES and its `rdkit_mol_id`, so it stayed searchable under a structure the source no longer supports — worse than the reaction case it was supposed to mirror.

Both now delete the row, which gives one invariant everywhere: **a row is present exactly when a SMILES was derived.** A stale value can never stay joinable, and an entity that derives nothing is retried by any later pass. Only rows flipping from derivable to underivable are touched, and by definition they have no usable value, so nothing searchable is lost that was not already wrong.

Confirmed a normal derive writes no NULL rows at all — the fixture has 80 reactions and 79 derived rows, the underivable one having no row rather than a NULL one. (The `test_update_rdkit_tables_idempotent` docstring calls that reaction "NULL reaction_smiles"; it is stale wording for "no row", left alone here.)

New parametrized test over both tables asserts the row is gone and that no NULL row is left behind.

888 tests pass, `ty` at the same 42 diagnostics as `main`.

PR #948, opened 2026-08-08
Compile a query the model emits instead of accepting the SQL it writes

skearnes:
## Summary

The model emits a structured `Query`, and the library compiles it to DuckDB SQL. Not for safety alone — bound parameters already gave that — but for **cost**: SQL's expensive programs cannot be enumerated, so every guard on generated SQL is a blacklist someone keeps adding to. This grammar has one relation, no join, no recursion, and no set-returning function, so the worst program expressible in it is one pass over the corpus plus a sort. Expensive queries are not discouraged; they are unwritable.

It still reaches all 442 leaves, because a column is a *parameter* rather than part of the grammar — the difference between this and the per-predicate enumeration it replaces.

**Stacked on #949**, which supplies the schema description this queries against; review that first. Design record in [ord-logbook#25](https://github.com/open-reaction-database/ord-logbook/pull/25).

## Changes

- `agent/query.py` — the grammar and its compiler. Paths resolve against `projection.SCHEMA` before any SQL exists, so an unknown field is a compile error with a near-match suggestion, and an operator that cannot suit its leaf's type is refused rather than run.
- `agent/sql.py` — a backstop on the compiler's own output, not the guard it started as.
- `agent/README.md`, and the `agent` extra carrying `duckdb` and `pydantic`.

## Testing

`uv run pytest` — 913 passed, 69 in the agent package.

Verified against the projection of `ord_dataset-488402f6` (40,000 reactions): compiled queries agree with hand-written SQL on every case tried — co-membership, `forall`, nested quantifiers — at the same speed, and their output passes `sql.validate`.

Two refusals are the point and are pinned: a path crossing a repeated level without a quantifier, where `UNNEST` would silently mean "any" *and* multiply the row count; and repeated levels compiling to list lambdas rather than `UNNEST` in a `FROM` clause, so the 27–200× cliff is unreachable rather than policed.

Fixed under review: `sql.validate` read DuckDB's *rendered* plan, which draws into a fixed-width canvas and silently drops subtrees — eleven `UNION ALL` branches made the exact query the tests assert is refused pass instead. It reads `EXPLAIN (FORMAT json)` now, walking operator names structurally, and refuses a plan it cannot read. And `compile_query` inserted its `table` argument verbatim, so a caller could pass `reactions, range(1000000000)` and get a cross join the `Query` never asked for.

## Notes

`sql.validate` still does not bound cost, and says so — that is a property of accepting SQL, and the reason the model no longer writes it. A caller running arbitrary SQL needs its own statement timeout or row cap.

Execution still needs a sandbox, unchanged: `enable_external_access=false` cannot be combined with a lazy Parquet view, so the real corpus needs `allowed_directories` or a subprocess. An IR removes the reasons to fear the query, not the reasons to contain the process.

Ibis was measured as a possible backend — same SQL shape, same speed — but its expressions are built from Python lambdas, which a model cannot emit without evaluating model-supplied Python. Worth revisiting if a second dialect is ever needed.

The PR introduces a structured query grammar and DuckDB SQL compiler so model-generated queries are constrained to one relation, while strengthening SQL-plan validation and documenting the execution boundary.
- Adds schema-aware predicates, quantifiers, aggregation, ordering, and query compilation.
- Restricts relation overrides to plain identifiers, closing the previously reported grammar bypass.
- Reads DuckDB plans as structured JSON and fails closed when a plan cannot be interpreted.
- Adds agent dependencies, documentation, and comprehensive query and validation tests.
</details>

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| ord_schema/agent/query.py | Adds the schema-aware query IR and compiler, including the identifier restriction that fixes the previously reported table-expression bypass. |
| ord_schema/agent/query_test.py | Exercises path resolution, quantifier semantics, aggregation, injection-resistant values, grammar boundaries, and table-name refusal. |
| ord_schema/agent/sql.py | Implements read-only SQL planning against an empty projection and structurally inspects DuckDB JSON plans while accurately documenting that arbitrary SQL cost is not bounded. |
| ord_schema/agent/sql_test.py | Covers statement restrictions, disabled file access, parameter binding, exploding-plan rejection, and fail-closed plan parsing. |
| pyproject.toml | Adds the optional agent dependency group and includes it in the test environment. |
| uv.lock | Resolves the newly added DuckDB and Pydantic dependencies; advisory-matched Keras entries are unchanged and unrelated to the agent dependency path. |
| ord_schema/agent/README.md | Documents the structured-query design, supported grammar, compilation workflow, affordability boundary, and remaining execution-sandbox responsibility. |

</details>

```mermaid
flowchart LR
    Model[Model emits structured Query] --> Validate[Validate paths and operators against projection schema]
    Validate --> Compile[Compile constrained DuckDB SQL]
    Compile --> Plan[Validate structured JSON plan]
    Plan --> Execute[Caller binds parameters and executes against reactions]
```
</details>

<sub>Reviews (9): Last reviewed commit: ["Compile a query the model emits instead ..."](https://github.com/open-reaction-database/ord-schema/commit/c73c838f9d472cc97efb00807e4161436ed4c372) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51379074)</sub>

### DIFF
--- a/ord_schema/agent/query.py
+++ b/ord_schema/agent/query.py
@@ -1,0 +1,485 @@
+# Copyright 2026 Open Reaction Database Project Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A query a model can emit, and its compilation to DuckDB SQL.
+
+A model emits one of these rather than SQL. The reason is not safety alone -- bound
+parameters already gave that -- but **cost**. SQL's expensive programs cannot be
+enumerated, so every guard on generated SQL is a blacklist someone has to keep adding
+to. This grammar has one relation, no join, no recursion, and no set-returning
+function, so the worst program expressible in it is one pass over the corpus and a
+sort. Expensive queries are not discouraged here; they are unwritable.
+
+It reaches every column the projection has, because a column is a *parameter* rather
+than part of the grammar. That is the difference between this and the per-predicate
+enumeration it replaces: adding a field upstream adds a queryable column and costs
+nothing here.
+
+Two things the compiler settles that SQL leaves to whoever writes it:
+
+* **Quantifiers are stated, never assumed.** A path crossing a repeated level is
+  refused unless an ``exists`` or ``forall`` binds it. The same intent in SQL is spelled
+  ``UNNEST``, which silently means "any" *and* multiplies the row count.
+* **Repeated levels compile to list lambdas**, never to ``UNNEST`` in a ``FROM`` clause
+  -- measured at 27-200x the cost for identical answers, and the idiom every SQL
+  tutorial teaches. A compiler cannot reach for the wrong one.
+
+Co-membership survives, which a reaction-keyed flat table cannot express:
+``exists(components, A and B)`` is one component that is both, while
+``exists(components, A) and exists(components, B)`` is a reaction containing each.
+They are different nodes rather than a subtlety of join keys.
+
+Compounds are named, never spelled. A ``{"compound": "thf"}`` value compiles to a bound
+parameter, so the model never writes a structure and the caller resolves the name
+through :mod:`ord_schema.resolvers` before binding.
+"""
+
+import dataclasses
+import difflib
+import re
+from typing import Annotated, Any, Literal
+
+import pyarrow as pa
+from pydantic import BaseModel, Field, model_validator
+
+from ord_schema import projection
+
+# The relation a compiled query reads. The only one in scope, so nothing else is
+# nameable and a join has nothing to join to.
+TABLE = "reactions"
+
+_COMPARISONS = {"eq": "=", "ne": "<>", "lt": "<", "le": "<=", "gt": ">", "ge": ">="}
+_ORDERED = frozenset({"lt", "le", "gt", "ge"})
+_TEXT = {"contains": "contains", "starts_with": "starts_with", "ends_with": "ends_with"}
+_AGGREGATES = {
+    "count": "count",
+    "count_distinct": "count",
+    "sum": "sum",
+    "avg": "avg",
+    "min": "min",
+    "max": "max",
+}
+
+# Measure names and the bound-parameter names for compounds are the only model-supplied
+# strings that reach the SQL as text rather than as parameters, so both are held to an
+# identifier shape a quoting mistake cannot escape.
+_NAME = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+
+class QueryError(ValueError):
+    """A query that cannot be compiled against the projection schema."""
+
+
+@dataclasses.dataclass(frozen=True)
+class Compiled:
+    """A compiled query, and the compound names its parameters still need."""
+
+    sql: str
+    compounds: tuple[str, ...]
+
+
+@dataclasses.dataclass(frozen=True)
+class _Resolved:
+    """Where a path landed: the expression, whether it is a list, and its type."""
+
+    expression: str
+    repeated: bool
+    type: pa.DataType
+
+
+def _members(current: pa.Schema | pa.DataType) -> list[str]:
+    """Returns the field names available on a schema or struct type."""
+    if isinstance(current, pa.Schema):
+        return list(current.names)
+    if pa.types.is_struct(current):
+        return [field.name for field in current]
+    return []
+
+
+def _lookup(current: pa.Schema | pa.DataType, name: str, path: str) -> pa.Field:
+    """Returns the field ``name`` within ``current``, or raises a helpful QueryError."""
+    members = _members(current)
+    if name not in members:
+        if not members:
+            raise QueryError(f"{path}: {name!r} has no fields to descend into")
+        close = difflib.get_close_matches(name, members, n=3)
+        suggestion = f"; did you mean {', '.join(map(repr, close))}?" if close else ""
+        raise QueryError(f"{path}: no field named {name!r}{suggestion}")
+    if isinstance(current, pa.Schema):
+        return current.field(name)
+    return current.field(name)
+
+
+def resolve(
+    path: str, *, schema: pa.Schema = projection.SCHEMA, root: str | None = None
+) -> _Resolved:
+    """Resolves a dotted path against the projection schema.
+
+    Descending through a repeated level turns the expression into a list of the
+    elements beneath it, so the caller can tell a scalar it may compare from a level it
+    must quantify over.
+
+    Args:
+        path: Dotted column path, e.g. ``conditions.temperature.setpoint_kelvin``.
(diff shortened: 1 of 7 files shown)


ord_schema/agent/sql.py
```diff
@@ -0,0 +1,131 @@
...
+    Returns:
+        A connection scoped to a single empty relation.
+    """
+    connection = duckdb.connect()
+    connection.register(TABLE, schema.empty_table())
+    connection.execute("SET enable_external_access=false")
+    return connection
+
+
+def validate(
```
skearnes: do we really want agents writing sql vs. creating a programmatic query generation tool like we have for the (much smaller) ORM query surface?

ord_schema/agent/schema.py
```diff
@@ -0,0 +1,126 @@
...
+            lines.append(f"{indent}{name}: MAP<{key}, STRUCT>")
+            for field in value:
+                _render(field.name, field.type, depth + 1, lines)
+        else:
+            lines.append(f"{indent}{name}: MAP<{key}, {_scalar_name(value)}>")
+    else:
+        lines.append(f"{indent}{name}: {_scalar_name(dtype)}")
+
+
+def describe(schema: pa.Schema = projection.SCHEMA) -> str:
```
skearnes, line 118: should we be checking the generated description into git?

skearnes: Yes — done in 92a6764.

`projection_schema.txt` is checked in and `test_the_snapshot_matches_what_describe_renders` regenerates and compares. The argument that convinced me is the same one behind `views_test.test_the_written_schema_is_the_documented_one`: an artifact asserted against its own generator proves nothing. This text *is* the contract with the model, and a proto field added upstream silently rewrites it — checking it in puts that in the diff where a reviewer sees it, instead of leaving it to be discovered from a model's behavior.

It is deliberately a review artifact and never a source: `describe()` still generates what ships, so the file cannot drift into being load-bearing. The test failure message carries the regeneration command.

_Analysis performed with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)._

ord_schema/agent/sql.py
skearnes: I think you're right, and the Greptile finding on this same PR is the evidence.

**The design-log argument does not reach a programmatic surface.** [The entry](https://github.com/open-reaction-database/ord-logbook/blob/main/entries/2026-07-31-nl-query-over-the-projection.md) rejected structure because `NLQuery` is "an enumeration, not a query language" — one field per ORM predicate, so every new question shape costs a field. True of *that* design, but it conflates two things. Enumerating **question shapes** (`min_yield`, `max_yield`, `reaction_smarts`…) does not scale. Enumerating **operators** (`eq`, `lt`, `contains`, `any`, `all`) with the column path as a *parameter* scales to 389 columns without 389 fields, because the columns are data rather than grammar. I wrote that entry's conclusion into this PR without noticing the gap.

**Accepting SQL makes cost unguardable, not just unguarded.** Everything below passes `validate()` today:

| query | |
|---|---|
| `FROM reactions a, reactions b` | 5.9×10¹² rows |
| `FROM reactions, range(1000000000)` | accepted |
| recursive CTE counting to 10⁹ | accepted, touches no table |
| `SELECT * FROM reactions` | accepted |

I can add entries, but not finish: the set of expensive programs expressible in SQL cannot be listed. A compiler makes it structural — no self-join, no recursive CTE, no cross product, because none is in the grammar. The `UNNEST` hazard the entry treats as *the* known risk is a symptom of the same thing: the model reaches for the textbook idiom because the grammar allows it, where a compiler emits list lambdas by construction and the 27–200× cliff stops existing.

**And it lands on the line the agent-access entry already drew** — lookup and search stay mediated, analysis goes direct. A predicate IR is the right shape for the mediated surface. Analysts wanting aggregation already have DuckDB against the file, where the user is a human who is not the injection risk.

The one thing a pure predicate IR gives up is aggregation, and "which catalyst gives the highest average yield" is a fair NL question — so I would add a bounded `group_by` + aggregate form rather than pretend it is out of scope. That keeps the grammar small and closed.

**What survives either way:** `schema.describe()` is unchanged (a compiler has to tell the model the columns too), and `sql.validate()` demotes from primary guard to a backstop on compiler output, which is a much more honest job for it.

Happy to build the IR — say the word and I'll write up the grammar first rather than start coding. Also worth a revision to that logbook entry, since its conclusion is what I followed.

_Analysis performed with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)._

ord_schema/agent/sql.py
```diff
@@ -0,0 +1,131 @@
...
+            f"expected a single statement, got {len(statements)}; "
+            "a query answers one question"
+        )
+    if statements[0].type != duckdb.StatementType.SELECT:
+        raise InvalidQueryError(
+            f"expected a SELECT, got {statements[0].type.name}; this surface reads"
+        )
+    connection = _connect(schema)
+    try:
+        plan = connection.execute(f"EXPLAIN {sql}", parameters or {}).fetchone()
```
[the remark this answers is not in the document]
skearnes: Confirmed, and it is worse than one bypass — fixed the claim rather than the code, in 92a6764.

I ran your case plus neighbours. All accepted by `validate()`:

| query | |
|---|---|
| `FROM reactions a, reactions b WHERE a.smiles = b.smiles` | 2.4M² |
| triple self-join | accepted |
| `FROM reactions, range(1000000000)` | accepted |
| `generate_series(1, 1000000000)` | accepted |
| recursive CTE counting to 10⁹ | accepted, touches no table |
| correlated subquery per row | accepted |
| `SELECT * FROM reactions` | accepted |

So the module was describing a class of defect ("answerable, but not affordably") that it catches exactly one spelling of. I could keep adding entries, but not finish — the set of expensive programs expressible in SQL cannot be enumerated, which is a property of accepting SQL rather than of this check.

The module docstring now carries a warning saying so explicitly, and points out that a caller running validated SQL against a real corpus needs its own statement timeout, row cap, or memory limit and must not read this module as having provided one. `test_cost_is_not_bounded` pins four of the above so the claim and the behavior cannot drift apart again — with a note that if a later change *does* bound cost, that test should fail and be rewritten rather than deleted.

This also fed a broader design question the author raised on this PR: whether a compiled query IR beats accepting SQL at all, precisely because a compiler makes affordability structural instead of policed.

_Analysis performed with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)._

ord_schema/agent/query.py
```diff
@@ -0,0 +1,480 @@
...
+    """Compiles a query to DuckDB SQL over the projection.
+
+    Args:
+        query: The query to compile.
+        schema: Schema to resolve paths against; the projection schema by default.
+        table: Relation name to read.
+
+    Returns:
+        The SQL, and the compound names whose resolved SMILES the caller binds.
+
```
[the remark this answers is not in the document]
skearnes: Fixed — `table` is now held to the same identifier shape as measure names, and refused otherwise.

You are right that it defeated the boundary. The parameter is caller-supplied rather than model-supplied, so exploiting it needs a caller who could run arbitrary SQL anyway — but the module claims the worst expressible program is one pass plus a sort, and that claim held only by convention while one argument reached the `FROM` clause as text. A guarantee with an asterisk is worth less than the sentence it took to remove it.

Three cases pinned, plus one confirming a plain relation name still works:

```
reactions, range(1000000000)     refused
reactions AS a, reactions AS b   refused
read_parquet('/etc/hosts')       refused
projections                      accepted
```

_Analysis performed with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)._

PR #960, opened 2026-08-10
Answer structure queries from an occurrence index

skearnes:
## Summary

"Which reactions contain this molecule, in this role?" no longer scans every reaction. An occurrence index — one row per structure occurrence, carrying the corpus-wide ID, the path it sat at, and the element's own `reaction_role` — turns that scan into a filter over a narrow table: 14.1M rows and 130 MB against the projections' 1.65 GB.

**Read the scope before the speedups.** The planner is all-or-nothing: it replaces the whole query or declines. So the index answers exactly one shape today — a bare `exists` at an indexed path with one structure predicate and an optional role — and the numbers below are that shape's. See [What routes, and what does not](#what-routes-and-what-does-not).

## Changes

- `Corpus._occurrences` builds the index lazily, under a lock, on the first query that can use it.
- `Corpus._plan` routes a query to the index when the index can answer it, and returns None otherwise.
- Paths and traversals come from the same schema walk and the same `query.resolve` the compiler uses, so the index reaches the elements the projection does by construction rather than by one list agreeing with another.

## Why the index carries `reaction_role`

A plain `structure_id → reaction_ids` index answers "reactions containing pyridine" but not "reactions where pyridine is the *solvent*" — that is the intersection which over-returns by 94%, and the reason element binding exists at all. Keeping the role beside the structure keeps the bound query a condition on one row.

## Measurements

End to end over the whole corpus through `Corpus.search`, with identical match sets on every row. All three are the shape the planner routes; a query one clause away from these pays the projection column:

| query | projection | index | speedup | reactions |
| --- | --- | --- | --- | --- |
| pyridine among the inputs | 7.56s | 1.46s | 5.2x | 660,671 |
| pyridine as the solvent | 5.00s | 1.74s | 2.9x | 25,805 |
| a boronic acid | 2.93s | 0.15s | 19.8x | 132,600 |

The SQL alone is 17-67x faster (0.220s vs 3.755s unbound, 0.046s vs 3.072s bound). End to end is lower because the RDKit screen and verify is untouched and now dominates: pyridine's match alone is about 1.4s of its 1.46s, so the SQL fell from roughly 3.5s to 0.2s and the bottleneck moved rather than shrank. Boronic acid, whose match is cheap, shows the full effect.

Cost: the build is four passes over the projections, upwards of a minute at corpus scale alongside the library build. Paid on the first query that can use it, not at open. Collapsing it to a single pass is an obvious follow-up.

## What routes, and what does not

Every query below is one somebody would actually ask. Only the first routes:

| query | answered by |
| --- | --- |
| pyridine as the solvent | **index** |
| yield > 50% **and** pyridine as the solvent | projection |
| how many reactions use pyridine as a solvent | projection |
| pyridine as the solvent, ordered | projection |
| pyridine as the solvent **and** a boronic acid input | projection |

The cause is `Corpus._plan`'s shape rather than the index's contents. An occurrence row is perfectly able to contribute to a bigger query; what is missing is a way to use it for *part* of one. So a second structure predicate, a scalar beside the quantifier, an aggregate, an ordering, a negation, a disjunction, or a `forall` all fall back, and the projection answers them exactly as it did before this.

The generalization is a substitution one level down, where `query._quantifier` compiles an element predicate to `len(list_filter(inputs.components, e -> get_bit(...) AND e.reaction_role = 'SOLVENT')) > 0`. When the index can answer that clause, emit `reaction_id IN (SELECT reaction_id FROM occurrences WHERE ...)` instead and let the rest of the query compile unchanged. Aggregates, orderings, and limits then compose for free; negation and disjunction work too, since `NOT (reaction_id IN ...)` means what the negated `list_filter` means; two structure predicates become two index lookups. Only `forall` stays out, because "every element matches" is not a membership test.

That rewrite deletes `_plan` and `_index_terms` — about 100 lines of this PR. Everything else, which is most of it, is what the rewrite needs: the schema walk deriving the indexed paths from the compiler's own traversals, the build, the pairing, and the differential tests. Follow-up rather than a blocker, but it is where the value is, and nothing further should be built on `_plan`.

## What the review round changed

Eight reviewers went over this. What they found:

**An empty index answered every structure query with "no matches."** Every path the projection can carry a structure at is indexed, so a corpus holding structures and indexing none of them means a traversal reached nothing — now a `PairingError`, with a per-path count logged so a path contributing nothing is visible before it costs an answer.

**The build could wedge the corpus.** It ran on the shared connection, which the module docstring forbids for anything running while a search is in flight, and `CREATE TABLE` left an interruption between the create and the flag colliding with itself forever. It takes a cursor and replaces the table.

**Neither path said it answered.** The two relations are supposed to agree, so a disagreement was only debuggable from process state nobody recorded. Both log now.

**`limit` is where the two paths visibly differ.** Neither orders what it was not asked to order, so a limited query with no `order_by` takes some of the match set and they take different rows of it. Stated in the planner, the module docstring, and the README; tested for what holds — the count is the limit, the rows are the match set's.

`_index_terms` returned `(True, role)` where the bool could not be `False` and the caller discarded it.

## Testing

Two paths to one answer is the risk here, so the tests are differential: every shape the planner accepts is run both ways and the results compared. Sabotages confirm they bite — dropping the role condition, dropping the path condition, and dropping `DISTINCT` each fail the suite.

Worth recording: my first version of these tests caught only the first of those three. The fixture's chemistry made the path-condition sabotage agree by accident, and comparing results as *sets* hid the duplicate rows that losing `DISTINCT` produces. Both gaps are covered by cases chosen to discriminate — a structure that lives only at another path, and a query matching two components of one reaction.

**But the suite verified what `_plan` decided, not what `search` ran.** Replacing `compiled = indexed ...` with the projection compile passed all 56 tests: the feature became a no-op that still paid the build, and the differential test compared the projection to itself. A routed query now reads a row planted only in `occurrences`, so a search that stops routing fails.

Nine more sabotages, each previously missed and now caught: the routing discarded, the role literal unescaped (`SOLVENT' OR '1'='1` matches every row at every path), the limit dropped, the index lock removed, the empty-index guard removed, and the four role guards relaxed one at a time — any element field, any comparison, any value type, last role wins.

A second fixture holds a workup component and an authentic standard, so `workups.input.components` and `outcomes.products.measurements.authentic_standard` — the two deepest traversals, and the two the main fixture had no data for — are compared against the projection rather than comparing nothing to nothing.

`pytest -n auto` — 1054 passed, 2 skipped. `pre-commit` clean.

The PR adds a lazily built occurrence index and narrowly routes eligible structure queries through it while retaining projection execution as the fallback.
- Derives indexed structure-bearing paths from the projection schema and compiler traversal logic.
- Preserves element-level role binding and corpus-wide structure IDs in occurrence rows.
- Adds locking, index-completeness checks, path-level logging, and extensive differential and concurrency tests.
- Documents planner scope, initialization cost, and unordered-limit behavior.
</details>

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| ord_schema/agent/execute.py | Adds schema-derived occurrence indexing, narrow query planning, synchronized lazy construction, completeness validation, and routed execution without an accepted blocking issue. |
| ord_schema/agent/execute_test.py | Adds broad planner-shape, differential-result, deep-path, escaping, deduplication, refusal, routing, and concurrent-build coverage. |
| ord_schema/agent/README.md | Documents occurrence-index semantics, routing limits, build costs, role binding, and unordered-limit behavior consistently with the implementation. |

</details>

```mermaid
flowchart TD
  Q[Corpus.search request] --> P{Occurrence planner accepts shape?}
  P -->|No| C[Compile against projection]
  P -->|Yes| B{Occurrence index built?}
  B -->|No| I[Build occurrence rows under lock]
  B -->|Yes| O[Use occurrence relation]
  I --> O
  C --> M[Resolve and evaluate structure predicates]
  O --> M
  M --> E[Execute compiled SQL]
  E --> R[Arrow result]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["Count the structures the index reached, ..."](https://github.com/open-reaction-database/ord-schema/commit/02b38a0b1cb4b66fa40d9e230fbaff6aa1d43447) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51697038)</sub>

**Context used:**

- Knowledge Base — [Agent Query Surface](https://app.greptile.com/open-reaction-database/-/custom-context/knowledge-base/open-reaction-database/ord-schema/-/docs/agent-query.md)

### DIFF
--- a/ord_schema/agent/README.md
+++ b/ord_schema/agent/README.md
@@ -109,18 +109,21 @@ larger one. An **occurrence index** — one row per structure occurrence, carryi
 corpus-wide ID, the path, and the element's own `reaction_role` — makes that a filter
 over a narrow table rather than a scan of every reaction. Keeping the role beside the
 structure is what keeps a bound query a condition on one row. At corpus scale the index
-is 14.1M rows and about 130 MB, held in memory for the life of the `Corpus`, and answers
-pyridine among the inputs in 1.46s, pyridine as the solvent in 1.74s, and a boronic acid
-in 0.15s. What remains for common patterns is the RDKit screen and verify, which the
-index does not touch.
+is 14.1M rows and about 130 MB, resident for the life of the `Corpus`. A search for
+pyridine among the inputs returns in 1.46s **end to end**, pyridine as the solvent in
+1.74s, and a boronic acid in 0.15s — and for a common pattern nearly all of that is the
+RDKit screen and verify, which the index does not touch: pyridine's match alone is about
+1.4s of its 1.46s. What the index cut is the reaction lookup, from roughly 3.5s to 0.2s.
 
 The index answers only the shape it holds: a bare `exists` at an indexed path whose body
 asks for one structure and at most one role. A query reaching another element field, or
 needing a projection column to group or sort by — and every `forall`, negation, and
-disjunction — runs against the projection. Both paths screen and verify through the same
-compiler, and the log line says which one answered. The index is built on the first query
-that routes to it, four passes over the projections, so a server wanting the first real
-query to be fast should warm it rather than budget for it at open.
+disjunction — runs against the projection, so "reactions with yield > 50% where pyridine
+is the solvent" is answered there rather than here. Both paths screen and verify through
+the same compiler, and the log line says which one answered. The index is built on the
+first query that routes to it, one pass over the projections per indexed path, so a
+server that wants its first real query to be fast should issue a throwaway structure
+query at startup.
 
 A `limit` with no `order_by` is the one place the two paths visibly differ: neither
 relation orders what it was not asked to order, so each returns some rows of the same
@@ -128,12 +131,13 @@ match set and they do not agree on which.
 
 Similarity needs no verification — Tanimoto is defined on the Morgan fingerprint, so
 the screen is the answer — and stays in SQL. The verified match set re-enters the query
-as a `BITSTRING` parameter indexed by the corpus-wide ID (`structure_id` plus the
-file's offset), which is what preserves element binding: `exists(components,
+as a `BITSTRING` parameter indexed by the corpus-wide ID (`structure_id` plus the file's
+offset), which is how the projection path tests a single element: `exists(components,
 substructure(pyridine) and role = SOLVENT)` means one component that is both. The
 alternative — intersecting reaction-ID sets — over-returns by 94% on exactly that
 query; see [ord-logbook#28](https://github.com/open-reaction-database/ord-logbook/pull/28),
-finding 4.
+finding 4. Every corpus-scale figure on this page was measured against that logbook
+entry's snapshot of ORD.
 
 ```python
 from ord_schema.agent import execute, query
--- a/ord_schema/agent/README.md
+++ b/ord_schema/agent/README.md
@@ -106,18 +106,26 @@ should set it to the core count divided by that concurrency instead of leaving i
 
 Finding which reactions hold a match is a separate cost from the chemistry, and the
 larger one. An **occurrence index** — one row per structure occurrence, carrying the
-corpus-wide ID, the path, and the element's own `reaction_role` — replaces that scan
-with a filter over a narrow table (14.1M rows, 130 MB against the projections' 1.65 GB).
-Keeping the role beside the structure is what preserves element binding. End to end at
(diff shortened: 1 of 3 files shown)


skearnes:
Superseded by #962, which carries this branch's commits and review-round fixes and reshapes how the index is spent (one quantifier at a time, as a semi-join).

PR #992, opened 2026-08-27
DEMO: what prefixing enum values would cost

skearnes:
## Summary

Not for merge — a demonstration of what undoing the nested-enum decision costs. Two commits, deliberately at different scales, because the first one alone was misleading.

Both changes are wire-compatible: field and enum *numbers* are untouched, so archived datasets still parse. What moves is the **names**, and text format, protojson, the projection, and a surprising amount of Python all carry names rather than numbers.

## Changes

**Commit 1 — `ElectrochemistryCellType`** (four values, the smallest enum carrying `CUSTOM`). Four proto lines → 159 lines across 15 files, but only **3 hand-edited lines** outside generated code. This is the floor, and on its own it makes the cost look trivial.

**Commit 2 — `CompoundIdentifierType`** (18 values, the most-referenced enum). The realistic case:

| surface | cost |
| --- | --- |
| `CompoundIdentifier.<VALUE>` attribute references | 33, across 7 files |
| **string-form references** | **161, across 21 files** |
| `ord-nielsen-example.pbtxt` | 1,059 lines, every one a rename |
| example submission fixtures | 810 lines |
| `projection_schema.txt` | +1,800 characters (**+8.3%**) for one enum of ~100 |

## Testing

- Wrappers regenerated with protoc 22.3 and the plugin versions CI pins, then `addlicense`.
- Commit 1: 412 tests pass.
- Commit 2: **70 failed, 18 errors** — `ValueError: unknown enum label "SMILES"`. Left failing on purpose; see below.

## Notes

Three findings, none of which I expected going in.

**1. The type checker catches the easy half and misses the hard half.** Pre-commit's `ty` rejected commit 1 immediately:

```
error[unresolved-attribute]: Attribute `UNSPECIFIED` is not defined on `ElectrochemistryCell`
   --> ord_schema/validations.py:400:24
```

`_check_type_and_details` reads `message.UNSPECIFIED` and `message.CUSTOM` off whatever it is handed, and **28 message types** route through it. That one generic validator exists only because every enum spells those two values identically.

But it cannot see the 161 string-form references. Six are library modules, not tests, and the worst is not Python at all:

```python
# ord_schema/orm/database.py:918
AND ord.compound_identifier.type IN ('CXSMILES', 'SMILES')
```

An enum value name inside raw SQL, checked by nothing, that would silently stop matching rows in an already-loaded database.

**2. Text fixtures need a real migration, not a search-and-replace.** A bare `type: SMILES` line does not say which enum owns it. The testdata fixtures went through a binary round-trip — parse under the old schema, re-emit under the new — because binary stores numbers. The example submission could not: it holds templating placeholders, so it was rewritten by indentation depth, and my first threshold was wrong and silently corrupted the reaction-level identifiers before a parse caught it.

**3. Pre-existing, unrelated to this PR.** Both files in `examples/submissions/2_Nielsen_Deoxyfluorination_Screen/` already fail to parse against the current schema — they use `type: NAME` for a `ReactionIdentifier`, and `ReactionIdentifierType` has no `NAME` value. Worth a separate issue.

Related: #993 rewrites the rationale comment in `reaction.proto`, since the [style guide](https://protobuf.dev/programming-guides/style/#enums) permits nesting rather than forbidding it.

The PR demonstrates the source and generated-artifact impact of prefixing protobuf enum values while preserving their numeric wire representation.
- Prefixes enum symbols in the canonical protobuf schema.
- Regenerates Python, protobuf-javascript, protobufjs, fixture, and projection artifacts.
- Updates affected source references and serialized text fixtures.
</details>

<h3>Confidence Score: 4/5</h3>

The PR is not safe to merge because the previously reported protobufjs legacy-name conversion failure remains.

ElectrochemistryCell.fromObject still ignores legacy symbolic strings and leaves the field at zero, silently changing persisted nonzero enum values during conversion.

**Files Needing Attention:** js/ord-schema-protobufjs/index.js

| Filename | Overview |
|----------|----------|
| proto/reaction.proto | Renames enum symbols while retaining their numeric values and binary wire compatibility. |
| js/ord-schema-protobufjs/index.js | Regenerates protobufjs enum conversion logic, with the previously reported legacy-name conversion defect still present. |
| js/ord-schema/proto/reaction_pb.js | Regenerates the protobuf-javascript enum exports using prefixed symbolic keys. |
| ord_schema/validations.py | Updates enum references involved in the demonstrated migration while retaining the intentionally documented generic-validator limitation. |
| ord_schema/search/projection_schema.txt | Updates the model-facing enum-name projection to the renamed symbols. |

</details>

```mermaid
flowchart LR
  Schema[Prefixed protobuf enum names] --> Python[Generated Python bindings]
  Schema --> GPB[protobuf-javascript bindings]
  Schema --> PBJS[protobufjs bindings]
  Schema --> Text[Text-format fixtures]
  Schema --> Projection[Projection schema]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["DEMO: prefix a common enum, CompoundIden..."](https://github.com/open-reaction-database/ord-schema/commit/bc090a39c638cf904462aeef97abdfcf809dcda5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57350508)</sub>


skearnes:
Closing — this did its job as a measurement, and the conclusion now lives in the logbook rather than in an open PR.

Write-up: [entries/2026-08-26-migrating-the-proto-build-to-buf](https://github.com/open-reaction-database/ord-logbook/blob/main/entries/2026-08-26-migrating-the-proto-build-to-buf/README.md), under "Why `ENUM_VALUE_PREFIX` stays off". It records the numbers from this branch — 33 attribute references, 161 string-form references across 21 files, 1,059 lines in one fixture, 8.3% growth in the model-facing projection schema, all for one enum of roughly 100 — alongside the generated-code evidence: `protoc` already emits the prefix for C++, and `prost` strips it back off for Rust, so the rule guards against a collision nesting has already prevented.

The branch stays pushed if anyone wants to re-run it.

